PR #4, opened 2018-10-09
Modify StructureUploadWidget

yakutovicha:
Change StructureUploadWidget to be similar to MultiStructureUploadWidget, so that they can later be derived from the same base class

### DIFF
--- a/aiida_widgets/structures.py
+++ b/aiida_widgets/structures.py
@@ -21,11 +21,9 @@ class StructureUploadWidget(ipw.VBox):
         """
 
         self.file_upload = FileUploadWidget(text)
-        self.output = ipw.Output()
         self.viewer = nglview.NGLWidget()
-
         self.btn_store = ipw.Button(description='Store in AiiDA', disabled=True)
-        self.structure_description = ipw.Text(placeholder="Label (optional)")
+        self.structure_description = ipw.Text(placeholder="Description (optional)")
 
         self.structure_ase = None
         self.structure_node = None
@@ -39,7 +37,7 @@ class StructureUploadWidget(ipw.VBox):
             self.data_format.value = node_class
             store = ipw.HBox([self.btn_store, self.structure_description])
 
-        children = [self.file_upload, self.output, self.viewer, store]
+        children = [self.file_upload, self.viewer, store]
 
         super(StructureUploadWidget, self).__init__(
             children=children, **kwargs)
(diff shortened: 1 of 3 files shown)


aiida_widgets/structures.py
```diff
@@ -25,27 +25,26 @@ def __init__(self, text="Upload Structure", node_class=None, **kwargs):
         self.viewer = nglview.NGLWidget()
 
         self.btn_store = ipw.Button(description='Store in AiiDA', disabled=True)
-        self.structure_label = ipw.Text(placeholder="Label (optional)")
+        self.structure_description = ipw.Text(placeholder="Label (optional)")
```
ltalirz: This should then also be called "Description (optional)"

aiida_widgets/structures.py
```diff
@@ -59,21 +58,26 @@ def _on_file_upload(self, change):
         with self.output:
             clear_output()
 
-            tmp = tempfile.mkdtemp() + '/' + self.file_upload.filename
+            self.tmp_folder = tempfile.mkdtemp()
+            tmp = self.tmp_folder + '/' + self.file_upload.filename
             with open(tmp, 'w') as f:
                 f.write(self.file_upload.data)
             self.tmp_file = tmp
```
ltalirz: this can be deleted, right?

aiida_widgets/structures.py
```diff
@@ -59,21 +58,26 @@ def _on_file_upload(self, change):
...
-            formula = atoms.get_chemical_formula()
-            self.structure_label.value = "{} ({})".format(
-                formula, self.file_upload.filename)
-            self.atoms = atoms
-            self.refresh_view()
-            self.btn_store.disabled = False
+            self.select_structure(name=self.file_upload.filename)
+
+    def select_structure(self, name=None):
+        self.tmp_file = self.tmp_folder + '/' + name
```
ltalirz: function has parameter name=None, but this won't work with name=None

aiida_widgets/structures.py
```diff
@@ -59,21 +58,26 @@ def _on_file_upload(self, change):
...
+        traj = ase.io.read(self.tmp_file, index=":")
```
ltalirz: the try/except will come in a separate pull request?

aiida_widgets/structures.py
```diff
@@ -103,14 +107,8 @@ def _on_click_store(self, change):
         else:
             source_format = 'ASE'
 
-        # determine target format
-        if self.structure_node_class is None:
```
ltalirz: I think I remember now why I added this `structure_node_class` variable: as a way for a user of the widget to get the node class without having to know about how it is implemented (data_format.value). E.g. you can just tab to autocomplete in ipython.

There is, however, a very simple way of combining your and my approach, which I think is better than either of them: use a property
```
@property
def node_class(self):
    return self.data_format.value
@node_class.setter
def node_class(self, value):
    self.data_format.value = value
```

aiida_widgets/structures.py
yakutovicha: Yes.

aiida_widgets/structures.py
yakutovicha: Ok, I deleted the default value

aiida_widgets/structures.py
yakutovicha: I am adding this in the next commit

aiida_widgets/structures.py
yakutovicha: Done.

aiida_widgets/structures.py
```diff
@@ -21,31 +21,28 @@ def __init__(self, text="Upload Structure", node_class=None, **kwargs):
         """
 
         self.file_upload = FileUploadWidget(text)
-        self.output = ipw.Output()
```
ltalirz: Can you maybe explain a bit the reasoning behind removing the output?
E.g. if we want to clear at some point what we have written, is it still possible?

yakutovicha: Yes, the Output was appearing above the structure, which does not look nice. All the other warning or error message were appearing below the structure

PR #10, opened 2018-12-12
two functions to check crystal equivalence

cpignedoli:
to use them:
formula=ase_structure.get_chemical_formula()
structures_with_same_formula=formula_in_qb(formula)
for one in structures_with_same_formula:
    if check_crystal_equivalence(ase_structure,one[0]):
        print("find equivalent structure with PK",one[1])

### DIFF
--- a/aiidalab_widgets_base/crystal_sim_crystal.py
+++ b/aiidalab_widgets_base/crystal_sim_crystal.py
@@ -1,0 +1,55 @@
+import itertools
+import numpy as np
+import spglib
+
+from ase.spacegroup import crystal
+from ase.data import atomic_numbers, atomic_names
+from ase.spacegroup import Spacegroup
+from numpy.linalg import norm
+
+
+def ase_to_spgcell(ase_atoms):
+    return (ase_atoms.get_cell(),
+            ase_atoms.get_scaled_positions(),
+            ase_atoms.get_atomic_numbers())
+def a_equiv_b(a,b):
+    """Function that identifies whether two crystals are equivalent"""
+
+    # getting symmetry datasets for both crystals
+    cryst_a = spglib.get_symmetry_dataset(ase_to_spgcell(a), symprec=1e-5, angle_tolerance=-1.0, hall_number=0)
+    cryst_b = spglib.get_symmetry_dataset(ase_to_spgcell(b), symprec=1e-5, angle_tolerance=-1.0, hall_number=0)
+
+    samecell = np.allclose(cryst_a['std_lattice'], cryst_b['std_lattice'], atol=1e-5)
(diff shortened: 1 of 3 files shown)


aiidalab_widgets_base/search_qb.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import print_function
+from collections import OrderedDict
+from IPython.display import display, clear_output, HTML
```
ltalirz: Perhaps check a few of these imports, most won't be necessary

aiidalab_widgets_base/crystal_sim_crystal.py
```diff
@@ -0,0 +1,55 @@
...
+from ase.data import atomic_numbers, atomic_names
+from ase.spacegroup import Spacegroup
+from numpy.linalg import norm
+
+
+def ase_to_spgcell(ase_atoms):
+    return (ase_atoms.get_cell(),
+            ase_atoms.get_scaled_positions(),
+            ase_atoms.get_atomic_numbers())
+def a_equiv_b(a,b):
```
ltalirz: perhaps a more expressive  name might be "check_crystal_equivalence(crystal_a, crystal_b)"

yakutovicha:
@cpignedoli , you said you have and idea for the update of this `a_equiv_b`. Should we wait until this is done?

aiidalab_widgets_base/crystal_sim_crystal.py
```diff
@@ -0,0 +1,55 @@
...
+            sorted_b = np.array(sorted([ list(row) for row in list(fullb) ]))
+            if np.allclose(sorted_n, sorted_b, atol=1e-5):
+                return True
+        return False
+
+    if samecell and samenatoms and samespg:
+        cell = cryst_a['std_lattice']
+        rng1 = range(1, int(norm(cell[0])/2.))
+        rng2 = range(1, int(norm(cell[1])/2.))
+        rng3 = range(1, int(norm(cell[2])/2.))
```
yakutovicha: Please leave a comment here saying why you are dividing by 2. Like saying that cell vectors can't be smaller than 2.

cpignedoli:
> @cpignedoli , you said you have and idea for the update of this `a_equiv_b`. Should we wait until this is done?

No it was not here, i do not know how this function can be made faster my comment was for "structures_multi.py": there, instead of calling formula_in_qb, one should probably create a list of all chemical formulas present in the database then check the PKs that hve a given formula....

cpignedoli:
I implemented your suggestions, please check if we can now merge.

yakutovicha:
Thanks @cpignedoli

Issue #1, opened 2018-05-31
figure out how to expose `changes` of children


ltalirz:
Figure out how to expose changes in children of the aiida widgets so that one can `observe` the aiida widgets directly.

Note: Until this is fixed, one can still observe changes in the children but needs to know how they are stored internally (e.g. `StructureUploadWidget.btn_store`)

yakutovicha:
Traitles might be a good candidate to look at: https://github.com/ipython/traitlets

ltalirz:
Indeed, traitlets are the way to go.
For observing changes in variables of built-in types (e.g. a `dict`), this is all that is needed:

```python
from traitlets import Dict

class MyWidget(iwp.VBox):
   output_dict = Dict()
   ...

def my_callback(change):
    print("dict changed")

w = MyWidget()
w.observe(my_callback, names=['output_dict'])
```

Since the output of many of our widgets will be AiiDA nodes, we just need to figure out how to make a traitlet type for them (should not be too difficult).

yakutovicha:
just not to forget:
https://traitlets.readthedocs.io/en/stable/using_traitlets.html#observe
https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Custom.html#Making-the-widget-stateful

yakutovicha:
I think we have already figured out, so I close the issue

ltalirz:
We actually haven't figured out the last point I was mentioning.

See here for an example of how to write a custom traitlet type:
https://github.com/jupyter-widgets/traittypes/blob/609b402bcbf8dd3bc0ad4b0dd76fc32d68221871/traittypes/traittypes.py#L28
https://github.com/jupyter-widgets/traittypes/blob/609b402bcbf8dd3bc0ad4b0dd76fc32d68221871/traittypes/traittypes.py#L83

yakutovicha:
> Since the output of many of our widgets will be AiiDA nodes, we just need to figure out how to make a traitlet type for them (should not be too difficult).

This can be achieved simply like this:
```
class MyClass(traitlets.HasTraits):
    structure =  traitlets.Instance(StructureData)
```
That means the structure will become a traitlet that accepts StructureData types only. One can observe this traitlet in the standard way:

```
a = MyClass()
a.observe(my_function, names='structure')
```

@ltalirz please close the issue or let me know if there are other open questions.

ltalirz:
thanks, sounds good!

PR #18, opened 2019-05-02
Work on StructureUploadWidget

yakutovicha:
1) Add `has_structure` traitlet that provides information whether the object contains a structure.
Looks like https://github.com/aiidalab/aiidalab-widgets-base/issues/1 can be closed already
2) Add `storable` argument in `__init__` that controls whether the Store button is displayed
3) Add possibility to provide `examples` through examples argument in `__init__`. The argument
expects to receive a list of paths to files with structures.
4) Add `self.file_path` attribute that contains path to the file with structure
5) Syncronize functionality of `_on_file_upload` and `_on_select_example` functions
6) `structure_node` is now a property of the class


aiidalab_widgets_base/structures.py
```diff
@@ -1,88 +1,138 @@
...
         :param node_class: AiiDA node class for storing the structure.
             Possible values: 'StructureData', 'CifData' or None (let the user decide).
             Note: If your workflows require a specific node class, better fix it here.
+        :param examples: list of paths to the example structures
+        :type examples: list
 
         """
-
         self.file_upload = FileUploadWidget(text)
+        self.select_example = ipw.Dropdown(
```
ltalirz: can this be moved down under "if examples"?

aiidalab_widgets_base/structures.py
```diff
@@ -1,88 +1,138 @@
...
+
         self.viewer = nglview.NGLWidget()
         self.btn_store = ipw.Button(
             description='Store in AiiDA', disabled=True)
         self.structure_description = ipw.Text(
             placeholder="Description (optional)")
 
         self.structure_ase = None
-        self.structure_node = None
+        self._structure_node = None
```
ltalirz: why is this no longer public?
I would say this variable should actually be the traitlet, no?
Then there is no need for has_structure

ltalirz: see https://github.com/aiidalab/aiidalab-widgets-base/issues/1 on how to achieve this

aiidalab_widgets_base/structures.py
```diff
@@ -1,88 +1,138 @@
...
 
         self.file_upload.observe(self._on_file_upload, names='data')
+        self.select_example.observe(self._on_select_example, names=['value'])
         self.btn_store.on_click(self._on_click_store)
 
+    @staticmethod
+    def get_example_structures(examples):
+        if examples:
+            to_return=[("Select structure", False)]
+            to_return += [(s.split('/')[-1], s) for s in examples]
```
ltalirz: please use the portable `os.basename`

ltalirz: instead of `to_return` call it after what it contains (I guess `options`)

aiidalab_widgets_base/structures.py
```diff
@@ -1,88 +1,138 @@
...
+            return to_return
+        else:
+            return []
+
     # pylint: disable=unused-argument
     def _on_file_upload(self, change):
-        self.tmp_folder = tempfile.mkdtemp()
-        tmp = self.tmp_folder + '/' + self.file_upload.filename
-        with open(tmp, 'w') as f:
+        self.file_path = tempfile.mkdtemp() + '/' + self.file_upload.filename
```
ltalirz: can we transform this into another `with` (one for the folder, one for the file) so that we are sure both are deleted?

ltalirz: ah hm... the file is still needed lateron right?

aiidalab_widgets_base/structures.py
```diff
@@ -160,3 +178,38 @@ def node_class(self):
     @node_class.setter
     def node_class(self, value):
         self.data_format.value = value
+
+    @property
+    def structure_node(self):
```
ltalirz: Ok, now I understand your change.

I'm still a bit unsure whether this is the right approach but I get there is value in being able to fetch both a stored and an unstored node "on demand".
That would have to be a very customized traitlet, your implementation makes more sense.

aiidalab_widgets_base/structures.py
ltalirz: scratch that - I think your approach is fine.

aiidalab_widgets_base/structures.py
```diff
@@ -1,88 +1,138 @@
...
 
 
 class StructureUploadWidget(ipw.VBox):
-
+    has_structure = Bool(False)
     DATA_FORMATS = ('StructureData', 'CifData')
-
-    def __init__(self, text="Upload Structure", node_class=None, **kwargs):
+    def __init__(self, text="Upload Structure", storable=True, node_class=None, examples=[], **kwargs):
         """ Upload a structure and store it in AiiDA database.
```
ltalirz: perhaps add an explanation here of how this class should be used (`has_structure`, `structure_node`)

aiidalab_widgets_base/structures.py
```diff
@@ -1,88 +1,138 @@
...
+        :type structure_ase: ASE Atoms
+        :param name: File name with extension but without path
+        :type name: str
+        """
+        self.name = name
+        self._structure_node = None
+        if not structure_ase:
             self.structure_ase = None
             self.btn_store.disabled = True
+            self.has_structure = False
```
ltalirz: what about the `self.structure_description.value`, does it also need to be reset here?

I'm getting a bit uncomfortable with the fact that there are so many things to reset... we should think about whether we can reduce them.
Anyhow, I would suggest to split the resetting part of `select_structure` up into a `reset_structure` function that does nothing but reset the widget.

aiidalab_widgets_base/structures.py
```diff
@@ -99,59 +149,27 @@ def refresh_view(self):
...
-        else:
-            structure_node.description = description
-        structure_node.label = ".".join(name.split('.')[:-1])
-        structure_node.store()
-        self.structure_node = structure_node
-        print("Stored in AiiDA: " + repr(structure_node))
+        if self.structure_node.is_stored:
+            print("Already stored in AiiDA: " + repr(self.structure_node) + " skipping..")
+            return
+        if label:
```
ltalirz: so, since you are no longer passing the label in `_on_click_store`, will the node not have the correct label?

E.g. say you put label+desc. Then another widget asks for the node (=> takes label+desc).
Then you change the label in the widget & click "store".
Will the node still have the old  label?

ltalirz: that is a bit the issue with this "on demand" design - you need to reset the cache whenever any field is edited that could affect the node

aiidalab_widgets_base/structures.py
yakutovicha: > ah hm... the file is still needed lateron right?

yes, right
A not for myself: replace it with `path.join()`

aiidalab_widgets_base/structures.py
yakutovicha: Ok, I will add

aiidalab_widgets_base/structures.py
yakutovicha: After discussion we agreed that I will keep it here for the moment

aiidalab_widgets_base/structures.py
yakutovicha: Ok

aiidalab_widgets_base/structures.py
yakutovicha: Ok, I will add a reset function

aiidalab_widgets_base/structures.py
yakutovicha: solution: reset cache whenever any field that has effect on the node has changed.

aiidalab_widgets_base/structures.py
yakutovicha: done

aiidalab_widgets_base/structures.py
yakutovicha: done

aiidalab_widgets_base/structures.py
yakutovicha: added `reset_structure` function and also put `self.structure_description.value = ''` inside it.

ltalirz:
Feel free to proceed with this without a final review from my side.

aiidalab_widgets_base/structures.py
yakutovicha: In the end I added `reset_structure` function that does only one thing: sets `self.structure_node` to `None`

yakutovicha:
I made the modifications that we agreed on during last MC meeting.

@giovannipizzi, @CasperWA, @ltalirz any volunteer for further review?

PR #23, opened 2019-05-17
Add COD importer in the widgets database

yakutovicha:
Also synchronize it with StructureUploadWidget


yakutovicha:
@ltalirz just try "Dealing with one structure" notebook - it should contain the COD importer tab.

ltalirz:
@yakutovicha unfortunately, I don't have time for a thorough review at the moment - just go ahead according to your needs.

a few things I noticed:
 * regular uploader: I would show the text on the right of the button, could just be a link "Supported structure formats"
 * I think the tab order & titles should be: "Upload" - "COD" - "Examples"
 * the COD link seems to be gone. I think it's very important to keep it, and I would actually make searching for a ID using the COD web site the first option.

I guess you agree that, while the explanation is now more clear, it's still far from ideal to have to read through this... this probably suggests that the Hill notation is not very suitable for this purpose.
Anyhow, we're limited by what the COD provides here, so probably best to simply move to the next database.
https://mpds.io/#start

PR #40, opened 2019-12-19
Work on structure widgets:

yakutovicha:
1) Rename structure upload widget with StructureManagerWidget
2) Extract StructureExamplesWidget out of StructureManagerWidget


yakutovicha:
@csadorf I think I am done here. The next step would be to take the viewer from here https://github.com/aiidalab/aiidalab-widgets-base/blob/72d0f5460136146c400d7511c346d1257f18a6d1/aiidalab_widgets_base/aiida_viewers.py#L59 to make the code of StructureManagerWidget even shorter.

But this I will do in the following PR.

aiidalab_widgets_base/structures.py
```diff
@@ -33,65 +32,52 @@ class StructureUploadWidget(ipw.VBox):  # pylint: disable=too-many-instance-attr
...
-            """<a href="https://wiki.fysik.dtu.dk/ase/_modules/ase/io/formats.html" target="_blank">
-            Supported structure formats
-            </a>""")
+        :type examples: list"""
 
-        self.select_example = ipw.Dropdown(options=self.get_example_structures(examples))
-        self.viewer = nglview.NGLWidget()
-        self.btn_store = ipw.Button(description='Store in AiiDA', disabled=True)
-        self.structure_description = ipw.Text(placeholder="Description (optional)")
+        if not isinstance(importers, list):
```
csadorf: It's typically bad practice to expect a certain file type. I assume that any sequence should be fine here? Why insisting on a list? Tuples not ok?

yakutovicha: Thanks @csadorf, I agree. Also, I was not sure whether to put this check here. I will take it away.

PR #54, opened 2020-02-06
Fix pybel import

yakutovicha:


yakutovicha:
> Is pybel installed as its own package now?

Well, there is a funny story with pybel. It is actually shipped together with openbabel, BUT, there is another package on PyPI that is called [pybel](https://pypi.org/project/pybel/#history). So one should still check whether openbabel is installed.

csadorf:
> > Is pybel installed as its own package now?
> 
> Well, there is a funny story with pybel. It is actually shipped together with openbabel, BUT, there is another package on PyPI that is called [pybel](https://pypi.org/project/pybel/#history). So one should still check whether openbabel is installed.

So the pybel package still requires the openbabel library? Is there any advantage for using this package over the openbabel "native" one?

yakutovicha:
> So the pybel package still requires the openbabel library? Is there any advantage for using this package over the openbabel "native" one?

pybel is a part of openbabel

csadorf:
Yes, but there is https://pypi.org/project/openbabel/ and https://pypi.org/project/pybel/ . So not clear to me what the difference is, but either way is fine with me.

yakutovicha:
> Yes, but there is https://pypi.org/project/openbabel/ and https://pypi.org/project/pybel/ . So not clear to me what the difference is, but either way is fine with me.

Yes, that is what I was trying to tell here:
> there is another package on PyPI that is called pybel

Pybel that is shipped with openbabel is different from the pybel that is available on PyPI. We need the first one, but we don't need the last one.

csadorf:
> Yes, that is what I was trying to tell here:
> 
> > there is another package on PyPI that is called pybel
> 
> Pybel that is shipped with openbabel is different from the pybel that is available on PyPI. We need the first one, but we don't need the last one.

Your PR makes it seem like we need the second one. I think that's the source for my confusion.

yakutovicha:
> Your PR makes it seem like we need the second one. I think that's the source for my confusion.

@csadorf, added a comment about `pybel` module. Hope there is no confusion now.

PR #58, opened 2020-02-24
Structure viewer: allow to display supercell.

yakutovicha:
This involves adding new displayed_structure trait that will be
pointing to the Atoms object that is actually displayed.


yakutovicha:
@ltalirz this is to complete the work on adding supercell representation to structure viewer.

PR #62, opened 2020-03-06
Improve structure-related widgets:

yakutovicha:
* Set all widgets to use 'structure' trait that can be linked among
structure-managers.

StructureManagerWidget:
* `structure` is now trait that can be set either using Atoms object or
using AiiDA node
* `structure_node` is now read_only trait that contains AiiDA node of
selected type
* `node_class` is now trait
* freeze() function is now removed as it is not used and not really
needed
* `has_structure` and `frozen` traits were removed.


aiidalab_widgets_base/structures.py
```diff
@@ -4,126 +4,113 @@
...
-        self.structure_ase = None
-        self._structure_node = None
+        # Store button.
+        self.btn_store = ipw.Button(description='Store in AiiDA', disabled=True)
+        self.btn_store.on_click(self._on_click_store)
+
+        # Setting traits' initial values
+        self.structure = None
+        self._inserted_structure = None
+        self.set_trait('structure_node', None)
```
csadorf: I think it is best practice to use [`@default` functions](https://traitlets.readthedocs.io/en/stable/using_traitlets.html#default-values-and-checking-type-and-value) for that.

aiidalab_widgets_base/structures.py
```diff
@@ -4,126 +4,113 @@
...
+    '''
+
+    structure = Union([Instance(Atoms), Instance(Data)], allow_none=True)
+    structure_node = Instance(Data, allow_none=True, read_only=True)
+    node_class = Unicode()
 
-    has_structure = Bool(False)
-    frozen = Bool(False)
-    DATA_FORMATS = ('StructureData', 'CifData')
+    DATA_FORMATS = {'CifData': 'cif', 'StructureData': 'structure'}
```
csadorf: Should this be `SUPPORTED_DATA_FORMATS`?

aiidalab_widgets_base/structures.py
```diff
@@ -136,49 +123,68 @@ def store_structure(self, label=None, description=None):
         self.structure_node.store()
         print("Stored in AiiDA: " + repr(self.structure_node))
```
csadorf: I thin it's a bad idea to report to the user using print-statements as part of the widget implementation.

aiidalab_widgets_base/structures.py
```diff
@@ -136,49 +123,68 @@ def store_structure(self, label=None, description=None):
...
+    @observe('node_class')
+    def _change_structure_node(self, change):
+        if change['new'] is not None:
+            self.structure = self._inserted_structure
+
+    @validate('structure')
+    def _valid_structure(self, change):
+        """Returns ASE atoms object and sets structure_node trait."""
+
+        self._inserted_structure = change['value']
```
csadorf: This function should not have any side-effects besides transforming the value and returning it.

aiidalab_widgets_base/structures.py
```diff
@@ -4,126 +4,113 @@
...
-    :vartype frozen: bool
-    :ivar structure_node: link to AiiDA structure object
-    :vartype structure_node: StructureData or CifData'''
+    Attributes:
+        structure(Atoms): trait that contains the selected structure. 'None' if no structure is selected.
+        structure_node(StructureData, CifData): trait that contains AiiDA structure object
+        node_class(str): trait that contains structure_node type (as string).
+    '''
+
+    structure = Union([Instance(Atoms), Instance(Data)], allow_none=True)
```
csadorf: If I understand this correctly, the intention here is to be able to connect this traitlet to either an `Atoms` instance or a `StructureData` instance, correct? Why not be consistent, require one of the two types and then let the user use `dlinks` instead of `link` to connect to this traitlet. The two data types are very similar and by allowing the traitlet to be of either type, it is going to be very difficult to use it in practice.

aiidalab_widgets_base/structures.py
```diff
@@ -287,8 +289,8 @@ def search(self, change=None):  # pylint: disable=unused-argument
             self.start_date = datetime.datetime.strptime(self.date_start.value, '%Y-%m-%d')
             self.end_date = datetime.datetime.strptime(self.date_end.value, '%Y-%m-%d') + datetime.timedelta(hours=24)
         except ValueError:  # Otherwise revert to the standard (i.e. last 7 days)
-            self.start_date = self.dt_end
-            self.end_date = self.dt_now + datetime.timedelta(hours=24)
+            self.start_date = datetime.datetime.now() - datetime.timedelta(days=10)
```
csadorf: Why are 10 days hard-coded here?

aiidalab_widgets_base/structures.py
```diff
@@ -339,19 +339,12 @@ def search(self, change=None):  # pylint: disable=unused-argument
...
-        if not self.results.value:
-            return
-        structure_ase = self.results.value.get_ase()
-        formula = structure_ase.get_chemical_formula()
-        if self.on_structure_selection is not None:
-            self.on_structure_selection(structure_ase=structure_ase, name=formula)
-
-    def on_structure_selection(self, structure_ase, name):
-        pass
+        self.structure = None if not self.results.value else self.results.value
```
csadorf:
```suggestion
        self.structure = self.results.value or None
```

aiidalab_widgets_base/structures.py
yakutovicha: agreed.

aiidalab_widgets_base/structures.py
yakutovicha: agreed. Put the output into `ipw.HTML`.

aiidalab_widgets_base/structures.py
yakutovicha: Indeed. When I set the value to `None` it is not even called. So I agree that all 'side-effects' should go into the observed methods.

aiidalab_widgets_base/structures.py
yakutovicha: Just a random default..

aiidalab_widgets_base/structures.py
yakutovicha: thanks!

aiidalab_widgets_base/structures.py
yakutovicha: thanks!

aiidalab_widgets_base/structures.py
yakutovicha: Please have a look into c91d1d3, not sure if it is an elegant solution, though. It turned out that 

1. ase1 and ase2 are not always the same after the following conversion:
ase1 -> Cif/Structure -> ase2

1. Cif/Structure are **never** the same after convering into `Atoms` and back.

So, in the end, I am not totally convinced that c91d1d3 is a good thing to do.

aiidalab_widgets_base/databases.py
```diff
@@ -105,3 +105,7 @@ def _on_select_structure(self, change):
         self.structure = selected['cif'].get_ase()
         struct_url = selected['url'].split('.cif')[0] + '.html'
         self.link.value = '<a href="{}" target="_blank">COD entry {}</a>'.format(struct_url, selected['id'])
+
+    @default('structure')
+    def _default_structure(self):  # pylint: disable=no-self-use
```
csadorf: I believe that if you have `allow_none=True`, then this is not necessary, but there is probably also no harm in keeping this.

aiidalab_widgets_base/structures.py
```diff
@@ -27,13 +29,13 @@ class StructureManagerWidget(ipw.VBox):
...
-    structure_node = Instance(Data, allow_none=True, read_only=True)
+    structure = Instance(Atoms, allow_none=True)
+    structure_node = Instance(Data, allow_none=True)
     node_class = Unicode()
 
-    DATA_FORMATS = {'CifData': 'cif', 'StructureData': 'structure'}
+    SUPPORTED_DATA_FORMATS = {'CifData': 'cif', 'StructureData': 'structure'}
 
-    def __init__(self, importers, storable=True, node_class=None, **kwargs):
+    def __init__(self, importers, storable=True, node_class=None, **kwargs):  # pylint: disable=too-many-branches
```
csadorf: Any way we could refactor the class structure to reduce the number of branches and thus complexity?

aiidalab_widgets_base/structures.py
```diff
@@ -95,15 +102,19 @@ def __init__(self, importers, storable=True, node_class=None, **kwargs):
             store_and_description.append(self.btn_store)
         if node_class is None:
             store_and_description.append(data_format)
-        elif node_class not in self.DATA_FORMATS:
-            raise ValueError("Unknown data format '{}'. Options: {}".format(node_class, list(self.DATA_FORMATS.keys())))
+        elif node_class not in self.SUPPORTED_DATA_FORMATS:
```
csadorf: proposed refactoring:
```
elif node_class in self.SUPPORTED_DATA_FORMATS:
   ...
else:
    raise ValueError(...)
```

aiidalab_widgets_base/structures.py
```diff
@@ -114,68 +125,84 @@ def store_structure(self, label=None, description=None):
...
-        print("Stored in AiiDA: " + repr(self.structure_node))
+        self.output.value = "Stored in AiiDA [{}]".format(self.structure_node)
 
-    @observe('node_class')
-    def _change_structure_node(self, change):
-        if change['new'] is not None:
-            self.structure = self._inserted_structure
+    @default('node_class')
+    def _default_node_class(self):
+        return 'StructureData'
```
csadorf: Is it really necessary to support multiple node classes?

aiidalab_widgets_base/structures.py
```diff
@@ -114,68 +125,84 @@ def store_structure(self, label=None, description=None):
...
+            with self.hold_trait_notifications():
+                self.structure = structure_ase
 
-        # If no structure is provided, the rest of the code can be skipped
-        if self._inserted_structure is None:
-            return None
+    @observe('structure')
+    def _update_structure_node(self, change=None):
+        """This function enables/disables `btn_store` widget if structure is provided/set to None.
+        Also, the function sets `structure_node` trait to the selected node type."""
```
csadorf: This doc-string is non standard. I'm surprised the linter didn't complain.

aiidalab_widgets_base/structures.py
csadorf: This was discussed offline.

aiidalab_widgets_base/structures.py
csadorf: What are the exact implications? Why not make that a modifiable traitlet and display it in the UI?

aiidalab_widgets_base/structures.py
yakutovicha: This was only to test having modifiable both: `structure` and `structure_node` (02f5b206d2536b600da59736c1cf9329527489ae). As we agreed to revert that commit - things have become easier again.

aiidalab_widgets_base/structures.py
yakutovicha: agreed.

aiidalab_widgets_base/structures.py
yakutovicha: I started to answer your comment, but then I realized that I had in mind many things to improve. So that is what I did in 5df97c2. I think now things should be more clear.

Answering to your questions: the widgets `self.start_date_widget` and `self.end_date_widget` are there and displayed.

yakutovicha: So what is happening here: if the user provides the date in a wrong format, it will be replaced by one week range

aiidalab_widgets_base/structures.py
yakutovicha: I don't see a reason why not doing this: if there are different structure types in AiiDA, we should support them.

aiidalab_widgets_base/databases.py
yakutovicha: ok, so we keep it, right?

aiidalab_widgets_base/structures.py
yakutovicha: How should it look like? I don't know what is the problem here.

yakutovicha:
@csadorf, I played around with this idea of converting `structure` trait into `structure_node` any time one accesses `structure` trait. And I came to the conclusion that we should not use this approach:
```
    def __getattribute__(self, name):
        if name == 'foo':
            self._on_foo_accessed()
        return super().__getattribute__(name)
```

This is why I think we should not do it:

1. The first and obvious reason - it is just too many heavy transformations that are kind of useless.

1. Second - the whole approach of traitlets that we are relying on is becoming pointless.  The `structure` trait that we sync with all connected widgets will only detect the modification if one does <obj>.structure = something.

1. It won't harm if to modify the structure we require to do `<obj>.structure = new_structure`

aiidalab_widgets_base/structures.py
csadorf: Well, I'm asking because whenever we introduce something like this all downstream code must perform checks about the data type they receive and then change behavior. So that would be a reason "why not".

aiidalab_widgets_base/structures.py
csadorf: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

csadorf:
> @csadorf, I played around with this idea of converting `structure` trait into `structure_node` any time one accesses `structure` trait. And I came to the conclusion that we should not use this approach:
> 
> ```
>     def __getattribute__(self, name):
>         if name == 'foo':
>             self._on_foo_accessed()
>         return super().__getattribute__(name)
> ```
> 
> This is why I think we should not do it:
> 
> 1. The first and obvious reason - it is just too many heavy transformations that are kind of useless.
> 2. Second - the whole approach of traitlets that we are relying on is becoming pointless.  The `structure` trait that we sync with all connected widgets will only detect the modification if one does .structure = something.
> 3. It won't harm if to modify the structure we require to do `<obj>.structure = new_structure`

That's perfectly fine with me. I was not really happy with this either.

yakutovicha:
> That's perfectly fine with me. I was not really happy with this either.

Please review then. I would like to finalize the PR and proceed with the aiidalab container update.

aiidalab_widgets_base/structures.py
```diff
@@ -1,184 +1,206 @@
...
-            self._structure_node.description = self.structure_description.value
-            self._structure_node.label = self.structure_ase.get_chemical_formula()
-        return self._structure_node
+
+        self.btn_store.disabled = False
+
+        # Chosing type for the structure node.
+        StructureNode = DataFactory(self.SUPPORTED_DATA_FORMATS[self.node_class])  # pylint: disable=invalid-name
+
+        # If structure trait is set to Atoms object, structure node must be created from it.
```
csadorf: I always thought that these kind of conversions are supposed to be handled in the "validate" functions, see the API documentation: https://traitlets.readthedocs.io/en/stable/defining_traits.html#traitlets.MyTrait.validate

> Check whether a given value is valid. If it is, it should return the value **(coerced to the desired type, if necessary)**[emphasis mine].

Either way is fine with me.

aiidalab_widgets_base/structures.py
```diff
@@ -193,12 +215,16 @@ def _on_file_upload(self, change):  # pylint: disable=unused-argument
         self.file_path = os.path.join(tempfile.mkdtemp(), self.file_upload.filename)
         with open(self.file_path, 'w') as fobj:
             fobj.write(self.file_upload.data.decode("utf-8"))
-        structure_ase = get_ase_from_file(self.file_path)
-        self.on_structure_selection(structure_ase=structure_ase, name=self.file_upload.filename)
+        self.structure = get_ase_from_file(self.file_path)
+
+    @default('structure')
+    def _default_structure(self):
+        return None
```
csadorf: I'm not sure if this is necessary, but fine to leave.

aiidalab_widgets_base/structures.py
```diff
@@ -215,143 +241,160 @@ def get_example_structures(examples):
 
     def _on_select_structure(self, change):  # pylint: disable=unused-argument
         """When structure is selected."""
-        if not self._select_structure.value:
-            return
-        structure_ase = get_ase_from_file(self._select_structure.value)
-        self.on_structure_selection(structure_ase=structure_ase, name=self._select_structure.label)
+
+        self.structure = None if not self._select_structure.value else get_ase_from_file(self._select_structure.value)
```
csadorf: I'd prefer the positive equivalent:
```suggestion
        self.structure = get_ase_from_file(self._select_structure.value) if self._select_structure.value else None
```

aiidalab_widgets_base/structures.py
```diff
@@ -215,143 +241,160 @@ def get_example_structures(examples):
...
-            [self.date_text, ipw.HBox([self.date_start, self.date_end]), self.btn_date],
-            layout={
-                'border': '1px solid #fafafa',
-                'padding': '1em'
-            })
 
-        # Labels
-        self.drop_label = ipw.Dropdown(options=({'All'}.union(process_labels)),
+        # Structure objects we want to query for.
+        self.what_query_for = (DataFactory('structure'), DataFactory('cif'))
```
csadorf: I propose to rename this attribute to "query_structure_type" or so. The current name is not very meaningful.

aiidalab_widgets_base/structures.py
```diff
@@ -215,143 +241,160 @@ def get_example_structures(examples):
...
+    def preprocess(self):
+        """Search structures in AiiDA database and add formula extra to them."""
 
-    @staticmethod
-    def preprocess():
-        """Search structures in AiiDA database."""
         queryb = QueryBuilder()
-        queryb.append(StructureData, filters={'extras': {'!has_key': 'formula'}})
+        queryb.append(self.what_query_for, filters={'extras': {'!has_key': 'formula'}})
         for itm in queryb.all():  # iterall() would interfere with set_extra()
```
csadorf: No need to abbreviate "item" here.

aiidalab_widgets_base/structures.py
yakutovicha: I think the reason for having it possible to select the output structure type is exactly to prevent downstream code of doing any conversions. The widget allows setting up the output node type prior to its usage.

To give a concrete example: "a downstream code" can be a workflow. When setting up a workflow user knows already which input structure type is required there (say `CifData`). So using the current functionality an app developer can already select `structure_node` of which type will be produced by the widget.

aiidalab_widgets_base/structures.py
yakutovicha: fixed.

aiidalab_widgets_base/structures.py
yakutovicha: ok. I will keep.

aiidalab_widgets_base/structures.py
yakutovicha: Ok.

aiidalab_widgets_base/structures.py
yakutovicha: Thanks for the suggestion. I could not come up with a good name here.

yakutovicha: done

aiidalab_widgets_base/structures.py
yakutovicha: thanks.

aiidalab_widgets_base/structures.py
yakutovicha: Sure, if I would set `structure` trait, I would do it in `validate` function. However, here I am setting `structure_node` trait value.

PR #63, opened 2020-03-10
Work on process widgets:

yakutovicha:
* `process.py`: move all the inputs to the top of the module

* SubmitButtonWidget:
  * Rename widgets_values argument of `__init__()` function to
    `input_dictionary_function`. Also add `message` argument that
    will be displayed on the submit button.
  * Add `run_after_submitted()` method that registers functions to
    be run after the process has been submitted.
  * Print output messages to `HTML` widget instead of `with Output:`.
  * Do not submit process when `input_dictionary_function` returned
     `None`.

* ProcessFollowerWidget:
  * Add `run_after_completed()` method to register functions that will
    be run after the followed process is completed.

### DIFF
--- a/aiidalab_widgets_base/process.py
+++ b/aiidalab_widgets_base/process.py
@@ -1,14 +1,16 @@
 """Widgets to work with processes."""
 
 import os
-from ipywidgets import Textarea, VBox
+from ipywidgets import Button, HTML, IntProgress, Layout, Textarea, VBox
+
+# AiiDA imports
+from aiida.engine import submit
+from aiida.orm import CalcJobNode, ProcessNode, WorkChainNode
 
 
 def get_running_calcs(process):
     """Takes a process and returns a list of running calculations. The running calculations
     can be either the process itself or its running children."""
-
-    from aiida.orm import CalcJobNode, ProcessNode, WorkChainNode
 
     # If a process is a running calculation - returning it
     if issubclass(type(process), CalcJobNode) and not process.is_sealed:
@@ -28,19 +30,20 @@ def get_running_calcs(process):
 class SubmitButtonWidget(VBox):
     """Submit button class that creates submit button jupyter widget."""
 
-    def __init__(self, process, widgets_values):
+    def __init__(self, process, input_dictionary_function, description="Submit"):
         """Submit Button
         :process: work chain to run
         :param_funtion: the function that generates input parameters dictionary
         """
-        from ipywidgets import Button, Output
 
         self.process = None
         self._process_class = process
-        self.widgets_values = widgets_values
-        self.btn_submit = Button(description="Submit", disabled=False)
+        self._run_after_submitted = []
+
+        self.input_dictionary_function = input_dictionary_function
+        self.btn_submit = Button(description=description, disabled=False)
         self.btn_submit.on_click(self.on_btn_submit_press)
-        self.submit_out = Output()
+        self.submit_out = HTML('')
         children = [
             self.btn_submit,
             self.submit_out,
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/process.py
```diff
@@ -52,28 +55,32 @@ def on_click(self, function):
...
             self.process = submit(self._process_class, **input_dict)
-            print("Submitted process {}".format(self.process))
-            return
+            self.submit_out.value = "Submitted process {}".format(self.process)
+            for func in self._run_after_submitted:
+                func(self.process)
+        else:
+            self.submit_out.value = "SubmitButtonWidget: did not recieve input dictionary."
+
+    def run_after_submitted(self, function):
```
csadorf: I think it would be better to use the standard way of implementing such callbacks here:
```
def _on_submitted(self):
    pass
```
The user can then override this function. If they want to register multiple functions, they can do that by creating a meta-function that calls the other functions.

aiidalab_widgets_base/process.py
```diff
@@ -102,13 +109,19 @@ def follow(self, detach=False):
             update_state.start()
         else:
             self._follow()
+            for func in self._run_after_completed:
+                func(self.process)
+
+    def run_after_completed(self, function):
```
csadorf: Same as my comment about `run_after_submitted()`.

aiidalab_widgets_base/process.py
```diff
@@ -52,28 +55,32 @@ def on_click(self, function):
...
             self.btn_submit.disabled = True
-            input_dict = self.widgets_values()
             self.process = submit(self._process_class, **input_dict)
-            print("Submitted process {}".format(self.process))
-            return
+            self.submit_out.value = "Submitted process {}".format(self.process)
+            for func in self._run_after_submitted:
+                func(self.process)
+        else:
+            self.submit_out.value = "SubmitButtonWidget: did not recieve input dictionary."
```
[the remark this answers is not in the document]
yakutovicha: Yes, while generating input using widgets, a user might forget things like specifying a code or a structure. In that case, `input_dictionary_function` will return `None`. And this is fine, I think.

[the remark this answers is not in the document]
csadorf: Ok.

aiidalab_widgets_base/process.py
yakutovicha: Here I was doing something similar to what is done for the `Button` widget.
```
from ipywidgets import Button
b = Button()
b.on_click(lambda _: print("hello world"))
b.on_click(lambda _: print("test"))
display(b)
```
Don't you like this approach?

aiidalab_widgets_base/process.py
```diff
@@ -52,28 +55,32 @@ def on_click(self, function):
...
     def on_btn_submit_press(self, _):
         """When submit button is pressed."""
-        from IPython.display import clear_output
-        from aiida.engine import submit
-
-        with self.submit_out:
-            clear_output()
+        self.submit_out.value = ''
+        input_dict = self.input_dictionary_function()
+        if input_dict is not None:
```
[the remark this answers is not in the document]
yakutovicha: agreed.

aiidalab_widgets_base/process.py
csadorf: My bad I did not remember correctly how the `Button.on_click()` is implemented. Please keep it *as-is*.

csadorf: The only thing I might suggest is renaming the function to `_on_submitted` or `on_submitted`.

aiidalab_widgets_base/process.py
```diff
@@ -102,13 +109,19 @@ def follow(self, detach=False):
             update_state.start()
         else:
             self._follow()
+            for func in self._run_after_completed:
+                func(self.process)
+
+    def run_after_completed(self, function):
+        """Run functions after a process has been completed."""
+        if callable(function):
+            self._run_after_completed.append(function)
```
csadorf: I don't think it necessary to check the argument type here. If the user provides something that is not callable, they will notice.

aiidalab_widgets_base/process.py
csadorf: Did you not like my suggestion of following the `on_<event>` pattern for registering callbacks?

aiidalab_widgets_base/process.py
yakutovicha: agreed.

aiidalab_widgets_base/process.py
yakutovicha: I loved it :)

PR #68, opened 2020-03-24
CodeDropdown: hidden codes and codes on disabled computers.

yakutovicha:
This PR adds two paramets to the __init__() method:
hidden_codes: to look for hidden codes or not.
disabled_computers: to look for disabled_computers or not.

The code is modified to handle those parameters.

### DIFF
--- a/aiidalab_widgets_base/codes.py
+++ b/aiidalab_widgets_base/codes.py
@@ -4,7 +4,7 @@ from subprocess import check_output
 
 import ipywidgets as ipw
 from IPython.display import clear_output
-from traitlets import Dict, Instance, Unicode, Union, link, validate
+from traitlets import Bool, Dict, Instance, Unicode, Union, link, observe, validate
 
 from aiida.orm import Code, QueryBuilder, User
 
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/codes.py
```diff
@@ -33,43 +33,38 @@ class CodeDropdown(ipw.VBox):
...
+    allow_disabled_computers = Bool(False)
 
-    def __init__(self,
-                 input_plugin,
-                 text='Select code:',
-                 hidden_codes=False,
-                 disabled_computers=False,
-                 path_to_root='../',
-                 **kwargs):
+    def __init__(self, input_plugin, text='Select code:', path_to_root='../', **kwargs):
```
csadorf, line 39: The `text` parameter is usually described as `description` for most ipywidgets.

aiidalab_widgets_base/codes.py
```diff
@@ -33,43 +33,38 @@ class CodeDropdown(ipw.VBox):
...
 
         self.input_plugin = input_plugin
-        self.hidden_codes = hidden_codes
-        self.disabled_computers = disabled_computers
-        self.output = ipw.Output()
+
+        if 'allow_hidden_codes' in kwargs:
+            self.allow_hidden_codes = kwargs['allow_hidden_codes']
+        if 'allow_disabled_computers' in kwargs:
+            self.allow_disabled_computers = kwargs['allow_disabled_computers']
```
csadorf, line 53: What's the motivation for sneakily setting these traits as part of the object construction?

aiidalab_widgets_base/codes.py
```diff
@@ -78,38 +73,26 @@ def __init__(self,
...
+            if c[0].computer.is_user_configured(user) and (self.allow_hidden_codes or not c[0].hidden) and
+            (self.allow_disabled_computers or c[0].computer.is_user_enabled(user))
         }
 
     @staticmethod
     def _full_code_label(code):
         return "{}@{}".format(code.label, code.computer.name)
 
+    @observe('allow_disabled_computers')
+    def _observe_allow_disabl_comp(self, _=None):
```
csadorf, line 92:
```suggestion
    def _observe_allow_disabled_computers(self, _=None):
```
Better to be consistent.

aiidalab_widgets_base/codes.py
```diff
@@ -78,38 +73,26 @@ def __init__(self,
...
+        self.refresh()
+
+    @observe('allow_hidden_codes')
```
csadorf, line 95: I'd suggest to add the callback as part of the constructor, together with `btn_refresh`.

aiidalab_widgets_base/codes.py
yakutovicha: Ok

aiidalab_widgets_base/codes.py
yakutovicha: What do you mean by "sneakily setting"? I see nothing sneaking here, honestly.

Moreover, I thin in ipywidgets similar approach is implemented:

```
import ipywidgets as ipw
s = ipw.Text(
    description="AiiDA code label:",         
    style={"description_width": "initial"}
)
display(s)


s.style = {"description_width": "200px"}
display(s)
```

aiidalab_widgets_base/codes.py
yakutovicha: ok

aiidalab_widgets_base/codes.py
yakutovicha: Right, I did it for the `_observe_allow_disabl_comp` function as well.

yakutovicha:
@csadorf I addressed your comments either by changing the code or by leaving a comment. 

One thing I wanted to double-check is the names of `allow_hidden_codes` and `allow_disabled_computers` traits. Should we keep these names, or do you have a better idea?

As pointed out in [this comment](https://github.com/aiidalab/aiidalab-widgets-base/pull/69#issuecomment-610357405), do you think it is a problem that here we use `allow_disabled_computers` while in `ComputerDropdown` we use `allow_select_disabled`. Both of them are actually for deciding whether to include disabled computers in the query.

aiidalab_widgets_base/codes.py
yakutovicha: As we discussed offline, those lines are not needed at all. Traits get set automatically through the `super().__init__(**kwars)` call.

PR #69, opened 2020-03-24
ComputerDropdown: manage disabled computers.

yakutovicha:
Add `disabled_computers` parameter to __init__() method that defines
whether to show or not the disabled computers.

### DIFF
--- a/aiidalab_widgets_base/computers.py
+++ b/aiidalab_widgets_base/computers.py
@@ -691,14 +691,14 @@ class ComputerDropdown(ipw.VBox):
         link((self._dropdown, 'value'), (self, 'selected_computer'))
 
         btn_refresh = ipw.Button(description="Refresh", layout=ipw.Layout(width="70px"))
-        btn_refresh.on_click(self._refresh)
+        btn_refresh.on_click(self.refresh)
 
         self._setup_another = ipw.HTML(
             value="""<a href={path_to_root}aiidalab-widgets-base/setup_computer.ipynb target="_blank">
             Setup new computer</a>""".format(path_to_root=path_to_root))
 
         children = [ipw.HBox([self._dropdown, btn_refresh, self._setup_another]), self.output]
-        self._refresh()
+        self.refresh()
         super().__init__(children=children, **kwargs)
 
     def _get_computers(self):
@@ -714,7 +714,7 @@ class ComputerDropdown(ipw.VBox):
         }
 
     @observe('allow_select_disabled')
-    def _refresh(self, _=None):
+    def refresh(self, _=None):
         """Refresh the list of configured computers."""
 
         with self.output:
--- a/aiidalab_widgets_base/computers.py
+++ b/aiidalab_widgets_base/computers.py
@@ -7,7 +7,7 @@ from subprocess import check_output, call
 import pexpect
 import ipywidgets as ipw
 from IPython.display import clear_output
-from traitlets import Dict, Instance, Int, Unicode, Union, validate, link
+from traitlets import Bool, Dict, Instance, Int, Unicode, Union, link, observe, validate
 
 from aiida.common import NotExistent
 from aiida.orm import Computer, QueryBuilder, User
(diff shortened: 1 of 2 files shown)


aiidalab_widgets_base/computers.py
```diff
@@ -668,13 +668,18 @@ class ComputerDropdown(ipw.VBox):
     selected_computer = Union([Unicode(), Instance(Computer)], allow_none=True)
     computers = Dict(allow_none=True)
 
-    def __init__(self, text='Select computer:', path_to_root='../', **kwargs):
+    def __init__(self, text='Select computer:', disabled_computers=False, path_to_root='../', **kwargs):
```
csadorf: Rather than making this a constructor argument, I'd propose to make this a `Bool` traitlet so that users can connect it to a checkbox etc.

I also propose to rename this to `allow_select_disabled` for clarity.

aiidalab_widgets_base/computers.py
```diff
@@ -693,14 +698,18 @@ def __init__(self, text='Select computer:', path_to_root='../', **kwargs):
         self._refresh()
         super().__init__(children=children, **kwargs)
 
-    @staticmethod
-    def _get_computers():
+    def _get_computers(self):
         """Get the list of available computers."""
-        query_b = QueryBuilder()
-        query_b.append(Computer, tag='computer')
+        querybuild = QueryBuilder()
```
csadorf: While I'm generally in favor of very explicit (and therefore usually long) variable names, this is a moment, where I'd prefer a short one to make the following code more readable.
```suggestion
        qb = QueryBuilder()
```

aiidalab_widgets_base/computers.py
```diff
@@ -693,14 +698,18 @@ def __init__(self, text='Select computer:', path_to_root='../', **kwargs):
...
+        querybuild.append(Computer, tag='computer')
+
+        if self.disabled_computers:
+            to_return = [c[0] for c in querybuild.all()]
+        else:
+            to_return = [c[0] for c in querybuild.all() if c[0].is_user_enabled(User.objects.get_default())]
 
         # Only computers configured for the current user.
-        return {c[0].name: c[0] for c in query_b.all() if c[0].is_user_configured(User.objects.get_default())}
+        return {c.name: c for c in to_return if c.is_user_configured(User.objects.get_default())}
```
csadorf: I think we can simplify this segment a little bit by converting this into one comprehension (with appropriate formatting of course):
```python
user = User.objects.get_default()
return {
    c[0].name: c[0] for c in query_b.all()
    if c[0].is_user_configured(user) and (self.disabled_computers or c[0].is_user_enabled(user)}
```

aiidalab_widgets_base/computers.py
yakutovicha: I made it a trait, but I would still prefer to allow providing the value of the trait through the constructor argument. So I added the following:

```
        if 'allow_select_disabled' in kwargs:
            self.allow_select_disabled = kwargs['allow_select_disabled']
```

aiidalab_widgets_base/computers.py
yakutovicha: agreed.

aiidalab_widgets_base/computers.py
yakutovicha: thanks, I did that.

yakutovicha:
thanks for the nice suggestions, @csadorf. I think the code became better.

yakutovicha:
@csadorf, before merging, I wanted to hear your opinion about the name of `allow_select_disabled` trait.

In the `CodeSelection` widget we use `allow_disabled_computers` which essentially plays the same role. Do you think it would be better to use the same name here?

csadorf:
> @csadorf, before merging, I wanted to hear your opinion about the name of `allow_select_disabled` trait.
> 
> In the `CodeSelection` widget we use `allow_disabled_computers` which essentially plays the same role. Do you think it would be better to use the same name here?

I think that `allow_select_disabled` is preferable over `allow_disabled_computers`, because it makes it clear that this is something that affects and is handled on the UI level. I suggest we just adjust the other name as well.

I'd also recommend to follow the ipywidgets standard of naming the primary "value" of a widget, actually `value`, since the type is already specified through the class name. For example, the "value" of an `IntSlider` is not `number` or `integer_number`, but `value`. 

In this case I believe that the "value" would correspond to the `selected_computer` trait. Since this is a `Dropdown` widget, I'd propose the following naming scheme:
 - `selected_computer` -> `value`
 - `computers` -> `options`

Feel free to make those adjustments in a separate PR if you agree.

PR #70, opened 2020-03-24
SubmitButtonWidget and ProcessFollowerWidget:

yakutovicha:
SubmitButtonWidget
* Configure whether to disable submit button after the process
was submitted.
* Configure whether to append or overwrite the widget output

ProcessFollowerWidget
* Run functions registered with on_completed method if the process
is followed asyncronosly.

### DIFF
--- a/aiidalab_widgets_base/process.py
+++ b/aiidalab_widgets_base/process.py
@@ -2,6 +2,7 @@
 
 import os
 from ipywidgets import Button, HTML, IntProgress, Layout, Textarea, VBox
+from traitlets import Instance
 
 # AiiDA imports
 from aiida.engine import submit
@@ -29,6 +30,7 @@ def get_running_calcs(process):
 
 class SubmitButtonWidget(VBox):
     """Submit button class that creates submit button jupyter widget."""
+    process = Instance(ProcessNode, allow_none=True)
 
     def __init__(self,
                  process_class,
@@ -49,7 +51,6 @@ class SubmitButtonWidget(VBox):
         append_output (bool): Whether to clear widget output for each subsequent submission.
         """
 
-        self.process = None
         self.disable_after_submit = disable_after_submit
         self.append_output = append_output
         self._process_class = process_class
--- a/aiidalab_widgets_base/process.py
+++ b/aiidalab_widgets_base/process.py
@@ -31,14 +31,14 @@ class SubmitButtonWidget(VBox):
     """Submit button class that creates submit button jupyter widget."""
 
     def __init__(self,
-                 process,
+                 process_class,
                  input_dictionary_function,
                  description="Submit",
                  disable_after_submit=True,
                  append_output=False):
         """Submit Button widget.
 
-        process (Process): Process to submit.
+        process_class (Process): Process class to submit.
 
         input_dictionary_function (func): Function that generates input parameters dictionary.
 
@@ -52,7 +52,7 @@ class SubmitButtonWidget(VBox):
         self.process = None
         self.disable_after_submit = disable_after_submit
         self.append_output = append_output
-        self._process_class = process
+        self._process_class = process_class
         self._run_after_submitted = []
 
         self.input_dictionary_function = input_dictionary_function
@@ -93,7 +93,7 @@ class SubmitButtonWidget(VBox):
                 func(self.process)
 
     def on_submitted(self, function):
-        """Run functions after a process has been submitted sucesfully."""
+        """Run functions after a process has been submitted successfully."""
         self._run_after_submitted.append(function)
 
 
--- a/aiidalab_widgets_base/process.py
+++ b/aiidalab_widgets_base/process.py
@@ -1,11 +1,12 @@
 """Widgets to work with processes."""
 
 import os
+from inspect import isclass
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/process.py
```diff
@@ -30,13 +30,28 @@ def get_running_calcs(process):
...
+
+        disable_after_submit (bool): Whether to disable the button after the process was submitted.
+
+        append_output (bool): Whether to clean the widget output before every submission.
         """
 
         self.process = None
+        self.disable_after_submit = disable_after_submit
+        self.append_output = append_output
         self._process_class = process
```
csadorf: It's confusing that the constructor argument is called `process`, but the corresponding class attribute is called `_process_class` and not even accessible.

aiidalab_widgets_base/process.py
```diff
@@ -30,13 +30,28 @@ def get_running_calcs(process):
...
+
+        process (Process): Process to submit.
+
+        input_dictionary_function (func): Function that generates input parameters dictionary.
+
+        description (str): Description written on the submission button.
+
+        disable_after_submit (bool): Whether to disable the button after the process was submitted.
+
+        append_output (bool): Whether to clean the widget output before every submission.
```
csadorf:
```suggestion
        append_output (bool): Whether to clear widget output for each subsequent submission.
```

aiidalab_widgets_base/process.py
```diff
@@ -30,13 +30,28 @@ def get_running_calcs(process):
...
         """
 
         self.process = None
```
csadorf: Since it is now possible to submit multiple times. What is the expected behavior here? We override this attribute and lose track of previous processes?

aiidalab_widgets_base/process.py
yakutovicha: you are right, the constructor argument should be named `process_class` for consistency

aiidalab_widgets_base/process.py
yakutovicha: Yes, I think it would even make sense for this guy to become a trait. This way the change can be detected and the process can be stored elsewhere.

What do you think?

yakutovicha: Made it in c301acdee3fc85686943f366e49f2d8ee0fb0d95

yakutovicha:
> From a first look this seems fine (with a few edit suggestions). What I'm more wondering is how much sense it makes to allow for subsequent submissions. I feel like it would be important from a UI perspective that one needs to explicitly cancel or clear a running process before a user can submit a new one. I'd like to have a brief discussion about this before approving this.

That is true in most of the cases, but not always. For example, I have a notebook to submit isotherm simulations. Every time I submit a simulation I specify quite a significant number of parameters (up to 10). But what if I want to do exactly the same simulation, but for different temperatures. In that case, it is much handier to change only one parameter and hit `submit` instead of redefining everything again before being able to do that.

aiidalab_widgets_base/process.py
```diff
@@ -55,19 +71,30 @@ def on_click(self, function):
...
         input_dict = self.input_dictionary_function()
         if input_dict is None:
-            self.submit_out.value = "SubmitButtonWidget: did not recieve input dictionary."
+            if self.append_output:
+                self.submit_out.value += "SubmitButtonWidget: did not recieve input dictionary.<br>"
+            else:
+                self.submit_out.value = "SubmitButtonWidget: did not recieve input dictionary."
         else:
-            self.btn_submit.disabled = True
+            self.btn_submit.disabled = self.disable_after_submit
```
csadorf, line 85: Are there any other states in this widget that would cause the button to be disabled?

yakutovicha: No, I don't think so. It can only be disabled in the case of a successful submission, which is right here.

An additional possibility would be to have`process_class` set to `None` which is currently allowed, but shouldn't. I should make a check for that, I think.

csadorf: Ok, that's along the lines of what I was thinking. I think we must either disallow `None` or add the additional check here (something like this):
```suggestion
            self.btn_submit.disabled = self.process_node is None or self.disable_after_submit
```

yakutovicha: I was thinking of adding a check to `process_class`. One can just verify that it is a subclass of `Process`. This will automatically take care of not allowing the `process_class` to be `None`.

yakutovicha: done

aiidalab_widgets_base/process.py
```diff
@@ -51,19 +52,31 @@ def __init__(self,
...
-        self.input_dictionary_function = input_dictionary_function
         self.btn_submit = Button(description=description, disabled=False)
         self.btn_submit.on_click(self.on_btn_submit_press)
         self.submit_out = HTML('')
         children = [
             self.btn_submit,
             self.submit_out,
         ]
+
+        self._run_after_submitted = []
```
csadorf, line 78: Do you have the corresponding registration function?

yakutovicha: yes, line 109, or what do you mean?

csadorf: Somehow missed that.

PR #90, opened 2020-05-18
Single process widgets

yakutovicha:


aiidalab_widgets_base/__init__.py
```diff
@@ -10,7 +10,8 @@
 from .databases import CodQueryWidget  # noqa
 #from .editors import editor  # noqa
 from .export import ExportButtonWidget  # noqa
-from .process import ProcessFollowerWidget, ProgressBarWidget, RunningCalcJobOutputWidget, SubmitButtonWidget  # noqa
+from .process import ProgressBarWidget, ProcessFollowerWidget, ProcessInputsWidget, ProcessOutputsWidget  # noqa
+from .process import ProcessCallStackWidget, RunningCalcJobOutputWidget, SubmitButtonWidget, ProcessReportWidget  # noqa
```
csadorf, line 14: If we add `# noqa` to basically every import then we can as well stop checking the style. What is the issue that is suppressed here?

aiidalab_widgets_base/process.py
```diff
@@ -21,9 +30,9 @@ def get_running_calcs(process):
     # If the process is a running work chain - returning its children
     if issubclass(type(process), WorkChainNode) and not process.is_sealed:
         calcs = []
-        for link in process.get_outgoing():
-            if issubclass(type(link.node), ProcessNode) and not link.node.is_sealed:
-                calcs += get_running_calcs(link.node)
+        for out_link in process.get_outgoing():
+            if issubclass(type(out_link.node), ProcessNode) and not out_link.node.is_sealed:
```
csadorf, line 34:
```suggestion
            if isinstance(out_link.node, ProcessNode) and not out_link.node.is_sealed:
```

aiidalab_widgets_base/process.py
```diff
@@ -21,9 +30,9 @@ def get_running_calcs(process):
...
+                calcs += get_running_calcs(out_link.node)
         return calcs
```
csadorf, line 36: Why not make this is a generator function?

aiidalab_widgets_base/process.py
```diff
@@ -51,7 +61,7 @@ def __init__(self,
 
         append_output (bool): Whether to clear widget output for each subsequent submission.
         """
-
+        self.path_to_root = kwargs['path_to_root'] if 'path_to_root' in kwargs else '../'
```
csadorf, line 64:
```suggestion
        self.path_to_root = kwargs.get('path_to_root', '../')
```

aiidalab_widgets_base/process.py
```diff
@@ -111,30 +125,91 @@ def on_submitted(self, function):
...
+
+
+class ProcessFollowerWidget(ipw.VBox):
     """A Widget that follows a process until finished."""
 
-    def __init__(self, process, followers=None, update_interval=0.1, **kwargs):
+    def __init__(self, process=None, followers=None, update_interval=0.1, path_to_root='../', **kwargs):
         """Initiate all the followers."""
-        if not isinstance(process, ProcessNode):
+        if not isinstance(process, (ProcessNode, type(None))):
```
csadorf, line 190:
```suggestion
        if not (process is None or isinstance(process, ProcessNode):
```

aiidalab_widgets_base/process.py
```diff
@@ -158,12 +233,77 @@ def on_completed(self, function):
...
+        self.indent_size = 2
+        self.levelname = 'REPORT'
+        self.update()
+        super().__init__(**kwargs)
+
+    def update(self):
+        """Update report that is shown."""
+        if self.process is None:
+            return
+        if isinstance(self.process, CalcJobNode):
```
csadorf, line 252:
```suggestion
            return

        if isinstance(self.process, CalcJobNode):
```

yakutovicha:
> I've started reviewing the code and I also tried to run the notebook, but immediately ran into an error (see screenshot).
> 
> Screen Shot 2020-05-18 at 6 38 11 PM
> 
> I'll probably have more comments, but maybe you can fix that bug in the meantime?

Hm, I can't reproduce the problem. Moreover, the line 213 of `process.py` has no code.

yakutovicha:
> I'll probably have more comments, but maybe you can fix that bug in the meantime?

should be fixed in  c8f1c78

aiidalab_widgets_base/__init__.py
yakutovicha: The issue was `wrong-import-position`. Made PR #92 to remove all the `# noqa`s, please have a look.

aiidalab_widgets_base/process.py
yakutovicha: Changed in c540efe. The only thing I am not sure about is this:
https://github.com/aiidalab/aiidalab-widgets-base/blob/c540efee4cb0e50cedec1593e3b0567f5ee7e39d/aiidalab_widgets_base/process.py#L36

Do you know of any better way to return a "falsy" iterator?


I also took this occasion to redesign `RunningCalcJobOutputWidget` to allow selecting a running sub-process.

aiidalab_widgets_base/__init__.py
csadorf: Got it, #92 is approved.

aiidalab_widgets_base/process.py
```diff
@@ -220,24 +368,74 @@ def __init__(self, process, **kwargs):
...
+        if 'remote_folder' in self.calculation.outputs:
+            output_file_path = os.path.join(self.calculation.outputs.remote_folder.get_remote_path(),
+                                            self.calculation.attributes['output_filename'])
+        if output_file_path and os.path.exists(output_file_path):
+            with open(output_file_path) as fobj:
+                difference = fobj.readlines()[len(self.output):-1]  # Only adding the difference
+                self.output += difference
+                self.value += ''.join(difference)
+
+        # Auto scroll down. Doesn't work in detached mode.
```
csadorf, line 399: I assume that's a problem?

aiidalab_widgets_base/process.py
csadorf: It is not possible to return anything from a generator function. Could you briefly explain under what circumstances it would be important to check the return value? I can then probably recommend a good solution.

csadorf:
I'm going to wait for this PR to be merged before reviewing #91 , ok?

yakutovicha:
> I'm going to wait for this PR to be merged before reviewing #91 , ok?

yes, sure

aiidalab_widgets_base/process.py
yakutovicha: I would like to ensure, that `get_running_calcs` does 0 iterations in case it receives not a process object. In other words:
```
res = list(get_running_calcs(not_a_process)) 
```
`res` should become an empty list.

aiidalab_widgets_base/process.py
```diff
@@ -111,30 +125,92 @@ def on_submitted(self, function):
...
+            options=[('Select output', '')] + outputs_list,
+            label='Select output',
+            description='Select outputs:',
+            style={'description_width': 'initial'},
+            disabled=False,
+        )
+        outputs.observe(self.view_output, names=['value'])
+        super().__init__(children=[ipw.HBox([outputs, self.info]), self.output], **kwargs)
+
+    def view_output(self, change=None):
```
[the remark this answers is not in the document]
yakutovicha: Ah, I forgot one thing here. I wanted to make a base class for `ProcessInputsWidget` and `ProcessOutputsWidget` classes since code is mostly the same. Do you think it is a good idea?

aiidalab_widgets_base/process.py
```diff
@@ -145,7 +221,13 @@ def _follow(self):
             func(self.process)
 
     def follow(self, detach=False):
-        """Follow the process in blocking or non-blocking manner."""
+        """Follow the process in non-blocking manner."""
```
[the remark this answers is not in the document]
yakutovicha: yes, right. The old docstring was correct. I was probably drunk.
Should I keep it as:
```"""Follow the process in blocking or non-blocking manner."""```
or something like
```"""Initiate following the process with or without blocking."""```
is better?

File: aiidalab_widgets_base/process.py [hunk 10]
```diff
@@ -111,30 +125,92 @@ def on_submitted(self, function):
         self._run_after_submitted.append(function)
 
 
-class ProcessFollowerWidget(VBox):
+class ProcessInputsWidget(ipw.VBox):
+    """Widget to select and show process inputs."""
+
+    def __init__(self, process=None, **kwargs):
+        self.process = process
```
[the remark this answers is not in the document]
yakutovicha: Sure, this I can do. Do you think this work, taking into account that `Process` is base class for calculations and workchains?
```
process = Instance(Process, allow_none=True)
```
or should I make a union of sort
```
process = Instance(Union(Process, Calculation, WorkChain), allow_none=True)
```
?

aiidalab_widgets_base/process.py
yakutovicha: Well, that is a sort of philosophical question. Essentially, we have to choices here:

1. Have auto scrolling only in the blocked mode
1. Do not have auto-scrolling at all.

To me, the first option does make sense because if you are using the output in the blocking mode, you are kind of keen to follow it. In detached mode, you just want to have the output.

I do agree that the second option is cleaner, but I am not sure if this is too important.

aiidalab_widgets_base/process.py
```diff
@@ -226,7 +369,9 @@ def __init__(self, process, **kwargs):
 
     def update(self):
         """Update the displayed output."""
-        calcs = get_running_calcs(self.main_process)
+        if self.process is None:
+            return
+        calcs = get_running_calcs(self.process)
```
csadorf: Since `get_running_calcs()` is now a generator function, you will need to refactor the code as described above. I would suggest the following pattern:
```python
calc = None
for calc in get_running_calcs(self.process):
    if calc.id == self.previous_calc_id:
        break
else:
    self.output = []
    # ...
if calc is not None and 'remote_folder' in calc.outputs:
    f_path = # ...
```
However, I'm actually a little bit confused by that. The `calc` chosen on line 238/383, is that just the (random) last calc produced by `get_running_calcs()`?

aiidalab_widgets_base/process.py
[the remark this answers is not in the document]
csadorf: Yes.

aiidalab_widgets_base/process.py
csadorf: I think it's fine to have that slightly inconsistent behavior as long as it is well documented. So make sure to make that clear not only in the in-code comments, but also in the doc-string of the affected function(s). Maybe we can also create an issue for that to keep track of this? I'd classify this as a (tolerated) bug.

aiidalab_widgets_base/process.py
yakutovicha: ah, great, so in my case I should just remove the `return []`

yakutovicha: done

aiidalab_widgets_base/process.py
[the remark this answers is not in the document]
yakutovicha: Changed for both: `ProcessInputsWidget` and `ProcessOutputsWidget`

aiidalab_widgets_base/process.py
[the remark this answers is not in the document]
yakutovicha: done

aiidalab_widgets_base/process.py
yakutovicha: Added the docstring in 89234fb

aiidalab_widgets_base/process.py
yakutovicha: I've no idea how you ended up in this outdated version of the file, but things have changed quite a bit since that version :).

yakutovicha: this is how it looks now:
https://github.com/aiidalab/aiidalab-widgets-base/blob/89234fb960e77f02f6b6efbc2dce9e7f4d039836/aiidalab_widgets_base/process.py#L437

and works pretty well I would say.

aiidalab_widgets_base/process.py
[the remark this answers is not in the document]
yakutovicha: done in  e56d334

aiidalab_widgets_base/process.py
```diff
@@ -2,34 +2,37 @@
...
-    can be either the process itself or its running children."""
+    """Takes a process and yeilds running children calculations."""
 
     # If a process is a running calculation - returning it
     if issubclass(type(process), CalcJobNode) and not process.is_sealed:
-        return [process]
+        yield process
 
     # If the process is a running work chain - returning its children
     if issubclass(type(process), WorkChainNode) and not process.is_sealed:
```
csadorf, line 29:
```suggestion
    eliif issubclass(type(process), WorkChainNode) and not process.is_sealed:
```

aiidalab_widgets_base/process.py
```diff
@@ -111,41 +115,111 @@ def on_submitted(self, function):
...
+                ]))
         self.update()
-        super(ProcessFollowerWidget, self).__init__(children=self.followers, **kwargs)
+        self.output = ipw.HTML()
+        super().__init__(children=[self.output] + self.followers, **kwargs)
 
     def update(self):
         for follower in self.followers:
-            follower.update()
+            follower.children[1].update()
```
csadorf, line 202: Is this to be recursive? Wouldn't it need to be something like `for child in follower.children[1:]: child.update()`?

yakutovicha: No, it is not recursive. It is because I am appending follower together with the title in one `VBox` block.
```
                self.followers.append(ipw.VBox([
                    ipw.HTML("<h2><b>{}</b></h2>".format(follower.title)),
                    follower,
                ]))
```

so to access the follower I use the index.

PR #92, opened 2020-05-18
Remove '#noqa' from all imports in __init__.py

yakutovicha:


yakutovicha:
Travis passed, but for some reason the info is not updated here.

aiidalab_widgets_base/__init__.py
```diff
@@ -1,20 +1,19 @@
 """Reusable widgets for AiiDA lab applications."""
-# pylint: disable=unused-import
+# pylint: disable=unused-import,wrong-import-position
 from aiida import load_profile
 load_profile()
```
csadorf, line 4: I assume the 'wrong-import-position' is caused by this intermittent call? Could we just move it after the imports or disable the warning only for this line?

yakutovicha: No, otherwise we won't be able to load AiiDA objects

yakutovicha: But I can disable the warning for this line, this is true.

yakutovicha: Nope, didn't work. Prospector starts to complain about every subsequent import line.

csadorf: Ok, then let's go with the module-wide override.

PR #94, opened 2020-05-24
Changing StructureManager API

### LINKED ISSUE #65: API for structuremanager
The current API of the Structuremanager requires the user to know that you have to pass a list of tuples where the first item is the title and the second the importer widget.
These implicit things are not easy to remember and one usually tries to avoid them in APIs.

One way to avoid this would be this:
```
StructureManager(importers=[ Importer1(title="asdfasf"), Importer2(title="sfesdf"), ...])
```

It also lets you set a default title if people don't need a specific one.

### LINKED ISSUE #75: StructureManagerWidget: do not allow to modify the description if the structure is stored.
Currently, the description field remains active even if the structure is already stored. That is not right and is definintely confusing.

### LINKED ISSUE #95: StructureManagerWidget: add link between structures
If the input structure is already stored in the AiiDA database, while the output structure has some modifications, there should be a link between them in the AiiDA database.

yakutovicha:
fixes #65
fixes #75 
fixes #95 

Changing API of the structure Manager.

After some tests/considerations, this is the final API I came up with:

1. **Traits**. `StructureManager` has 3 structure-related traits: `input_structure`, `structure`, `structure_node`.
    - `input_structure` should be mostly used by the structure importers. It is where the new structure is coming from.
    - `structure` is a trait that contains the current structure as an ASE object.
    - `structure_node` is the current structure represented as an AiiDA-type structure object.

2. **Importers/editors**.  Both importers and editors can now be incorporated in similar way:

```python
StructureManagerWidget(
    importers=[ Importer1(title="Title 1"), Importer2(title="Title 2"), ...],
    editors = [Editor1(title="Title 3"), Editor2(title="Title")]
)
```

- `StructureManagerWidget` object links itself through the `manager` trait. Using this trait importers/editors can link any trait of `StructureManagerWidget` at their convinience as shown below:

```python
    @observe('manager')
    def _change_manager(self, value):
        """Set structure manager trait."""
        manager = value['new']
        if manager is None:
            return
        dlink((self, 'structure'), (manager, 'input_structure'))
```
 

   - The title for the tabs will be taken from the `.title` attribute of importers/editors.

3. **Links between structures**. `StructureManagerWidgets` sometimes creates a link between structures `input_structure` and  `structure_node`. For this to happen the following conditions must be met: 
    - `input_structure` trait should point at a structure that is already stored in the AiiDA database
    - `structure_node` should be the `input_structure` with at least one modification.


aiidalab_widgets_base/databases.py
```diff
@@ -58,6 +59,14 @@ def __init__(self, **kwargs):
         ]
         super(CodQueryWidget, self).__init__(children=children, **kwargs)
 
+    @observe('manager')
+    def _change_manager(self, value):
+        """Set structure manager trait."""
```
ltalirz, line 64: I think one line of explanation of what this actually does would be helpful here (and in the other places where this
```suggestion
        """Set structure manager trait
        
        Links `input_structure` trait of manager to `structure` trait.
        ."""
```

ltalirz:
> 1. **Traits**. `StructureManager` has 3 structure-related traits: `input_structure`, `structure`, `structure_node`.
>    - `input_structure` should be mostly used by the structure importers. It is where the new structure is coming from.
>    - `structure` is a trait that contains the current structure as an ASE object.
>    - `structure_node` is the current structure represented as an AiiDA-type structure object.

If you want to homogenize the names, you could use something like `structure_input`, `structure_live`, `structure_node` 
Anyhow, up to you.

aiidalab_widgets_base/databases.py
yakutovicha: I can do that, thanks for the suggestion.

yakutovicha:
> If you want to homogenize the names, you could use something like `structure_input`, `structure_live`, `structure_node`
> Anyhow, up to you.

I like that. Let's wait for @csadorf opinion, and, if we all agree, I will implement those changes.

aiidalab_widgets_base/structures.py
```diff
@@ -180,33 +193,71 @@ def _default_node_class():
...
+
+            # If the input_structure trait is set to AiiDA node, check what type
+            elif isinstance(self.input_structure, Data):
+
+                # Transform the structure to the StructureNode if needed.
+                if isinstance(self.input_structure, StructureNode):
+                    structure_node = self.input_structure
+                else:
+                    # self.structure was already converted to the ASE Atoms object.
+                    structure_node = StructureNode(ase=self.structure)
```
csadorf, line 218: I'm inclined to suggest that we refactor this converter code into a dedicated private class method: `_convert_to_structure_node(cls, structure)` and then call here with `structure_node = self._convert_to_structure_node(self.input_srtucture)`.

aiidalab_widgets_base/structures.py
```diff
@@ -180,33 +193,71 @@ def _default_node_class():
...
+                self.structure_label.value = self.structure_node.label
+                self.structure_label.disabled = True
+                self.structure_description.value = self.structure_node.description
+                self.structure_description.disabled = True
+            else:
+                self.btn_store.disabled = False
+                self.structure_label.value = self.structure.get_chemical_formula()
+                self.structure_label.disabled = False
+                self.structure_description.value = ''
+                self.structure_description.disabled = False
```
csadorf, line 233: This should be in a function decorated with `@observe('structure_node')`.

aiidalab_widgets_base/structures.py
```diff
@@ -273,14 +313,24 @@ def _default_structure(self):
...
         self._select_structure.observe(self._on_select_structure, names=['value'])
         super().__init__(children=[self._select_structure], **kwargs)
 
+    @observe('manager')
+    def _change_manager(self, value):
+        """Set structure manager trait."""
+        manager = value['new']
+        if manager is None:
+            return
+        dlink((self, 'structure'), (manager, 'input_structure'))
```
csadorf, line 332: Second duplication, if kept, this should be refactored into an abstract base class. But I hope that we can actually drop this (see previous comment).

aiidalab_widgets_base/structures.py
```diff
@@ -357,6 +409,14 @@ def disable_drop_label(change):
         self.search()
         super().__init__([box, h_line, self.results])
 
+    @observe('manager')
+    def _change_manager(self, value):
```
csadorf, line 413: Third duplication.

aiidalab_widgets_base/structures.py
```diff
@@ -463,6 +525,14 @@ def __init__(self):
         self.output = ipw.HTML("")
         super().__init__([self.smiles, self.create_structure_btn, self.output])
 
+    @observe('manager')
+    def _change_manager(self, value):
```
csadorf, line 529: 4th duplication

csadorf:
@yakutovicha Could you respond to my comments, please?

yakutovicha:
> @yakutovicha Could you respond to my comments, please?

sure, it is in my todo list.

aiidalab_widgets_base/databases.py
```diff
@@ -58,6 +59,14 @@ def __init__(self, **kwargs):
...
+        manager = value['new']
+        if manager is None:
+            return
+        dlink((self, 'structure'), (manager, 'input_structure'))
```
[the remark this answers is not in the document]
yakutovicha: My motivation here is the following. I would like to give to the importer/editor developers a possibility to access any element of the manager if needed. For the `editors` the situation is quite clear to me: when editing a structure one wants to have the freedom to sync any useful trait of the viewer, be it camera position or currently selected atom. Also, this approach looks more future-proof to me. If a developer realizes that they need to link some additional trait they can just do it without a need to change the behaviour of the `StructureManagerWidget`.

With the importers, however, the situation appears less clear. It is true that currently, it is enough to only link the`input_structure` trait. Nevertheless, I don't know what might be needed in future, and, I think it would be nice to manage importers and editors in a similar way.

So I would really prefer to give to the importers/editors link to the StructureManager object and let them decide which traits they want to sync.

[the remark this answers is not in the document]
csadorf: This appears like premature generalization and a leak of responsibilities to me. Specifically, having this bidirectional dependency is going to make it extremely difficult to have a clear separation of concerns between the manager and the various importers/editors in the future. In conclusion, I am strongly opposed to this.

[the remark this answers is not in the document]
yakutovicha: I replaced the `manager` trait with `connect_to_manager()` method as we discussed offline.

aiidalab_widgets_base/databases.py
yakutovicha: We discussed with @csadorf and agreed to replace the trait with `connect_to_manager` method. There should be much less of the code duplication now.

yakutovicha:
> If you want to homogenize the names, you could use something like `structure_input`, `structure_live`, `structure_node`
> Anyhow, up to you.

Is anyone against of such a naming convention?

csadorf:
> > If you want to homogenize the names, you could use something like `structure_input`, `structure_live`, `structure_node`
> > Anyhow, up to you.
> 
> I like that. Let's wait for @csadorf and @CasperWA opinions, and, if we all agree, I will implement those changes.

I'd much prefer to stick to `.structure` instead of `.structure_live`. None of the other widgets further qualify their "main value", it is usually just `.value`. Furthermore, traitlets are always expected to be "current" or "live", so this would be a misnomer.

If you want to express that `.structure` is the one shown in the editor, then you might want to go with `structure_editor` or so, but that is already clear, because it is *the structure of the editor*.

Since I'm against this renaming, I am also against renaming `input_structure` to `structure_input`.

aiidalab_widgets_base/structures.py
```diff
@@ -119,7 +119,8 @@ def _structure_importers(self, importers):
         # If there is only one importer - no need to make tabs.
         if len(importers) == 1:
             # Assigning a function which will be called when importer provides a structure.
-            importers[0].connect_to_manager(self)
+            if importers[0].has_trait('structure'):
+                dlink((importers[0], 'structure'), (self, 'input_structure'))
```
csadorf: I think there is an expectation that *every* importer and editor *must have* the `.structure` trait, correct? We should only guard against potentially not existing trait that are explicitly optional.

PR #105, opened 2020-06-05
fixed issue related to modify_element in structure editor

### LINKED ISSUE #99: Wrong atom replace in the structure editor.
https://github.com/aiidalab/aiidalab-widgets-base/blob/5440b75b74d6a8aee745271baa8defc0f573dd6c/aiidalab_widgets_base/structures.py#L738

I just realized it by chance thanks to an old "assert" that was introduced by Ole:

**Suppose in the ase structure _atoms_ a given atom is C, and I modify it in H
and suppose other H atoms are present in _atoms_**

if I simply change the Symbol, this new H atom will have mass 12 instead of 1
then when doing new_structure=StructureData(ase=atoms) there will be H atoms with mass 1 and at least 1 H atom with mass 12 **that will be H1 instead of H in new_structure.get_site_kindnames()** 

when changing a atom type (also with Ligands)we should do
newatom=Atom(self.element.value)
atoms[idx].symbol = self.element.value
atoms[idx].mass = newatom.mass

still there could be  other similar issues in cases charges or magnetic moments are defined...

cpignedoli:
fixes #99

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -735,7 +735,13 @@ class BasicStructureEditor(ipw.VBox):
 
         if self.ligand.value == 0:
             for idx in self.selection:
-                atoms[idx].symbol = self.element.value
+                new = Atom(self.element.value)
+                atoms[idx].mass = new.mass
+                atoms[idx].magmom = new.magmom
+                atoms[idx].momentum = new.momentum
+                atoms[idx].symbol = new.symbol
+                atoms[idx].tag = new.tag
+                atoms[idx].charge = new.charge
         else:
             initial_ligand = self.ligand.rotate(align_to=self.action_vector, remove_anchor=True)
             for idx in self.selection:


aiidalab_widgets_base/structures.py
```diff
@@ -735,7 +735,13 @@ def mod_element(self, _=None):
...
         if self.ligand.value == 0:
             for idx in self.selection:
-                atoms[idx].symbol = self.element.value
+                new = Atom(self.element.value)
+                atoms[idx].mass = new.mass
+                atoms[idx].magmom = new.magmom
+                atoms[idx].momentum = new.momentum
+                atoms[idx].symbol = new.symbol
+                atoms[idx].tag = new.tag
+                atoms[idx].charge = new.charge
```
yakutovicha, line 744:
```suggestion
                atoms[idx] = new
```
would this work?

yakutovicha, line 744:
```suggestion
                atoms[idx] = Atom(self.element.value)
```
would this work as well?

yakutovicha: apparently not, I've made [an issue](https://gitlab.com/ase/ase/-/issues/667) to the ASE, let's see what they reply.

yakutovicha: According to what was proposed [here](https://gitlab.com/ase/ase/-/issues/667), the right way should be:
```suggestion
                new = Atoms([Atom(self.element.value)])
                for key in atoms.arrays.keys():
                    atoms.arrays[key][idx] = new.arrays[key][0]
```

yakutovicha: @cpignedoli I realized that the latter change I proposed is not really good, because if `new` contains arrays that `atoms` object doesn't - they aren't going to be added. So I am going to merge your solution for the time being.

yakutovicha (approved):

PR #108, opened 2020-06-23
Run tests with GitHub actions workflow.

csadorf:
Using the aiidalab/aiidalab-test-app-action .


csadorf:
Tests succeeded, see here: https://github.com/csadorf/aiidalab-widgets-base/actions/runs/144662902

PR #110, opened 2020-08-04
config ssh by providing private key

unkcpz:
In setting the remote computer, can firstly connect to the remote
computer with private key provided. Then the password-less connection is
getting set. 
If the private key file name is already in `~/.ssh/` a shortuuid
will be append to distinguish it from the exist one.

Also, the ssh port is explicitly setable with default value 22.

### DIFF
--- a/aiidalab_widgets_base/computers.py
+++ b/aiidalab_widgets_base/computers.py
@@ -1,10 +1,12 @@
 """All functionality needed to setup a computer."""
 
+import os
 from os import path
 from copy import copy
 from subprocess import check_output, call
 
 import pexpect
+import shortuuid
 import ipywidgets as ipw
 from IPython.display import clear_output
 from traitlets import Bool, Dict, Instance, Int, Unicode, Union, link, validate
@@ -17,7 +19,7 @@ from aiida.transports.plugins.ssh import parse_sshconfig
 from aiidalab_widgets_base.utils import predefine_settings, valid_arguments
 
 STYLE = {"description_width": "200px"}
-VALID_SSH_COMPUTER_SETUP_ARGUMETNS = {'hostname', 'username', 'proxy_hostname', 'proxy_username'}
+VALID_SSH_COMPUTER_SETUP_ARGUMETNS = {'hostname', 'username', 'proxy_hostname', 'proxy_username', 'port'}
 VALID_AIIDA_COMPUTER_SETUP_ARGUMETNS = {
     'name', 'hostname', 'description', 'workdir', 'mpirun_cmd', 'ncpus', 'transport_type', 'scheduler', 'prepend_text',
     'append_text'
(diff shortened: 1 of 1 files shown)


unkcpz:
@yakutovicha Do you have time to have a look at this, I think this works well in the server in China. Qiang configures the computer by using this, therefore merging this pr would be helpful for me to then maintain the server.

yakutovicha:
> @yakutovicha Do you have time to have a look at this, I think this works well in the server in China. Qiang configures the computer by using this, therefore merging this pr would be helpful for me to then maintain the server.

Thanks @unkcpz I will have a look ASAP. Before I do that, though, can you please fix the pre-commit failure?

unkcpz:
Sure! fixed. @yakutovicha

yakutovicha (commented):
Thanks, @unkcpz. Generally, looks good to me. One thing I would like to ask is to not split `on_setup_ssh` function into two sup-parts. This leads to a lot of code duplication. Can you please merge them into one and add some `if-else` statement?

Also, I am not sure why you changed the order of the arguments in `_configure_proxy`. Is that really needed?

unkcpz:
Hi Sasha, I put the functions into one with `if-else`. Well, a little bit untidy. 
As for `_configure_proxy` I put `password` after the `password_proxy` since in setting with private key the `password`  is not set. Anyway, I change it back to the original order, and set `password` to None explicitly when using private key mode.

aiidalab_widgets_base/computers.py
```diff
@@ -250,68 +297,106 @@ def _configure_proxy(self, password, proxy_password):
...
 
-            # step 7: modify the ssh config file if necessary
-            if not self.is_in_config():
+        # modify the ssh config file if necessary
+        if not self.is_in_config():
+            if mode == 'password':
                 self._write_ssh_config(proxycmd=proxycmd)
+            if mode == 'private_key':
+                private_key_abs_fname = "~/.ssh/" + private_key_fname
+                self._write_ssh_config(proxycmd=proxycmd, private_key_fname=private_key_abs_fname)
```
yakutovicha, line 379: I think those if statements can be avoided entirely. If I understood correctly, the behaviour is determined by the value of the `private_key_abs_fname`. If, by this time, it is `None`, then no private key will be set.

aiidalab_widgets_base/computers.py
```diff
@@ -141,17 +163,42 @@ def is_in_config(self):
...
+                if content == private_key_content:
+                    return path.basename(fname)
+
+            fname = fname + '_' + shortuuid.uuid()
+        with open(fname, "wb") as file:
+            file.write(private_key_content)
+
+        os.chmod(fname, 0o600)
+
+        return path.basename(fname)
```
yakutovicha, line 200: Can you please make this function return the full path?

yakutovicha (commented):
Thanks @unkcpz. 

>  I put the functions into one with if-else. Well, a little bit untidy.

Yep, I think so too. I added a couple of suggestions that would considerably reduce the `mode` branching. Can you please have a look?

aiidalab_widgets_base/computers.py
```diff
@@ -250,68 +297,106 @@ def _configure_proxy(self, password, proxy_password):
...
+        if mode == 'password':
+            self._ssh_keygen()
+            # temporary passwords
             password = self.__password
             proxy_password = self.__proxy_password
+        if mode == 'private_key':
+            # unwrap private key file and setting temporary private_key content
+            private_key_fname, private_key_content = self.__private_key
+            proxy_password = self.__proxy_password
+            password = None
```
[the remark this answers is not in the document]
unkcpz: Good point! No reason to separate them apart. Fixed.

aiidalab_widgets_base/computers.py
```diff
@@ -250,68 +297,106 @@ def _configure_proxy(self, password, proxy_password):
...
-            if not success:
+        if mode == 'private_key':
+            if private_key_fname is None:  # check private key file
+                print("Please upload your private key file")
                 return
 
-            # step 5: make host known by ssh on the proxy server
-            if not self.is_host_known():
-                self._make_host_known(self.hostname, ['ssh'] + [proxycmd] if proxycmd else [])
+        if mode == 'private_key':
```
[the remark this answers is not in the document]
unkcpz: True! removed.

unkcpz:
@yakutovicha Thanks for the comments. I change the code as your suggestion. Looks more clear now. But I only test password and private_key, not sure the new changes will cause problems on 'proxy' setting.

unkcpz:
By the way, for you to clear review, I didn't rebase the last commit.

aiidalab_widgets_base/computers.py
```diff
@@ -250,68 +297,98 @@ def _configure_proxy(self, password, proxy_password):
...
+        if not success:
+            return
+
+        # make host known by ssh on the proxy server
+        if not self.is_host_known():
+            self._make_host_known(self.hostname, ['ssh'] + [proxycmd] if proxycmd else [])
+
+        if mode == 'password':
+            # sending public key to the main host
+            print(self.hostname, self.username, password, proxycmd)
```
yakutovicha, line 364:
```suggestion
```

Almost missed it :)

yakutovicha (changes requested):
Thanks @unkcpz, looks good. I tested several scenarios and all seems to work properly. Please apply the last requested change and I will approve/merge. 

I think this is a good step forward.

While reviewing the PR I realized that there is some code duplication in `_configure_proxy` and `_on_setup_ssh`. But this should be a subject of a separate PR. I made an issue for that: https://github.com/aiidalab/aiidalab-widgets-base/issues/152

aiidalab_widgets_base/computers.py
unkcpz: Terrible carelessness ;)

yakutovicha (approved):
Looks good, thanks @unkcpz !

PR #111, opened 2020-08-04
CodeDropdown widget: allow to specify URL arguments.

yakutovicha:
This PR introduces the possibility to specify URL parameters
to the code setup link. This allows, for instance, to the QE app
to pre-define all the data needed to setup pw@localhost code.


yakutovicha:
@ltalirz, just fyi, this is a step forward to fix the issue you've reported here: https://github.com/aiidalab/aiidalab-docker-stack/issues/92. Instead of setting up the `pw@localhost` container, one can now predefine it in the `CodeDropdown` widget: https://github.com/aiidalab/aiidalab-qe/pull/9.

I will then remove the code setup from the aiidalab container.

PR #116, opened 2020-08-26
default cell to uploaded  structures without cell

### LINKED ISSUE #97: Add cell in case none was specified in the uploaded file
https://github.com/aiidalab/aiidalab-widgets-base/blob/791d88b2d0b2108d00aeea8b79b55b0891616793/aiidalab_widgets_base/structures.py#L224

A typical case is when the user uploads a simple xyz file for a molecule:
ASE allows to import **cc.xyz**:

2
molecule
C 10 10 10
C 11.35 10 10

```
from ase.io import read
atoms = read('cc.xyz')

atoms.cell will return:
array([[0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.]])
```

**Then the code will break at the creation of the node returning:**
/opt/conda/lib/python3.7/site-packages/aiida/orm/nodes/data/structure.py in _get_valid_cell(inputcell)
     53 
     54     if abs(calc_cell_volume(the_cell)) < _volume_threshold:
---> 55         raise ValueError('The cell volume is zero. Invalid cell.')
     56 
     57     return the_cell

ValueError: The cell volume is zero. Invalid cell.

**Please add cell according to:**
```
sys_size = np.ptp(atoms.positions,axis=0)
no_cell=atoms.cell[0][0] <0.1 or atoms.cell[1][1] <0.1 or atoms.cell[2][2] <0.1
if no_cell:
     # set bounding box as cell
      atoms.cell = sys_size + 10
```

**values lower than 7-10 are not safe!**

eimrek:
Fixes #97

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -259,22 +259,6 @@ class StructureManagerWidget(ipw.VBox):
             self.structure_description.value = ''
             self.structure_description.disabled = False
 
-    def _validate_and_fix_ase_cell(self, ase_structure, vacuum_ang=10.0):
-        """
-        Checks if the ase Atoms object has a cell set,
-        otherwise sets it to bounding box plus specified "vacuum" space
-        """
-        cell = ase_structure.cell
-
-        if (np.linalg.norm(cell[0]) < 0.1 or np.linalg.norm(cell[1]) < 0.1 or np.linalg.norm(cell[2]) < 0.1):
-            # if any of the cell vectors is too short, consider it faulty
-            # set cell as bounding box + vacuum_ang
-            bbox = np.ptp(ase_structure.positions, axis=0)
-            new_structure = ase_structure.copy()
-            new_structure.cell = bbox + vacuum_ang
-            return new_structure
-        return ase_structure
-
     @observe('input_structure')
     def _observe_input_structure(self, change):
         """Returns ASE atoms object and sets structure_node trait."""
@@ -282,7 +266,7 @@ class StructureManagerWidget(ipw.VBox):
         self.history = []
 
         if isinstance(change['new'], Atoms):
-            self.structure = self._validate_and_fix_ase_cell(change['new'])
+            self.structure = change['new']
 
         # If the `input_structure` trait is set to AiiDA node, then the `structure` trait should
         # be converted to an ASE Atoms object.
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/structures.py
```diff
@@ -224,6 +232,7 @@ def _convert_to_structure_node(self, structure):
 
         # If the input_structure trait is set to Atoms object, structure node must be created from it.
         if isinstance(structure, Atoms):
+            self._validate_and_fix_ase_cell(structure)
```
yakutovicha, line 235: Please put it to `_observe_input_structure`, as it receives the external data the first.

(same for below)

aiidalab_widgets_base/structures.py
```diff
@@ -215,6 +215,14 @@ def _sync_structure_node(self):
         else:
             structure_node = self._convert_to_structure_node(self.input_structure)
         self.set_trait('structure_node', structure_node)
+        
+    def _validate_and_fix_ase_cell(self, ase_structure, vacuum_ang=10.0):
+        cell = ase_structure.cell
+        no_cell = cell[0][0] < 0.1 or cell[1][1] < 0.1 or cell[2][2] < 0.1
+        if no_cell:
```
yakutovicha, line 222: maybe merge those two lines.

yakutovicha: Also, strictly speaking, this check is not correct. One should check the vector length, not just one its component.

aiidalab_widgets_base/structures.py
```diff
@@ -118,7 +118,7 @@ def _structure_importers(self, importers):
         """Preparing structure importers."""
         if not importers:
             raise ValueError("The parameter importers should contain a list (or tuple) of "
-                             "importers, got a falsy object.")
+                             "importers, got a false object.")
```
yakutovicha, line 121: Why? By "falsy" object I meant any of the following: `''`, `{}`, `False`, `[]`, `None`.

yakutovicha (changes requested):
Thanks @eimrek, a few comments/suggestions.

aiidalab_widgets_base/structures.py
eimrek: If any of the diagonal components is 0, I would say the cell is faulty

yakutovicha: no, it is not correct, comments @cpignedoli?

aiidalab_widgets_base/structures.py
```diff
@@ -259,13 +259,27 @@ def _observe_structure_node(self, change):
...
+        Checks if the ase Atoms object has a cell set,
+        otherwise sets it to bounding box plus specified "vacuum" space
+        """
+        cell = ase_structure.cell
+
+        if (np.linalg.norm(cell[0]) < 0.1 or np.linalg.norm(cell[1]) < 0.1 or np.linalg.norm(cell[2]) < 0.1):
+            # if any of the cell vectors is too short, consider it faulty
+            # set cell as bounding box + vacuum_ang
+            bbox = np.ptp(ase_structure.positions, axis=0)
+            ase_structure.cell = bbox + vacuum_ang
```
yakutovicha, line 273:
```suggestion
            bbox = np.ptp(ase_structure.positions, axis=0)
            new_structure = ase_structure.copy()
            new_structure.cell = bbox + vacuum_ang
            return new_structure
        return ase_structure
```

aiidalab_widgets_base/structures.py
```diff
@@ -259,13 +259,27 @@ def _observe_structure_node(self, change):
...
+
     @observe('input_structure')
     def _observe_input_structure(self, change):
         """Returns ASE atoms object and sets structure_node trait."""
         # If the `input_structure` trait is set to Atoms object, then the `structure` trait should be set to it as well.
         self.history = []
 
         if isinstance(change['new'], Atoms):
+            self._validate_and_fix_ase_cell(change['new'])
             self.structure = change['new']
```
yakutovicha, line 283:
```suggestion
            self.structure = self._validate_and_fix_ase_cell(change['new'])
```

yakutovicha (changes requested):
2 little changes, just accept them and we are good to go.

aiidalab_widgets_base/structures.py
yakutovicha: The point here is that I don't want to modify the original structure object with some unexpected consequences.

yakutovicha (approved):
looks good, thanks @eimrek

PR #131, opened 2020-09-25
Create OptimadeQueryWidget

### LINKED ISSUE #130: Implement AiiDAlab-specific OPTIMADE structure query and importer widget
Reuse the current widget from [`aiidalab/aiidalab-optimade`](https://github.com/aiidalab/aiidalab-optimade).

This was decided in a meeting between @yakutovicha and myself on 25.09.2020.

CasperWA:
Closes #130.

This widget is meant to mainly be used in AiiDAlab applications, it's specifically good in combination with the StructureManagerWidget.

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -882,15 +882,18 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
 class OptimadeQueryWidget(ipw.VBox):
     """AiiDAlab-specific OPTIMADE Query widget
 
-    Useful as a widget to integrate with the StructureManagerWidget, embedded into applications.
+    Useful as a widget to integrate with the
+    :class:`aiidalab_widgets_base.structures.StructureManagerWidget`,
+    embedded into applications.
 
-    Arguments:
-        embedded (bool): Whether or not to show extra database and provider information.
-            When set to `True`, the extra information will be hidden, this is useful
-            in situations where the widget is used in a Tab or similar, e.g., for the
-            StructureManagerWidget.
-        title (str): Title used for Tab header if employed in StructureManagerWidget.
-
+    :param embedded: Whether or not to show extra database and provider information.
+        When set to `True`, the extra information will be hidden, this is useful
+        in situations where the widget is used in a Tab or similar, e.g., for the
+        :class:`aiidalab_widgets_base.structures.StructureManagerWidget`.
+    :type embedded: bool
+    :param title: Title used for Tab header if employed in
+        :class:`aiidalab_widgets_base.structures.StructureManagerWidget`.
+    :type title: str
     """
 
     structure = Instance(StructureData, allow_none=True)
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -8,17 +8,22 @@ import numpy as np
 import ipywidgets as ipw
 from traitlets import Instance, Int, List, Unicode, Union, dlink, link, default, observe
 
-from sklearn.decomposition import PCA
+from optimade_client.query_filter import OptimadeQueryFilterWidget
+from optimade_client.query_provider import OptimadeQueryProviderWidget
 
 # ASE imports
 import ase
 from ase import Atom, Atoms
 from ase.data import chemical_symbols, covalent_radii
 
-# AiiDA and AiiDAlab imports
+# AiiDA imports
 from aiida.engine import calcfunction
 from aiida.orm import CalcFunctionNode, CalcJobNode, Data, QueryBuilder, Node, WorkChainNode
 from aiida.plugins import DataFactory
+
+from sklearn.decomposition import PCA
+
+# Local imports
 from .utils import get_ase_from_file
 from .viewers import StructureDataViewer
 from .data import LigandSelectorWidget
@@ -106,7 +111,7 @@ class StructureManagerWidget(ipw.VBox):
             ipw.HBox(store_and_description + [self.structure_label, self.structure_description])
         ]
 
-        structure_editors = self._struture_editors(editors)
+        structure_editors = self._structure_editors(editors)
         if structure_editors:
             structure_editors = ipw.VBox([btn_undo, structure_editors])
             accordion = ipw.Accordion([structure_editors])
@@ -137,7 +142,7 @@ class StructureManagerWidget(ipw.VBox):
             dlink((importer, 'structure'), (self, 'input_structure'))
         return importers_tab
 
-    def _struture_editors(self, editors):
+    def _structure_editors(self, editors):
         """Preparing structure editors."""
         if editors and len(editors) == 1:
             link((editors[0], 'structure'), (self, 'structure'))
@@ -150,7 +155,7 @@ class StructureManagerWidget(ipw.VBox):
         # If more than one editor was defined.
         if editors:
             editors_tab = ipw.Tab()
-            editors_tab.children = [i for i in editors]
+            editors_tab.children = tuple(editors)
             for i, editor in enumerate(editors):
                 editors_tab.set_title(i, editor.title)
                 link((editor, 'structure'), (self, 'structure'))
@@ -872,3 +877,46 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
         atoms = self.structure.copy()
         del [atoms[self.selection]]
         self.structure = atoms
+
+
+class OptimadeQueryWidget(ipw.VBox):
+    """AiiDAlab-specific OPTIMADE Query widget
+
+    Useful as a widget to integrate with the StructureManagerWidget, embedded into applications.
+
+    Arguments:
+        embedded (bool): Whether or not to show extra database and provider information.
+            When set to `True`, the extra information will be hidden, this is useful
+            in situations where the widget is used in a Tab or similar, e.g., for the
+            StructureManagerWidget.
+        title (str): Title used for Tab header if employed in StructureManagerWidget.
+
+    """
+
+    structure = Instance(StructureData, allow_none=True)
+
+    def __init__(
+            self,
+            embedded: bool = True,
+            title: str = None,
+            **kwargs,
+    ) -> None:
+        providers = OptimadeQueryProviderWidget(embedded=embedded)
+        filters = OptimadeQueryFilterWidget()
+
+        ipw.dlink((providers, 'database'), (filters, 'database'))
+
+        filters.observe(self._update_structure, names='structure')
+
+        self.title = title if title is not None else 'OPTIMADE'
+        layout = kwargs.pop('layout') if 'layout' in kwargs else {'width': 'auto', 'height': 'auto'}
+
+        super().__init__(
+            children=(providers, filters),
+            layout=layout,
+            **kwargs,
+        )
+
+    def _update_structure(self, change: dict) -> None:
+        """New structure chosen"""
+        self.structure = change['new'].as_aiida_structuredata if change['new'] else None
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -8,9 +8,6 @@ import numpy as np
 import ipywidgets as ipw
 from traitlets import Instance, Int, List, Unicode, Union, dlink, link, default, observe
 
-from optimade_client.query_filter import OptimadeQueryFilterWidget
-from optimade_client.query_provider import OptimadeQueryProviderWidget
-
 # ASE imports
 import ase
 from ase import Atom, Atoms
@@ -877,49 +874,3 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
         atoms = self.structure.copy()
         del [atoms[self.selection]]
         self.structure = atoms
-
-
-class OptimadeQueryWidget(ipw.VBox):
-    """AiiDAlab-specific OPTIMADE Query widget
-
-    Useful as a widget to integrate with the
-    :class:`aiidalab_widgets_base.structures.StructureManagerWidget`,
-    embedded into applications.
-
-    :param embedded: Whether or not to show extra database and provider information.
-        When set to `True`, the extra information will be hidden, this is useful
-        in situations where the widget is used in a Tab or similar, e.g., for the
-        :class:`aiidalab_widgets_base.structures.StructureManagerWidget`.
-    :type embedded: bool
-    :param title: Title used for Tab header if employed in
-        :class:`aiidalab_widgets_base.structures.StructureManagerWidget`.
-    :type title: str
-    """
-
-    structure = Instance(StructureData, allow_none=True)
-
-    def __init__(
-            self,
-            embedded: bool = True,
-            title: str = None,
-            **kwargs,
-    ) -> None:
-        providers = OptimadeQueryProviderWidget(embedded=embedded)
-        filters = OptimadeQueryFilterWidget()
-
-        ipw.dlink((providers, 'database'), (filters, 'database'))
-
-        filters.observe(self._update_structure, names='structure')
-
-        self.title = title if title is not None else 'OPTIMADE'
-        layout = kwargs.pop('layout') if 'layout' in kwargs else {'width': 'auto', 'height': 'auto'}
-
-        super().__init__(
-            children=(providers, filters),
-            layout=layout,
-            **kwargs,
-        )
-
-    def _update_structure(self, change: dict) -> None:
-        """New structure chosen"""
-        self.structure = change['new'].as_aiida_structuredata if change['new'] else None
--- a/aiidalab_widgets_base/__init__.py
+++ b/aiidalab_widgets_base/__init__.py
@@ -15,6 +15,7 @@ from .process import ProcessListWidget
 from .structures import StructureManagerWidget
 from .structures import StructureBrowserWidget, StructureExamplesWidget, StructureUploadWidget, SmilesWidget
 from .structures import BasicStructureEditor
+from .structures import OptimadeQueryWidget
 from .structures_multi import MultiStructureUploadWidget
 from .viewers import viewer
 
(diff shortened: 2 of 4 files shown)


csadorf (commented):
Are we sure that this belongs into the widget-base package? I think it would be perfectly fine to leave it in aiidalab-optimade, no? One major issue I have with moving it into this package is, that it would require us to add "optimade" to the aiidalab-widgets-base dependencies.

CasperWA:
> Are we sure that this belongs into the widget-base package? I think it would be perfectly fine to leave it in aiidalab-optimade, no? One major issue I have with moving it into this package is, that it would require us to add "optimade" to the aiidalab-widgets-base dependencies.

As per the discussion with @yakutovicha on 25.09.2020 this was the decided conclusion.

Let me try to give some reasons and motivations, but note this may not be complete:
@yakutovicha stressed that he would like to remove explicitly installing `optimade-client` or `aiidalab-optimade` in the docker stack.
This specific widget is not available in the `optimade-client` package.
We discussed moving it to the `optimade-client` package, but since we would have to do a `lambda` function transformation or similar each time one would use the widget in any case, to set the `structure` traitlet to either an `ase.Atoms` or `aiida.orm.StructureData`, we moved in another direction.
We discussed simply keeping things the way they were, but that would also mean importing some widgets from the `optimade-client` and the other from `aiidalab-optimade` - also not ideal.
I suggested simply adding the OPTIMADE `Structure` class to the `StructureManagerWidget`, but @yakutovicha was against this for maintenance reasons.
In the end, we settled on moving the only widget present in `aiidalab-optimade` to this repository and package, since the dependencies are not big to begin with, the use of this particular widget is AiiDAlab-specific, especially for use in the `StructureManagerWidget`, and if one wants to use it outside AiiDAlab, the widget is very simple to recreate from the widgets in `optimade-client`. In the end it's all MIT licensed, so a copy-paste of the widget we use here is no issue.

csadorf:
> @yakutovicha stressed that he would like to remove explicitly installing `optimade-client` or `aiidalab-optimade` in the docker stack.

I think that makes a lot of sense.

> This specific widget is not available in the `optimade-client` package.

Sure.

> We discussed moving it to the `optimade-client` package, but since we would have to do a `lambda` function transformation or similar each time one would use the widget in any case, to set the `structure` traitlet to either an `ase.Atoms` or `aiida.orm.StructureData`, we moved in another direction.

I don't think that's an issue. Using small transformation functions is very common practice to resolve minor API friction.

> We discussed simply keeping things the way they were, but that would also mean importing some widgets from the `optimade-client` and the other from `aiidalab-optimade` - also not ideal.

Agreed, that would not be ideal. However I thought the widget was not in `optimade-client` to begin with? Furthermore, it makes total sense to me to import a widget that is designed to be used with AiiDAlab from the `aiidalab-optimade` package.

> I suggested simply adding the OPTIMADE `Structure` class to the `StructureManagerWidget`, but @yakutovicha was against this for maintenance reasons.

What maintenance burden exactly? I don't follow this argument.

> In the end, we settled on moving the only widget present in `aiidalab-optimade` to this repository and package, since the dependencies are not big to begin with, the use of this particular widget is AiiDAlab-specific, especially for use in the `StructureManagerWidget`, and if one wants to use it outside AiiDAlab, the widget is very simple to recreate from the widgets in `optimade-client`. In the end it's all MIT licensed, so a copy-paste of the widget we use here is no issue.

Is it not correct that `aiidalab-widgets-base` will now need to be extended by all requirements of `aiidalab-optimade`?

CasperWA:
> > @yakutovicha stressed that he would like to remove explicitly installing `optimade-client` or `aiidalab-optimade` in the docker stack.
> 
> I think that makes a lot of sense.
> 
> > This specific widget is not available in the `optimade-client` package.
> 
> Sure.
> 
> > We discussed moving it to the `optimade-client` package, but since we would have to do a `lambda` function transformation or similar each time one would use the widget in any case, to set the `structure` traitlet to either an `ase.Atoms` or `aiida.orm.StructureData`, we moved in another direction.
> 
> I don't think that's an issue. Using small transformation functions is very common practice to resolve minor API friction.

Sure. But this would probably greatly reduce the ease with which it can be/is implemented in `StructureManagerWidget` or other similar widgets in AiiDAlab - but I partly agree with your assessment here.

> > We discussed simply keeping things the way they were, but that would also mean importing some widgets from the `optimade-client` and the other from `aiidalab-optimade` - also not ideal.
> 
> Agreed, that would not be ideal. However I thought the widget was not in `optimade-client` to begin with? Furthermore, it makes total sense to me to import a widget that is designed to be used with AiiDAlab from the `aiidalab-optimade` package.

Indeed. The widgets in `optimade-client` are the core widgets for this client. The widget in `aiidalab-optimade` is simply utilizing these widgets, making a combined "choose provider+database" and "set filters for searching" widget. Both of these "sub"-widgets are located in `optimade-client`.
The issue with keeping it in `aiidalab-optimade`, is that it violates the first problem - then we have to keep it in the docker stack, since `aiidalab-optimade` is not an installable PyPI package, and definitely doesn't need to be either.

> > I suggested simply adding the OPTIMADE `Structure` class to the `StructureManagerWidget`, but @yakutovicha was against this for maintenance reasons.
> 
> What maintenance burden exactly? I don't follow this argument.
>
> > In the end, we settled on moving the only widget present in `aiidalab-optimade` to this repository and package, since the dependencies are not big to begin with, the use of this particular widget is AiiDAlab-specific, especially for use in the `StructureManagerWidget`, and if one wants to use it outside AiiDAlab, the widget is very simple to recreate from the widgets in `optimade-client`. In the end it's all MIT licensed, so a copy-paste of the widget we use here is no issue.
> 
> Is it not correct that `aiidalab-widgets-base` will now need to be extended by all requirements of `aiidalab-optimade`?

Sorry, I meant the dependencies of this package (@yakutovicha was worried that if people wished to use this tiny widget outside AiiDAlab, they would suddenly have to rely on a major package - I don't see this as an issue for the reasons I have already stated).
The dependency of `aiidalab-optimade` is only `optimade-client`, this dependency was what we originally wanted to include in the list of docker stack requirements, i.e., installable packages from PyPI - not directly in the Dockerfile. Hence, the added dependency here would be 0 (effectively).

yakutovicha:
> We discussed moving it to the `optimade-client` package, but since we would have to do a `lambda` function transformation or similar each time one would use the widget in any case, to set the `structure` traitlet to either an `ase.Atoms` or `aiida.orm.StructureData`, we moved in another direction.

I would like to comment on this. Essentially, if we will extend the `StructureManager`'s API to include a little transformation function, we could import the `OptimadeClient` that returns `OptimadeStructure` as is.

Currently, however, the `structure` trait of the widget is `dlink`ed to the `input_structure` trait:
https://github.com/aiidalab/aiidalab-widgets-base/blob/cce7876ab87b3b5876683c4a71c62dfc3117fbd9/aiidalab_widgets_base/structures.py#L128

csadorf:
> > We discussed moving it to the `optimade-client` package, but since we would have to do a `lambda` function transformation or similar each time one would use the widget in any case, to set the `structure` traitlet to either an `ase.Atoms` or `aiida.orm.StructureData`, we moved in another direction.
> 
> I would like to comment on this. Essentially, if we will extend the `StructureManager`'s API to allow including a little transformation function, we could import the `OptimadeClient` that returns `OptimadeStructure` as is.

That's not necessary. It is very easily possible to add that transformation without API changes to `StructureManager`.

yakutovicha:
> That's not necessary. It is very easily possible to add that transformation without API changes to `StructureManager`.

@csadorf can you make an example please?

CasperWA:
> > > We discussed moving it to the `optimade-client` package, but since we would have to do a `lambda` function transformation or similar each time one would use the widget in any case, to set the `structure` traitlet to either an `ase.Atoms` or `aiida.orm.StructureData`, we moved in another direction.
> > 
> > 
> > I would like to comment on this. Essentially, if we will extend the `StructureManager`'s API to allow including a little transformation function, we could import the `OptimadeClient` that returns `OptimadeStructure` as is.
> 
> That's not necessary. It is very easily possible to add that transformation without API changes to `StructureManager`.

Indeed. However, I would probably go so far as to consider the possibility for a "catch-all" traitlet in the `StructureManagerWidget`, which can consume a large number of supported "Structure" classes, e.g., `ase.Atoms`, `aiida.orm.StructureData`, `aiida.orm.CifData`, `pymatgen.Structure`, `pymatgen.Molecule`, `optimade.adapters.Structure`, and have an internal transformation available for all these, but then only use one of these internally otherwise. This would make the `StructureManagerWidget` much more flexible in terms of developing widgets for it, and if specified correctly, the transformations should not be significant.

csadorf:
> > That's not necessary. It is very easily possible to add that transformation without API changes to `StructureManager`.
> 
> @csadorf can you make an example please?

Sorry, you edited your comment after I had already replied. The simplest way I can think of for such a transformation under existing trait name is to simply write a thin wrapper:

```python

class FooWrapper(HBox):

    number = Int()
    def __init__(self, *args, **kwargs):
        self.foo = Foo(*args, **kwargs)
        dlink((foo, 'number'), (self, 'number'), transform=int)
        super().__init__(children=[self.foo])
```

csadorf:
We might want to think a bit more about how best to enable the connection of importers with varying import formats. The kind of thin wrapper I proposed in my previous comment could be provided by a utility function that you only pass the transformation function or so. I'm not a fan of having "catch-all" traits that do internal transformations (something I've actively developed away from), because it makes the implementation of the manager very, very complex and difficult. But I see the problem and there is probably a way we can enable this a bit more intelligently without completely messing up the code internally.

yakutovicha:
> The kind of thin wrapper I proposed in my previous comment could be provided by a utility function that you only pass the transformation function or so.

That what essentially what I was proposing here:

> Essentially, if we will extend the `StructureManager`'s API to allow including a little transformation function, we could import the `OptimadeClient` that returns `OptimadeStructure` as is.

If we know that the object returned by the Importer is different from (a) ASE Atoms and (b) AiiDA structure object, we can add a possibility to inject a little transformation. How exactly we do it is another question.

CasperWA:
So it seems we are all talking for and against the same thing at the same time, but maybe with different scopes laid out in our heads (transformers for the `structure` traitlet).

yakutovicha:
> So it seems we are all talking for and against the same thing at the same time, but maybe with different scopes laid out in our heads (transformers for the `structure` traitlet).

I feel we are converging :)

yakutovicha:
Guys, what if we do something like:

widget = StructureManagerWidget(
    importers=[
        StructureUploadWidget(title="From computer"),
        CodQueryWidget(title="COD"),
        StructureBrowserWidget(title="AiiDA database"),
        SmilesWidget(title="SMILES"),  # requires OpenBabel!
        StructureExamplesWidget(
            title="From Examples",
            examples=[
                ("Silicon oxide", "miscellaneous/structures/SiO2.xyz")
            ]),
    ],
    editors = [
        BasicStructureEditor(title="Basic Editor"),
    ],
   transformators: {SmilesWidget: my_transform_function, ...}  # <<<--- The change.
)

csadorf:
I'd prefer something like this:
```python
widget = StructureManagerWidget(
    importers=[
        StructureUploadWidget(title="From computer"),
        CodQueryWidget(title="COD"),
        StructureBrowserWidget(title="AiiDA database"),
        transform_structure(SmilesWidget(title="SMILES"), transform_function),
        StructureExamplesWidget(
            title="From Examples",
            examples=[
                ("Silicon oxide", "miscellaneous/structures/SiO2.xyz")
            ]),
    ],
    editors = [
        BasicStructureEditor(title="Basic Editor"),
    ],
)
```

csadorf:
Actually, the best way would be to move the `dlink` out the structure manager constructor and into "user space". Because that is creating this issue in the first place.

csadorf:
Happy to create a PR with revised API. If we are already breaking, then let's do it well. ;)

yakutovicha:
> Happy to create a PR with revised API. If we are already breaking, then let's do a proper break. ;)

well, adding an optional parameter as I described won't break the API.

csadorf:
> > Happy to create a PR with revised API. If we are already breaking, then let's do a proper break. ;)
> 
> well, adding an optional parameter as I described won't break the API.

My suggested alternative would also not break.

CasperWA:
So the decision here is to implement a new API rather than merge the already present PR?

csadorf:
> So the decision here is to implement a new API rather than merge the already present PR?

That would be my preferred option, but of course we need to find consensus.

yakutovicha:
> > So the decision here is to implement a new API rather than merge the already present PR?
> 
> That would be my preferred option, but of course we need to find consensus.

I also agree with adapting API, but I am not sure about the right way do it. Let's maybe ask @giovannipizzi's input, when we meet for the AiiDAlab meeting.

csadorf:
Proxy-based solution:
```python
from aiidalab_widgets_base import ImporterProxy

smiles_proxy = ImporterProxy()
dlink((SmilesWidget(title='SMILES'), 'some_structure_trait'), (smiles_proxy, 'structure'))

widget = StructureManagerWidget(
    importers=[
        StructureUploadWidget(title="From computer"),
        CodQueryWidget(title="COD"),
        StructureBrowserWidget(title="AiiDA database"),
        smiles_proxy,
        StructureExamplesWidget(
            title="From Examples",
            examples=[
                ("Silicon oxide", "miscellaneous/structures/SiO2.xyz")
            ]),
    ],
    editors = [
        BasicStructureEditor(title="Basic Editor"),
    ],
)
```

CasperWA:
Final question, specific to this PR. Do we want to rename the widget, e.g., to `OptimadeStructuresWidget` or `AiidalabOptimadeWidget`? The current name was based on the existing `CodQueryWidget` name. To match this, but maybe it doesn't have to?

yakutovicha:
> Final question, specific to this PR. Do we want to rename the widget, e.g., to `OptimadeStructuresWidget` or `AiidalabOptimadeWidget`? The current name was based on the existing `CodQueryWidget` name. To match this, but maybe it doesn't have to?

For me `AiidalabOptimadeWidget` is preferable.

aiidalab_widgets_base/structures.py
```diff
@@ -8,17 +8,22 @@
 import ipywidgets as ipw
 from traitlets import Instance, Int, List, Unicode, Union, dlink, link, default, observe
 
-from sklearn.decomposition import PCA
+from optimade_client.query_filter import OptimadeQueryFilterWidget
+from optimade_client.query_provider import OptimadeQueryProviderWidget
```
CasperWA, line 12: I could, in theory, move these imports down into the widget itself. They would then be imported _only_ together with the widget, but they would also be imported _every time_ with the widget ...

yakutovicha:
As the conclusion of the meeting with @CasperWA, @csadorf, @giovannipizzi, @dou-du and myself we will keep the current implementation of the widget.

CasperWA:
Ah, also. I just noticed the `CodQueryWidget` is actually in the module `databases`. Maybe I should move the OPTIMADE widget there as well?

aiidalab_widgets_base/structures.py
yakutovicha: It is fine like this I think. We typically move all the imports to the top.

CasperWA:
> > Final question, specific to this PR. Do we want to rename the widget, e.g., to `OptimadeStructuresWidget` or `AiidalabOptimadeWidget`? The current name was based on the existing `CodQueryWidget` name. To match this, but maybe it doesn't have to?
> 
> For me `AiidalabOptimadeWidget` is preferable.

Should we then also consider renaming `CodQueryWidget` to `AiidalabCodWidget`?

yakutovicha:
> Ah, also. I just noticed the `CodQueryWidget` is actually in the module `databases`. Maybe I should move the OPTIMADE widget there as well?

yes, I would say so.

aiidalab_widgets_base/structures.py
```diff
@@ -872,3 +877,49 @@ def remove(self, _):
         atoms = self.structure.copy()
         del [atoms[self.selection]]
         self.structure = atoms
+
+
+class OptimadeQueryWidget(ipw.VBox):
```
csadorf, line 882: I would prefer to have the widget definition in a distinct module.

csadorf (changes requested):

aiidalab_widgets_base/structures.py
csadorf: But this way is also fine since this is how most widgets are defined right now.

yakutovicha:
> > > Final question, specific to this PR. Do we want to rename the widget, e.g., to `OptimadeStructuresWidget` or `AiidalabOptimadeWidget`? The current name was based on the existing `CodQueryWidget` name. To match this, but maybe it doesn't have to?
> > 
> > 
> > For me `AiidalabOptimadeWidget` is preferable.

Thinking more about it... I would like to withdraw my opinion about the widget's name. We have no widgets that start with `Aiidalab...`, so maybe `AiidalabOptimadeWidget` is still fine. Maybe we should agree on the naming policy as well.

> 
> Should we then also consider renaming `CodQueryWidget` to `AiidalabCodWidget`?

I would avoid any renamings for the moment, I don't want to deal with the compatibility issues now.

aiidalab_widgets_base/structures.py
yakutovicha: maybe move to the `databases` module as @CasperWA mentioned earlier.

CasperWA: As opposed in the general `databases` file, where only `CodQueryWidget` is now?

yakutovicha:
@CasperWA there is just a little pre-commit issue. Please fix it and I will approve.

csadorf (approved):
Approved pending that the style issue is fixed.

CasperWA:
> @CasperWA there is just a little pre-commit issue. Please fix it and I will approve.

Them's be the issues of working on different computers throughout the week...
Fix is pushed.

On another note, I am not fond of the yapf styling. The "fix" is a rather ugly extra indentation. Also, we already moved away from `prospector` in `aiida-core` due to it being slow and causing some problems. I think mainly with its formatting choices and dependencies. Perhaps we should move to `black` instead?

yakutovicha (approved):

yakutovicha:
Thanks @CasperWA

PR #144, opened 2020-10-30
CalcJobOutputWidget bug fix.

### LINKED ISSUE #104: CalcJobOutputWidget: gracefully handle situations where `output_filename` attribute is not set.
Currently, one gets an exception:

```python
KeyError                                  Traceback (most recent call last)
<ipython-input-6-f94cf28eb96a> in <module>
      2     process,
      3     followers=[ProgressBarWidget(), ProcessReportWidget(), ProcessCallStackWidget(), RunningCalcJobOutputWidget()], path_to_root='../',
----> 4     update_interval=2)
      5 display(follower)
      6 follower.follow(detach=True)

~/apps/aiidalab-widgets-base/aiidalab_widgets_base/process.py in __init__(self, process, followers, update_interval, path_to_root, **kwargs)
    197                     follower,
    198                 ]))
--> 199         self.update()
    200         self.output = ipw.HTML()
    201         super().__init__(children=[self.output] + self.followers, **kwargs)

~/apps/aiidalab-widgets-base/aiidalab_widgets_base/process.py in update(self)
    203     def update(self):
    204         for follower in self.followers:
--> 205             follower.children[1].update()
    206 
    207     def _follow(self):

~/apps/aiidalab-widgets-base/aiidalab_widgets_base/process.py in update(self)
    442             else:
    443                 self.output.calculation = self.selection.value
--> 444         self.output.update()
    445 
    446 

~/apps/aiidalab-widgets-base/aiidalab_widgets_base/process.py in update(self)
    393         if 'remote_folder' in self.calculation.outputs:
    394             output_file_path = os.path.join(self.calculation.outputs.remote_folder.get_remote_path(),
--> 395                                             self.calculation.attributes['output_filename'])
    396         if output_file_path and os.path.exists(output_file_path):
    397             with open(output_file_path) as fobj:

KeyError: 'output_filename'
```

yakutovicha:
fixes #104 

Gracefully handle situations where `output_filename` attribute isn't
set for a calculation class.

### DIFF
--- a/aiidalab_widgets_base/process.py
+++ b/aiidalab_widgets_base/process.py
@@ -412,7 +412,7 @@ class CalcJobOutputWidget(ipw.Textarea):
 
         output_file_path = None
         try:
-            output_file_path = os.path.join(self.calculation.outputs.remote_folders.get_remote_path(),
+            output_file_path = os.path.join(self.calculation.outputs.remote_folder.get_remote_path(),
                                             self.calculation.attributes['output_filename'])
         except KeyError:
             self.placeholder = "The `output_filename` attribute is not set for " \
--- a/aiidalab_widgets_base/process.py
+++ b/aiidalab_widgets_base/process.py
@@ -16,7 +16,7 @@ from aiida.orm import CalcFunctionNode, CalcJobNode, Node, ProcessNode, WorkChai
 from aiida.cmdline.utils.common import get_calcjob_report, get_workchain_report, get_process_function_report
 from aiida.cmdline.utils.ascii_vis import format_call_graph
 from aiida.cmdline.utils.query.calculation import CalculationQueryBuilder
-from aiida.common.exceptions import MultipleObjectsError, NotExistent
+from aiida.common.exceptions import MultipleObjectsError, NotExistent, NotExistentAttributeError
 
 # Local imports.
 from .viewers import viewer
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/process.py
```diff
@@ -412,8 +412,12 @@ def update(self):
...
         output_file_path = None
         if 'remote_folder' in self.calculation.outputs:
-            output_file_path = os.path.join(self.calculation.outputs.remote_folder.get_remote_path(),
-                                            self.calculation.attributes['output_filename'])
+            try:
+                output_file_path = os.path.join(self.calculation.outputs.remote_folder.get_remote_path(),
+                                                self.calculation.attributes['output_filename'])
+            except KeyError:
+                self.placeholder = f"The `output_filename` attribute isn't set for {self.calculation.process_class}. " \
+                "The output won't appear here."
```
csadorf, line 420:
```suggestion
                self.placeholder = f"The `output_filename` attribute is not set for {self.calculation.process_class}. " \
                "Nothing to show."
```
We should generally avoid contractions like "isn't" or "won't" in formal messages IMO. I also found the second sentence slightly awkward.

csadorf (changes requested):

aiidalab_widgets_base/process.py
```diff
@@ -412,8 +412,12 @@ def update(self):
 
         output_file_path = None
         if 'remote_folder' in self.calculation.outputs:
```
csadorf, line 414: I think we are mixing the two approaches of checking a specific condition in this line and then later using a try: except: block to determine what do if some information is missing. Maybe we can just check both conditions in this line or alternatively not check here and just catch both conditions ( no 'output_filename' and no 'remote_folder') via the `KeyError` except.

aiidalab_widgets_base/process.py
```diff
@@ -412,8 +412,12 @@ def update(self):
...
-            output_file_path = os.path.join(self.calculation.outputs.remote_folder.get_remote_path(),
-                                            self.calculation.attributes['output_filename'])
+            try:
+                output_file_path = os.path.join(self.calculation.outputs.remote_folder.get_remote_path(),
+                                                self.calculation.attributes['output_filename'])
+            except KeyError:
+                self.placeholder = f"The `output_filename` attribute is not set for {self.calculation.process_class}. " \
+                "Nothing to show."
         if output_file_path and os.path.exists(output_file_path):
```
csadorf, line 421: In relation to my previous comment on line 414, I believe the control flow is incorrect here. Is the `output_file_path` variable even set at this point if we caught a `KeyError` earlier?

aiidalab_widgets_base/process.py
```diff
@@ -411,13 +411,16 @@ def update(self):
...
-                f"{self.calculation.process_class}. Nothing to show."
+        try:
+            output_file_path = os.path.join(self.calculation.outputs.remote_folder.get_remote_path(),
+                                            self.calculation.attributes['output_filename'])
+        except KeyError:
+            self.placeholder = "The `output_filename` attribute is not set for " \
+            f"{self.calculation.process_class}. Nothing to show."
+        except NotExistentAttributeError:
+            self.placeholder = "The object `remote_folder` was not found among the process outputs. " \
+            "Nothing to show."
```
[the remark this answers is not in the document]
yakutovicha: I will keep it as is.

csadorf (approved):

PR #145, opened 2020-10-30
Add optimade-client dependency.

yakutovicha:
This PR should fix the Travis tests.

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -19,7 +19,7 @@
         "more_itertools",
         "nglview",
         "numpy",
-	"optimade-client",
+        "optimade-client",
         "pandas",
         "scikit-learn"
     ],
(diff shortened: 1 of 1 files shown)


setup.json
```diff
@@ -19,6 +19,7 @@
         "more_itertools",
         "nglview",
         "numpy",
+	"optimade-client",
```
csadorf, line 22: Looks like the indentation is wrong.

yakutovicha: Indeed, that was a Tab. Fixed.

csadorf (approved):
Looks good. However, I'm wondering whether overall we should be provide ranges here. All of these dependencies are very vague without version numbers.

yakutovicha:
> Looks good. However, I'm wondering whether overall we should be provide ranges here. All of these dependencies are very vague without version numbers.

I agree, made an issue to not forget about this: https://github.com/aiidalab/aiidalab-widgets-base/issues/146

PR #153, opened 2020-11-11
added advanced selection of atoms

cpignedoli:
Now, to select atoms, in addition to '1 4 10..30'
a boolean syntax is allowed example 'x>1 and name is [C,N] and d_from (2,2,2) >2 and id >30'

**USE OF eval**, do not know how to replace it

### DIFF
--- a/aiidalab_widgets_base/misc.py
+++ b/aiidalab_widgets_base/misc.py
@@ -141,10 +141,6 @@ class ReversePolishNotation:
             if is_number(ope):
                 stack.append(float(ope))
                 stackposition += 1
-            # Special case distance.
-            elif 'd_from' in ope:
-                stack.append(self.operators['d_from']['function'](ope))
-                stackposition += 1
             elif ope in self.operators:
                 nargs = self.operators[ope]['nargs']
                 arguments = [stack[stackposition + indx] for indx in list(range(-nargs + 1, 1))]
--- a/aiidalab_widgets_base/misc.py
+++ b/aiidalab_widgets_base/misc.py
@@ -1,5 +1,6 @@
 """Some useful classes used acrross the repository."""
-import re
+import io
+import tokenize
 import ipywidgets as ipw
 from traitlets import Unicode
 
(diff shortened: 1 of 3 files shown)


csadorf:
> **USE OF eval**, do not know how to replace it

I generally don't think that the use of `eval()` is a problem _per se_, since the whole point of AiiDAlab is essentially arbitrary code execution. However it would of course be a problem in the context of web applications that happen to use widgets from this repository. Have you looked into implementing the functionality with [`ast.literal_eval()`](https://docs.python.org/3.7/library/ast.html#ast.literal_eval)?

cpignedoli:
Thanks a lot Simon, I will check

cpignedoli:
I think it will not work with ast.literal:
we have a ase Atoms object : structure
and a string contaning a condition on the ase object e.g.:
'[atom.index for atom in structure if ( atom.symbol  in "H,N".replace(" ","").split(",") )]'
the string is built from a simple syntax entered in a text widget e.g. in this case the input would be:
name is [H,N]

yakutovicha:
> > **USE OF eval**, do not know how to replace it
> 
> I generally don't think that the use of `eval()` is a problem _per se_, since the whole point of AiiDAlab is essentially arbitrary code execution. However it would of course be a problem in the context of web applications that happen to use widgets from this repository. Have you looked into implementing the functionality with [`ast.literal_eval()`](https://docs.python.org/3.7/library/ast.html#ast.literal_eval)?

We looked to the code together with @cpignedoli, and it seems that we can replace the `eval` execution by the `f-string`.

something like this:
```python
literal_eval(f'{[atom.index for atom in structure if ( atom.symbol in "H,N".split(",") )]}')
```

cpignedoli:
TO BE DONE:
1)
the two classes 
class Stack:
    """Class defining the stack for RPN notation"""

and
class InfixConverter:
    """Class defining operations for RPN conversion"""
 that I adapted from:
 Author: Alaa Awad
 Description: program converts infix to postfix notation
 https://gist.github.com/awadalaa/7ef7dc7e41edb501d44d1ba41cbf0dc6

Should be moved in a more appropriate place

2)function     def advanced_selection(self, struc, postfix):
        """Applies rules of advanced selection."""
should use self.struct from the viewer class instead of having 'struc'' as argument
profiler complains: there are 25 variables out of 20 suggested but I would not change teh structure of the function

3) Be very careful if you change parsing and string replacements, even order matters.. assumptions where done to keep shorter the code.

yakutovicha (approved):

PR #154, opened 2020-11-17
Improve computer setup widgets:

yakutovicha:
- Straighten the logic
- Use traits that have the same names as aiida computer parameters
- Remove redundant and/or unnecessary code.

### DIFF
--- a/aiidalab_widgets_base/computers.py
+++ b/aiidalab_widgets_base/computers.py
@@ -744,19 +744,19 @@ class ComputerDropdown(ipw.VBox):
     def _validate_selected_computer(self, change):
         """Select computer either by name or by class instance."""
         computer = change['value']
+        with self.output:
+            clear_output()
+            if not computer:
+                return None
+            if isinstance(computer, str):
+                if computer in self.computers:
+                    return self.computers[computer]
+                print(f"No computer named '{computer}' was found in AiiDA database.")
 
-        if not computer:
-            return None
-        if isinstance(computer, str):
-            if computer in self.computers:
-                return self.computers[computer]
-            raise KeyError(f"No computer named '{computer}' was found in AiiDA database.")
+            if isinstance(computer, Computer):
+                if computer.name in self.computers:
+                    return computer
+                print(f"The computer instance '{computer}' supplied was not found in  the AiiDA database. "
+                      "Consider reloading")
 
-        if isinstance(computer, Computer):
-            if computer.name in self.computers:
-                return computer
-            raise ValueError(f"The computer instance '{computer}' supplied was not found in  the AiiDA database. "
-                             "Consider reloading")
-
-        # This place will never be reached, because the trait's type is checked before validation.
         return None
--- a/aiidalab_widgets_base/computers.py
+++ b/aiidalab_widgets_base/computers.py
@@ -690,7 +690,7 @@ class ComputerDropdown(ipw.VBox):
         path_to_root (str): Path to the app's root folder.
         """
 
-        self.output = ipw.Output()
+        self.output = ipw.HTML()
 
         self._dropdown = ipw.Dropdown(options={},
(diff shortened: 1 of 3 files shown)


aiidalab_widgets_base/computers.py
```diff
@@ -67,50 +65,53 @@ def __init__(self, **kwargs):
...
-                                            description="Proxy server username:",
-                                            layout=ipw.Layout(width="350px"),
-                                            style=STYLE)
+        inp_proxy_username = ipw.Text(value='',
+                                      description="Proxy server username:",
+                                      layout=ipw.Layout(width="350px"),
+                                      style=STYLE)
+        link((inp_proxy_username, 'value'), (self, 'proxy_username'))
+
         self._inp_proxy_password = ipw.Password(value='',
```
csadorf, line 88: Any reason why this is a private member, but `inp_proxy_username` for example is not?

aiidalab_widgets_base/computers.py
```diff
@@ -450,161 +455,145 @@ def __private_key(self):
...
-        return self._inp_username.value
-
-    @username.setter
-    def username(self, username):
-        self._inp_username.value = username
+    @validate('proxy_hostname')
+    def _validate_proxy_hostname(self, provided):
+        """Enable 'use proxy' widget if proxy hostname is provided."""
+        if provided['value']:
+            self.use_proxy = True
```
csadorf, line 462: Change of state should happen in the `observe` function, not in the validate function.

aiidalab_widgets_base/computers.py
```diff
@@ -450,161 +455,145 @@ def __private_key(self):
...
 
-    @property
-    def port(self):
-        return str(self._inp_port.value)
+    @validate('proxy_username')
+    def _validate_proxy_username(self, provided):
+        """Enable 'use proxy' and 'use different proxy username' widgets if proxy username is provided."""
+        if provided['value']:
+            self.use_proxy = True
+            self._use_diff_proxy_username.value = True
```
csadorf, line 470: Same comment as above.

aiidalab_widgets_base/computers.py
```diff
@@ -450,161 +455,145 @@ def __private_key(self):
...
+
+        # Mpirun command.
+        inp_mpirun_cmd = ipw.Text(value='mpirun -n {tot_num_mpiprocs}',
+                                  description="Mpirun command:",
+                                  layout=ipw.Layout(width="500px"),
+                                  style=STYLE)
+        link((inp_mpirun_cmd, 'value'), (self, 'mpirun_command'))
+
+        # Number of CPUs per node.
+        inp_computer_ncpus = ipw.IntText(value=12,
```
csadorf, line 530: Why is 12 the default?

csadorf (changes requested):
I have a few minor change requests, but looks overall fine. However, I did not have time to tests this in the browser

aiidalab_widgets_base/computers.py
yakutovicha: Any number is fine. I can also set it to 1. I don't have any preferences here.

aiidalab_widgets_base/computers.py
yakutovicha: I keep `_inp_proxy_password` as a class member because I access it later. It is private for the obvious reason - I want to prevent other classes to rely on it.

`inp_proxy_username` is not even a class member, as I don't want to expose the widget for the modifications.

aiidalab_widgets_base/computers.py
yakutovicha: ok

aiidalab_widgets_base/computers.py
yakutovicha: ok

PR #159, opened 2020-12-01
Fix pybel import from openbabel.

csadorf:


aiidalab_widgets_base/structures.py
```diff
@@ -555,7 +555,7 @@ def _optimize_mol(self, mol):
 
         # Note, the pybel module imported below comes together with openbabel package. Do not confuse it with
         # pybel package available on PyPi: https://pypi.org/project/pybel/
-        import pybel  # pylint:disable=import-error
+        from openbabel import pybel  # pylint:disable=import-error
```
yakutovicha, line 558:
```suggestion
        try:
            import pybel  # pylint:disable=import-error
        except ModuleNotFoundError:
            from openbabel import pybel  # pylint:disable=import-error
```

aiidalab_widgets_base/structures.py
```diff
@@ -578,7 +578,7 @@ def _on_button_pressed(self, change):  # pylint: disable=unused-argument
...
```
yakutovicha, line 581:
```suggestion
        try:
            import pybel  # pylint:disable=import-error
        except ModuleNotFoundError:
            from openbabel import pybel  # pylint:disable=import-error
```

yakutovicha (commented):
thanks, @csadorf for catching this.

As we cannot control the open babel version which is installed through conda, maybe use try-except (see suggestions).

csadorf:
> As we cannot control the open babel version which is installed through conda, maybe use try-except (see suggestions).

I am not necessarily against that, but why do you say that we cannot control that?

https://github.com/aiidalab/aiidalab-docker-stack/blob/a7a0b9e4077f7c8f66cc25ec1d2e46cefa852a2f/Dockerfile#L103

yakutovicha:
as we discussed, the issue has been resolved in #158

PR #160, opened 2020-12-02
Remove scikit-learn as a dependancy

chrisjsewell:
In the last commit 1abe24b @cpignedoli removed the only import of sklearn, so I don't think this is any longer a dependency

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -16,8 +16,7 @@
             "nglview",
             "numpy",
             "optimade-client",
-            "pandas",
-            "scikit-learn"
+            "pandas"
         ]
     }
 }
--- a/setup.json
+++ b/setup.json
@@ -20,8 +20,7 @@
         "nglview",
         "numpy",
         "optimade-client",
-        "pandas",
-        "scikit-learn"
+        "pandas"
     ],
     "extras_require": {
         "docs": [
--- a/setup.json
+++ b/setup.json
@@ -23,6 +23,10 @@
         "pandas"
     ],
     "extras_require": {
+        "smiles": [
+            "openbabel",
+            "rdkit"
+        ],
         "docs": [
             "sphinx",
             "sphinx-rtd-theme",


chrisjsewell:
I've also added an extras section in setup.json for `smiles`

yakutovicha:
> I've also added an extras section in setup.json for `smiles`

@chrisjsewell, I am not sure openbabel and rdkit can be installed through pip. Did you try that?

chrisjsewell:
> Did you try that?

hmm, no rdkit can't.
But I actually think there should still be this extra, just as a top-level means to declare the dependencies, as opposed to having to hunt in the code for them.

chrisjsewell:
or optionally provide a conda environment.yml with them in

yakutovicha:
> In the last commit [1abe24b](https://github.com/aiidalab/aiidalab-widgets-base/commit/1abe24b83cd3defc74342e09de1e1d8470553160) @cpignedoli removed the only import of sklearn, so I don't think this is any longer a dependency

I don't think it is true: https://github.com/aiidalab/aiidalab-widgets-base/blob/1abe24b83cd3defc74342e09de1e1d8470553160/aiidalab_widgets_base/structures.py#L552-L560

chrisjsewell:
ah oops, given its in a try/except, does this mean its optional now though?

yakutovicha:
> ah oops, given its in a try/except, does this mean its optional now though?

No, I think we should rather remove try/except. We used those try/except to prevent problems when openbabel and rdkit aren't installed (as they can only be installed through conda). However, sklearn can be installed through pypi, so try/except is unnecessary.

@chrisjsewell can you remove this particular try/except, please?

csadorf:
@yakutovicha @chrisjsewell What's the status on this?

yakutovicha:
closing in favour of #164

PR #167, opened 2021-01-20
ProcessInputs(Outputs)Widget change upon the process

unkcpz:
make the widget changed on the process so it can be used in the same notebook of generating process and to detect the inputs and results of it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,6 @@ venv.bak/
 
 # vim swap files
 *.swp
+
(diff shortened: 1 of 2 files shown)


csadorf:
@unkcpz What is the status of this PR? You never requested a review, is this still relevant?

unkcpz:
I think this is addressed after you refactor the code. I'll just close this since it is not a problem for me now.

PR #168, opened 2021-01-25
Allow configuring interval between connections during computer setup.

### LINKED ISSUE #141: settings for minimum poll
the configuration of a computer should have a default setting for minimum poll
so that a user does not need to do manually:

verdi computer configure ssh computer-name --safe-interval 120 -n

verdi shell
c=load_computer('computer_name')
c.set_minimum_job_poll_interval(120)

https://github.com/aiidalab/aiidalab-widgets-base/blob/efad328d0f83d099fdc529792201a946b48110fa/aiidalab_widgets_base/computers.py#L416

yakutovicha:
fixes #141

### DIFF
--- a/aiidalab_widgets_base/computers.py
+++ b/aiidalab_widgets_base/computers.py
@@ -9,7 +9,7 @@ import pexpect
 import shortuuid
 import ipywidgets as ipw
 from IPython.display import clear_output
-from traitlets import Bool, Dict, Instance, Int, Unicode, Union, link, observe, validate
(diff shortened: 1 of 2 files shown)


yakutovicha:
@cpignedoli let me know if this satisfies your needs.

cpignedoli:
I would suggest 60 as value, it should be safe and 1 minute waiting time is OK

yakutovicha:
> I would suggest 60 as value, it should be safe and 1 minute waiting time is OK

As we discussed elsewhere, we will keep 30 (same as AiiDA default).

PR #171, opened 2021-02-12
added mirroring of atoms

cpignedoli:


yakutovicha (commented):
Thanks @cpignedoli, I have a couple of comments:

1. Instead of 3 points, can we use vector and point? So that the action vector is perpendicular to the plane, while the action point is within the plane.
2. Can we avoid the `if-else` statement? It appears to be redundant.
3. I think there should be a warning if the action vector has 0 lengths. Otherwise, the results are unpredictable.

yakutovicha:
@cpignedoli I think the problems I mentioned in my review are now all solved.

yakutovicha (approved):

PR #172, opened 2021-02-14
added copy atoms

cpignedoli:
a selection of atoms can be copied and will appear shifted by 0.5A in x (same convention as some commercial editors)
selection will be set to the newly added atoms so that they can be easily identified and moved

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -835,7 +835,9 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
         link((use_covalent_radius, "value"), (self.bond_length, "disabled"))
 
         # Copy atoms
-        btn_copy_sel = ipw.Button(description='Copy selected', layout={'width': 'initial'})
+        btn_copy_sel = ipw.Button(
+            description="Copy selected", layout={"width": "initial"}
+        )
         btn_copy_sel.on_click(self.copy_sel)
 
         # Modify atom.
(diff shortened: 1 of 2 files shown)


yakutovicha:
@cpignedoli I modified the code so that the insertion happens not 0.5 A along the "real" X-axis, but 1.0 angstrom along the "screen" X-axis. 

Let me know if this is Ok with you.

cpignedoli:
no this is not ideal
because the user as then no control on exact positioning... typically when i copy a portion of an object i have in mind an exact position where to put it so i know i must shift by 0.5 in x and then.... That is why in MStudio they did like this

yakutovicha:
are you sure? In some cases, +0.5 along X means inserting atoms in some other atoms - so the whole thing becomes a mess. With the approach I implemented though, the copied atoms always go to the right with respect to the user's point of view.

cpignedoli:
that is why selection is set to the copied atoms...

yakutovicha:
Maybe we should be able to let the users select where it goes? Along action vector?

cpignedoli:
Keep it simple, the Materials Studio solution 
is immediate and I never had to complain about it, that is why I did exactly the same...

yakutovicha (approved):
Alright, so after a discussion with @cpignedoli I agree that moving atoms along X-direction is better than doing the movement with respect to the camera position. Despite, it doesn't always look nice, it is well defined. Any attempts at defining a moving vector will result in a more complicated code/interface.
Therefore, I am going to approve the PR in its current state.

yakutovicha:
On a side note: the pinhole camera implementation is added in a different PR: https://github.com/aiidalab/aiidalab-widgets-base/pull/193

yakutovicha:
The tests are not working properly, but the problem appears to be elsewhere.

PR #174, opened 2021-02-15
Fix accidental duplication of method name.

csadorf:


yakutovicha (approved):
oh, thanks a lot @csadorf!

PR #175, opened 2021-02-15
Misc/black formatting issue 133

csadorf:


CasperWA (commented):
Great!

I would suggest to update `setup.json` accordingly as well. Remove `yapf` and `prospector`. And update `pre-commit` to the latest version (2.10.1).

Also, you could try your luck with adding the [following lines](https://github.com/aiidateam/aiida-optimade/blob/develop/.github/workflows/ci.yml#L29:L35) to the CI job as well. It basically extracts the specified version from a requirements syntaxed file and uses it to install `pre-commit`. The extraction logic would need to be altered for use with the JSON, but otherwise all good.

Finally, this is a bit out-of-scope, but sort of related to the changes. You could add other relevant pre-commit hooks as well, like `check-yaml`, `check-json`, etc. as done in the [repo you're referencing](https://github.com/CasperWA/voila-optimade-client/blob/develop/.pre-commit-config.yaml). But again. It's slightly out-of-scope for this PR, but could probably be defended, I think?

yakutovicha (commented):
I have no experience with configuring black, so I would let @CasperWA review this PR. 

What concerns the switching to black - I am fine with it, as soon as we agree that all the other codes in this organization will switch to the same formatting.

csadorf:
@CasperWA Thx for your feedback. I applied the suggested changes. 👍

.pre-commit-config.yaml
```diff
@@ -1,20 +1,20 @@
-# # Install pre-commit hooks via
-# pre-commit install
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.4.0
+    hooks:
+      - id: check-yaml
+      - id: check-json
+      - id: trailing-whitespace
+        exclude: miscellaneous/structures
```
CasperWA, line 8: You probably want to exclude markdown files as well, as trailing spaces are valid syntax (double trailing spaces creates a new line, but not a new section).

README.md
```diff
@@ -6,7 +6,7 @@
 
 Install the `aiidalab_widgets_base` python package via:
 ```
-pip install aiidalab-widgets-base 
+pip install aiidalab-widgets-base
```
CasperWA, line 9: ... although it nicely caught this useless trailing space 😅

setup.json
```diff
@@ -31,10 +31,9 @@
             "sphinxcontrib-details-directive"
         ],
         "pre-commit": [
-            "pre-commit==1.17.0",
-            "yapf==0.28.0",
-            "prospector==1.1.7",
-      	    "pylint==2.3.1"
+            "black==20.8b1",
+            "flake8==3.8.4",
```
CasperWA, line 35: There's actually no need to install these, as they are installed in special environments in connection with `pre-commit`. I.e., `pre-commit` will take care of installing and running them all on its own using the versions specified in the config file.

Installing them separately here is only relevant if you want to run them separately as well (and not as part of `pre-commit`), which I don't think is necessary.
```suggestion
```

CasperWA: Having `pylint` is nice though still as most IDEs use it.

CasperWA (changes requested):

setup.json
csadorf: Right, I forgot that pre-commit actually installs the dependencies. I'm a bit confused by your suggestion though. Do you want to keep it or not? I actually do run `black` and other linters manually in addition to the hook.

.pre-commit-config.yaml
csadorf: I'd consider that exotic enough to simply not allow unless needed.

README.md
csadorf: Yeah, I'd prefer to lint the markdown files as well (see comment above).

.pre-commit-config.yaml
CasperWA: Ah, I use it often though :) But fair enough.

setup.json
CasperWA: The suggested change here is to remove it.
If you run `black` ouside of `pre-commit` often enough that it's useful, then it's fine to keep it - since it will get the same level of "acceptance" as `pylint`.

.pre-commit-config.yaml
csadorf: Ok, then let's leave _as is_.

setup.json
```diff
@@ -33,10 +33,7 @@
             "sphinxcontrib-details-directive"
         ],
         "pre-commit": [
-            "pre-commit==1.17.0",
-            "yapf==0.28.0",
-            "prospector==1.1.7",
-      	    "pylint==2.3.1"
+            "pre-commit==2.10.1"
```
CasperWA, line 36: Ah, I see you opted to leave out `black` and `pylint` completely. That's also fine.

CasperWA (approved):

PR #179, opened 2021-03-03
Use decorator for the registration of AiiDA data viewer widgets.

csadorf:
Simplifies registration of new viewers.


yakutovicha (approved):
very nice, thanks @csadorf!

PR #195, opened 2021-04-14
Add ipytree dependency.

csadorf:


chrisjsewell:
as we are already working on,
it is of note that ipytree likely needs to be installed in both the aiida environment (to use the python API) and jupyter environment (to load the notebook extension) to work

and the same for ipywidgets-extended

csadorf:
> as we are already working on,
> it is of note that ipytree likely needs to be installed in both the aiida environment (to use the python API) and jupyter environment (to load the notebook extension) to work
> 
> and the same for ipywidgets-extended (loaded into the aiida environment by optimade-client)

Both "aiida/python-kernel" dependencies and the "special" jupyter dependencies are going to be explicitly declared in the future, see here: https://github.com/aiidalab/aiidalab-registry/pull/60

yakutovicha (approved):

PR #197, opened 2021-05-06
Base DictViewer on ipw.VBox widget.

yakutovicha:
Previously, DictViewer was derived from ipw.HTML widget which made
it impossible to incorporate another widgets into it.


csadorf (approved):
LGTM.

PR #198, opened 2021-05-06
Prepare release 1.0.0b16

yakutovicha:


csadorf (approved):
1. The `aiida-core` dependency could be expressed as `aiida-core~=1.0`.
2. Overall I think we should be more specific with all our dependencies.
3. In case of `vapory` we should probably just depend on a specific commit.

I'm ok with leaving it _as is_ for this particular release.

yakutovicha:
Thanks, @csadorf, I made an issue #199 to remember your suggestions.

PR #201, opened 2021-05-11
Set version ranges for the dependencies.

### LINKED ISSUE #146: Provide version ranges for the dependencies.
Also, we might consider moving some of the dependencies of aiidalab package here.

### LINKED ISSUE #178: New dependency 'vapory' is unconstrained and undeclared
A new dependency ([vapory](https://github.com/Zulko/vapory)) was introduced in commit 917b2df:

https://github.com/aiidalab/aiidalab-widgets-base/blob/917b2df558957747ac6cc1d9c6ffa1b89f8ab172/setup.json#L26

I recommend to:
- [ ] To pin to a specific commit instead `@master` to avoid problems with accidentally pulling incompatible upstream changes. 
- [ ] Add the dependency to https://github.com/aiidalab/aiidalab-widgets-base/blob/develop/metadata.json
- [ ] Also add vapory as a dependency to https://github.com/aiidalab/aiidalab-docker-stack

yakutovicha:
fixes #178 
fixes #146

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -7,17 +7,17 @@
     "requires": {
       "~=1.0b": [
             "aiida-core>=1.0.0,<2.0.0",
-            "ase",
-            "bokeh",
-            "fileupload",
-            "ipytree",
-            "ipywidgets",
-            "more_itertools",
-            "nglview",
-            "numpy",
-            "optimade-client",
-            "pandas",
-            "scikit-learn"
+            "ase<3.20",
+            "bokeh~2.0",
+            "ipytree~=0.2",
+            "ipywidgets<8.0.0",
+            "more_itertools~=8.0",
+            "nglview~=3.0",
+            "numpy~=1.17",
+            "optimade-client==2021.5.7",
+            "pandas~=1.0",
+            "scikit-learn~=0.24",
+            "aiidalab-vapory==0.1.2"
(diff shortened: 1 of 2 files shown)


metadata.json
```diff
@@ -7,17 +7,17 @@
...
-            "more_itertools",
-            "nglview",
-            "numpy",
-            "optimade-client",
-            "pandas",
-            "scikit-learn"
+            "ase<3.20",
+            "bokeh~=2.0",
+            "ipytree~=0.2",
+            "ipywidgets<8.0.0",
```
csadorf, line 13: I would think that this would be reasonable:
```suggestion
            "ipywidgets~=7.6.0",
```

metadata.json
```diff
@@ -7,17 +7,17 @@
     "requires": {
       "~=1.0b": [
             "aiida-core>=1.0.0,<2.0.0",
```
csadorf, line 9: Why not?
```suggestion
            "aiida-core~=1.0",
```

csadorf (commented):
Two suggestions, but looks good.

As a next step (separate PR): Can we convert the package to use `setup.cfg` instead of `setup.json`? (see recent updates to aiidalab-qe, aiidalab-hello-world, and so on).

PR #211, opened 2021-07-12
Curate providers for OptimadeQueryWidget

### LINKED ISSUE #203: Disable providers and certain Materials Cloud databases for OPTIMADE widget
Some providers and Materials Cloud databases are either not compliant with the current version of the OPTIMADE client or they're not possible to query/shouldn't be queryable in production.

CasperWA:
Closes #203.

Extend the initialization of `OptimadeQueryProviderWidget` and
`OptimadeQueryFilterWidget` with relevant parameters, popping them from
the `kwargs` with default fallbacks being either `None` or "private"
variables defined for the `OptimadeQueryWidget` class.
These variables mirror what is used for the OPTIMADE Client on Materials
Cloud Tools, curating the providers and Materials Cloud databases to
show. Furthermore, it groups the Materials Cloud databases into
"General" and "Projects" according to what has been previously decided
with Nicola.

### DIFF
--- a/aiidalab_widgets_base/databases.py
+++ b/aiidalab_widgets_base/databases.py
@@ -148,25 +148,65 @@ class OptimadeQueryWidget(ipw.VBox):
 
     structure = Instance(Atoms, allow_none=True)
 
+    _disable_providers = [
+        "cod",
+        "tcod",
+        "nmd",
+        "omdb",
+        "oqmd",
+        "aflow",
+        "matcloud",
+        "httk",
+        "mpds",
+        "necro",
+    ]
+    _skip_databases = {"Materials Cloud": ["optimade-sample", "li-ion-conductors"]}
+    _database_grouping = {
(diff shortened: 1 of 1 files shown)


CasperWA:
Why are the `latest` required here instead of `stable`?

csadorf:
> Why are the `latest` required here instead of `stable`? (Talking about the CI.)

The `latest` tag is currently not well-defined, I suggest to remove the tests for that.

CasperWA:
> > Why are the `latest` required here instead of `stable`? (Talking about the CI.)
> 
> The `latest` tag is currently not well-defined, I suggest to remove the tests for that.

Right, so this means this PR is essentially passing? I'll switch the "Required" CI tests to the `stable` ones then.

yakutovicha (approved):
Thanks, @CasperWA, looks good to me!

PR #181, opened 2021-03-05
AiiDAlab-ELN ingegration.

yakutovicha:

PR #213, opened 2021-07-19
Add AiidaNodeViewWidget class.

yakutovicha:
Convenience class that can always be displayed with the `display` function.
Unlike viewer, `DisplayAiidaObject` always has a `node` trait that can be provided
later. Once the node is provided, the `DisplayAiidaObject` object will automatically
show it.

(extracted from #181)

### DIFF
--- a/aiidalab_widgets_base/__init__.py
+++ b/aiidalab_widgets_base/__init__.py
@@ -34,12 +34,13 @@ from .structures import (
     StructureManagerWidget,
     StructureUploadWidget,
 )
-from .viewers import AiidaNodeViewWidget, register_viewer_widget, viewer  # noqa
+from .viewers import AiidaNodeViewWidget, register_viewer_widget, viewer
 from .wizard import WizardAppWidget, WizardAppWidgetStep
 
 __all__ = [
     "AiiDACodeSetup",
     "AiidaComputerSetup",
+    "AiidaNodeViewWidget",
     "BasicStructureEditor",
     "CodQueryWidget",
     "CodeDatabaseWidget",
--- a/aiidalab_widgets_base/__init__.py
+++ b/aiidalab_widgets_base/__init__.py
@@ -34,7 +34,7 @@ from .structures import (
     StructureManagerWidget,
     StructureUploadWidget,
 )
-from .viewers import register_viewer_widget, viewer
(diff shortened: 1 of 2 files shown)


aiidalab_widgets_base/viewers.py
```diff
@@ -79,6 +80,29 @@ def viewer(obj, downloadable=True, **kwargs):
         raise exc
 
 
+class DisplayAiidaObject(ipw.VBox):
```
csadorf, line 83:
```suggestion
class AiidaNodeViewWidget(ipw.VBox):
```

csadorf: Do you want to expose the widget in the package root namespace?

csadorf (changes requested):
I've made a strong suggestion with respect to the class name. Otherwise, I know it works because of: https://github.com/aiidalab/aiidalab-qe/blob/8c05ee92d166dc894acbe4189ff932595e5f7e76/aiidalab_qe/widgets.py#L217-L231

aiidalab_widgets_base/viewers.py
yakutovicha: > Do you want to expose the widget in the package root namespace?

right, will do.

aiidalab_widgets_base/__init__.py
```diff
@@ -34,7 +34,7 @@
     StructureManagerWidget,
     StructureUploadWidget,
 )
-from .viewers import register_viewer_widget, viewer
+from .viewers import AiidaNodeViewWidget, register_viewer_widget, viewer  # noqa
```
csadorf, line 37: Why the noqa?

yakutovicha: I just received a random complaint from the pre-commit: 
Screenshot 2021-07-19 at 11 25 07

So this is just to silence it. 

Not sure why it wasn't complaining about other stuff, though.

csadorf: You forgot to add it to the `__all__` list. I've pushed a fix.

yakutovicha: ah, right, sorry :)

csadorf (approved):

PR #216, opened 2021-07-19
CI: Add GitHub actions workflow to release package on (Test) PyPI.

csadorf:


yakutovicha (approved):
I assume it is going to work, as the same [publish workflow](https://github.com/aiidalab/aiidalab/blob/develop/.github/workflows/publish.yml) works fine for the aiidalab repo.

PR #217, opened 2021-07-19
StructureManagerWidget: connect structure_node trait to the viewer.

yakutovicha:
Previously, it was structure trait that was connected to a viewer.
However, structure trait is not an AiiDA node, therefore it does
not contain the provenance information.

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -92,7 +92,7 @@ class StructureManagerWidget(ipw.VBox):
             self.viewer = viewer
         else:
             self.viewer = StructureDataViewer(downloadable=False)
-        dlink((self, "structure"), (self.viewer, "structure"))
+        dlink((self, "structure_node"), (self.viewer, "structure"))
 
         # Store button.
         self.btn_store = ipw.Button(description="Store in AiiDA", disabled=True)


csadorf:
> Previously, it was structure trait that was connected to a viewer.
> However, structure trait is not an AiiDA node, therefore it does
> not contain the provenance information.

But the `StructureViewer` class expects either Atoms or Structure nodes?

https://github.com/aiidalab/aiidalab-widgets-base/blob/d4b7500e962635f465b6673454a01e16a019bc2c/aiidalab_widgets_base/viewers.py#L581

yakutovicha:
> > Previously, it was structure trait that was connected to a viewer.
> > However, structure trait is not an AiiDA node, therefore it does
> > not contain the provenance information.
> 
> But the `StructureViewer` class expects either Atoms or Structure nodes?
> 
> https://github.com/aiidalab/aiidalab-widgets-base/blob/d4b7500e962635f465b6673454a01e16a019bc2c/aiidalab_widgets_base/viewers.py#L581

Yes, it can deal with both. It is just that `structure_node` has more information attached to it.

csadorf (approved):
I'm not sure if this change has any other implications. I know that there was a lot of fiddling with those structure-related widgets and data exchange. So I am going to trust that you have taken those into account.

yakutovicha:
> I'm not sure if this change has any other implications. I know that there was a lot of fiddling with those structure-related widgets and data exchange. So I am going to trust that you have taken those into account.

I am going to need this change for the ELN integration to simplify the life of the users. It is nice to have the provenance information attached to the structure. This information is further used to send the computed information back to the original place.

PR #218, opened 2021-07-19
Add OpenAiidaNodeInAppWidget

yakutovicha:
This widget simplifies the opening of a node in the selected app.

(extracted from #181)

### DIFF
--- a/aiidalab_widgets_base/nodes.py
+++ b/aiidalab_widgets_base/nodes.py
@@ -11,9 +11,61 @@ from aiida.orm import (
     WorkChainNode,
     load_node,
 )
+from aiidalab.app import AiidaLabApp
+from aiidalab.config import AIIDALAB_APPS
+from aiidalab.utils import load_app_registry
 from IPython.display import clear_output, display
 from ipytree import Node as TreeNode
 from ipytree import Tree
+
+CALCULATION_TYPES = [
+    (
+        "geo_opt",
+        "Geometry Optimization",
+        "Geometry Optimization - typically this is the first step needed to find optimal positions of atoms in the unit cell.",
+    ),
+    (
+        "geo_analysis",
+        "Geometry analysis",
+        "Geometry analysis - calculate parameters describing the geometry of a material.",
+    ),
+    (
+        "isotherm",
+        "Isotherm",
+        "Isotherm - compute adsorption isotherm of a small molecules in the selected material.",
+    ),
+]
+
+SELECTED_APPS = [
+    {
+        "name": "quantum-espresso",
+        "calculation_type": "geo_opt",
+        "notebook": "qe.ipynb",
+        "parameter_name": "structure_uuid",
+        "description": "Optimize atomic positions and/or unit cell employing Quantum ESPRESSO. Quantum ESPRESSO is preferable for small structures with no cell dimensions larger than 15 Å. Additionally, you can choose to compute electronic properties of the material such as band structure and density of states.",
+    },
+    {
+        "name": "aiidalab-lsmo",
+        "calculation_type": "geo_opt",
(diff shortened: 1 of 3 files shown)


csadorf:
Could you provide an example for how to use this?

yakutovicha:
> Could you provide an example for how to use this?

yes, but it will come in the next PR together with ELN widgets.

csadorf:
> > Could you provide an example for how to use this?
> 
> yes, but it will come in the next PR together with ELN widgets.

Ok, please re-request my review once those are ready, thank you!

yakutovicha:
@csadorf, I think it is ready for your review. I've implemented all I had in mind for this PR.

The example of usage can be found in #220, in the [eln_import notebook](https://github.com/aiidalab/aiidalab-widgets-base/pull/220/files#diff-6b7f94e60f827bc73b705fe083b57539ccd2f61588db698d49d2dc5132443a88).

aiidalab_widgets_base/nodes.py
```diff
@@ -210,3 +264,88 @@ def find_node(self, pk):
...
+        app_object = AiidaLabApp(name, app_data, AIIDALAB_APPS)
+        icon = os.path.splitext(app_object.url)[0]
+
+        # We expect it to always be a git repository
+        icon += "/master/" + app_object.metadata["logo"]
+        if "github.com" in icon:
+            icon = icon.replace("github.com", "raw.githubusercontent.com")
+            if icon.endswith(".svg"):
+                icon += "?sanitize=true"
+        self.icon = icon
```
csadorf, line 283: It would be much easier if the logo URL would simply always be a fully-qualified URL. Then there is no need for all of this guesswork here. The [AiiDAlab registry](https://github.com/aiidalab/aiidalab-registr) already does not try to guess the logo path, but simply uses the provided URLs.

aiidalab_widgets_base/nodes.py
```diff
@@ -210,3 +264,88 @@ def find_node(self, pk):
...
                 return node
         raise KeyError(pk)
+
+
+class _AppIcon:
+    def __init__(self, app, app_registry, path_to_root, node):
+
+        name = app["name"]
+        app_data = app_registry["apps"][name]
+        app_object = AiidaLabApp(name, app_data, AIIDALAB_APPS)
```
csadorf, line 274: Since you don't need a widget, it would probably be better to use the "internal" [`_AiidaLabApp` class](https://github.com/aiidalab/aiidalab/blob/45dfabb46f033c0d08db11a7a3ae435dcf7a5c01/aiidalab/app.py#L58). This one can also be much easier instantiated with [`.from_id(<app_name>)`](https://github.com/aiidalab/aiidalab/blob/45dfabb46f033c0d08db11a7a3ae435dcf7a5c01/aiidalab/app.py#L78). See the [`__main__` module](https://github.com/aiidalab/aiidalab/blob/45dfabb46f033c0d08db11a7a3ae435dcf7a5c01/aiidalab/__main__.py#L173) for examples.

csadorf (changes requested):

aiidalab_widgets_base/nodes.py
yakutovicha: I would like to use the `is_installed` function that is only available in the `AiidaLabApp` class. In principle, if it is ok with you to move that function in the `_AiidaLabApp` class, then I can use it.

yakutovicha:
@csadorf, thanks for the review. I addressed both of your comments. Those changes rely on the latest update of the aiidalab app: https://github.com/aiidalab/aiidalab/pull/194.

yakutovicha:
I am going to merge that PR as well.

PR #220, opened 2021-07-19
Add widgets to interact with ELNs

yakutovicha:
To be merged after the #218 

(extracted from #181)

### DIFF
--- a/aiidalab_widgets_base/elns.py
+++ b/aiidalab_widgets_base/elns.py
@@ -133,6 +133,7 @@ class ElnExportWidget(ipw.VBox):
 class ElnConfigureWidget(ipw.VBox):
     def __init__(self, **kwargs):
         self._output = ipw.Output()
+        self.eln = None
 
         self.eln_instance = ipw.Dropdown(
             description="ELN:",
@@ -174,6 +175,8 @@ class ElnConfigureWidget(ipw.VBox):
         )
         check_connection.on_click(self.check_connection)
 
+        self.my_output = ipw.HTML()
+
         self.display_eln_config()
 
         super().__init__(
@@ -182,6 +185,7 @@ class ElnConfigureWidget(ipw.VBox):
                 self.eln_types,
                 self._output,
                 ipw.HBox([default_button, save_config, erase_config, check_connection]),
+                self.my_output,
             ],
             **kwargs,
         )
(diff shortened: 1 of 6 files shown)


aiidalab_widgets_base/elns.py
```diff
@@ -113,7 +119,7 @@ def _observe_node(self, _=None):
     def send_to_eln(self, _=None):
         if self.eln.is_connected:
             self.error_message.value = ""
-            self.eln.export_data_object(data_object=self.node)
+            self.eln.export_data()
         else:
             self.error_message.value = f"""Warning! The access to ELN is not configured. Please follow <a href="{self.path_to_root}/aiidalab-widgets-base/eln_configure.ipynb" target="_blank">the link</a> to configure it."""
```
kjappelbaum, line 124: I would reuse the same message as above, i.e., also name the ELN instance

aiidalab_widgets_base/elns.py
```diff
@@ -0,0 +1,285 @@
...
+    parents=True, exist_ok=True
+)  # making sure that the folder exists.
+
+
+def connect_to_eln(eln_instance=None, **kwargs):
+    try:
+        with open(ELN_CONFIG, "r") as file:
+            config = json.load(file)
+    except (FileNotFoundError, json.JSONDecodeError, KeyError):
+        return None
```
kjappelbaum, line 22: Not sure how else this could fail, but do we want to propagate a more detailed exception to the user instead of returning?

kjappelbaum, line 22: The warning message `""Warning! The access to ELN {kwargs['eln_instance']} is not configured. Please follow <a href="{url}" target="_blank">the link</a> to configure it.` users get here (in case reading the config fails and you return `None`) might not be so clear.

kjappelbaum (dismissed):
lgtm, many thanks!

aiidalab_widgets_base/elns.py
yakutovicha: We had a separate discussion with @kjappelbaum regarding this comment. It turns out that if the file is erroneously formatted - it will be erased. So the warning message still holds since the user will have to configure the ELN access again. One might argue that this is not the best behaviour, but it looks reasonable to us.

PR #229, opened 2021-09-12
include jupyter notebooks in docs

ltalirz:
* move jupyter notebooks into `notebooks/` subfolder (were difficult to find otherwise)
 * Include Jupyter notebooks that illustrate the use of the widgets in the docs

### DIFF
--- a/docs/source/widget-list/index.rst
+++ b/docs/source/widget-list/index.rst
@@ -2,7 +2,7 @@
 List of widgets
 ***************
 
-Here we collect a set of Jupyter notebooks that illustrate how to use all available widgets.
+This is a collection of Jupyter notebooks that illustrate how to use the widgets available from this library.
 
 .. toctree::
     :titlesonly:
--- a/docs/source/widget-list/index.rst
+++ b/docs/source/widget-list/index.rst
@@ -2,15 +2,20 @@
 List of widgets
 ***************
 
-Here we collect the information about all the widgets available in AWB and provide a simple example of how to use them.
+Here we collect a set of Jupyter notebooks that illustrate how to use all available widgets.
 
 .. toctree::
-    :maxdepth: 1
+    :titlesonly:
+    :glob:
 
-    Codes <codes>
-    Computers <computers>
-    External databases <databases>
-    Export AiiDA DB <export>
-    Nodes <nodes>
-    Processes <processes>
-    Structures <structures>
+    notebooks/aiida_datatypes_viewers
+    notebooks/codes_computers
+    notebooks/setup_computer
+    notebooks/setup_code
+    notebooks/process
+    notebooks/process_list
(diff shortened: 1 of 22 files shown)


ltalirz:
Example:
image


Note that the notebooks currently don't contain outputs.
We may want to change that in the future

ltalirz:
If you want to test it locally, just `pip install -e --use-deprecated=legacy-resolver .[docs]`  and type `make`  in the docs dir

docs/source/widget-list/index.rst
```diff
@@ -2,15 +2,20 @@
 List of widgets
 ***************
 
-Here we collect the information about all the widgets available in AWB and provide a simple example of how to use them.
+Here we collect a set of Jupyter notebooks that illustrate how to use all available widgets.
```
csadorf, line 5:
```suggestion
This is a collection of Jupyter notebooks that illustrate how to use the widgets available from this library.
```

setup.cfg
```diff
@@ -45,6 +45,7 @@ docs =
     sphinx-rtd-theme
     sphinxcontrib-contentui
     sphinxcontrib-details-directive
+    nbsphinx
```
csadorf, line 48: When I tried to build the docs I also needed to install pandoc, I assume that's related to this? Anyway to document this?

notebooks/eln_configure.ipynb
```diff
@@ -1,5 +1,15 @@
 {
  "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "understanding-authority",
+   "metadata": {},
+   "source": [
+    "# Configure connection to ELN\n",
+    "\n",
+    "Authenticate the AiiDAlab with an electronic laboratory notebook"
```
csadorf, line 10:
```suggestion
    "Authenticate AiiDAlab with an Electronic Laboratory Notebook (ELN)"
```

csadorf (approved):
It would indeed be much better if we also showed the output of those notebooks, but ok to add this in a separate PR.

I have a few suggestions for slightly different wording, but overall good.

setup.cfg
ltalirz: I see, thanks for pointing this out. I'll document it in the "development installation" 

I'll check, and I'll also check with Chris whether there are now tools that could avoid the pandoc dependency.

ltalirz: @csadorf I now switched to the myst route which should get rid of the pandoc dependency.

ltalirz:
> It would indeed be much better if we also showed the output of those notebooks, but ok to add this in a separate PR.

Let's do this in a separate PR.

Both nbsphinx and myst-nb have machinery for running the notebooks automatically, so we can look into enabling it (although it does not work out of the box)

csadorf (approved):

PR #236, opened 2021-09-15
fix: issues encountered while running notebooks

ltalirz:
* Switch from computer.name to computer.label
* Remove obsolete multi-structure notebook


ltalirz:
P.S. Hm, it turns out that the html version of the notebooks in the docs actually does not contain a rendered version of the widgets - the outputs appear empty.

Nevertheless, this PR contains some fixes for the code as well that should be merged.

yakutovicha (approved):
LGTM! Thanks @ltalirz!

PR #238, opened 2021-09-15
Fix links to the notebooks.

yakutovicha:
Notebooks were recently placed in the 'notebooks' folder, so many
links became broken.

### DIFF
--- a/aiidalab_widgets_base/codes.py
+++ b/aiidalab_widgets_base/codes.py
@@ -69,7 +69,9 @@ class CodeDropdown(ipw.VBox):
             setup_code_params[key] = value
 
         # For later: use base_url here, when it will be implemented in the appmode.
-        url_string = f"<a href={path_to_root}aiidalab-widgets-base/setup_code.ipynb?"
+        url_string = (
+            f"<a href={path_to_root}aiidalab-widgets-base/notebooks/setup_code.ipynb?"
+        )
         url_string += "&".join(
             [f"{k}={v}".replace(" ", "%20") for k, v in setup_code_params.items()]
         )
--- a/aiidalab_widgets_base/computers.py
+++ b/aiidalab_widgets_base/computers.py
@@ -851,7 +851,7 @@ class ComputerDropdown(ipw.VBox):
         self.observe(self.refresh, names="allow_select_disabled")
 
         self._setup_another = ipw.HTML(
-            value=f"""<a href={path_to_root}aiidalab-widgets-base/setup_computer.ipynb target="_blank">
+            value=f"""<a href={path_to_root}aiidalab-widgets-base/notebooks/setup_computer.ipynb target="_blank">
             Setup new computer</a>"""
         )
 
--- a/aiidalab_widgets_base/elns.py
+++ b/aiidalab_widgets_base/elns.py
@@ -55,7 +55,7 @@ class ElnImportWidget(ipw.VBox):
         eln = connect_to_eln(**kwargs)
 
         if eln is None:
-            url = f"{path_to_root}aiidalab-widgets-base/eln_configure.ipynb"
+            url = f"{path_to_root}aiidalab-widgets-base/notebooks/eln_configure.ipynb"
             error_message.value = f"""Warning! The access to ELN {kwargs['eln_instance']} is not configured. Please follow <a href="{url}" target="_blank">the link</a> to configure it."""
             return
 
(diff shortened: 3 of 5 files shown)


ltalirz (approved):
thanks @yakutovicha 

Since the "notebook root directory" seems to be used in so many places, does it perhaps make sense to define a variable for it and reuse that?

csadorf (commented):
It looks fine to me, do you want me to test it?

yakutovicha:
> It looks fine to me, do you want me to test it?

It would be good if you can quickly check this. Thanks!

yakutovicha:
> thanks @yakutovicha
> 
> Since the "notebook root directory" seems to be used in so many places, does it perhaps make sense to define a variable for it and reuse that?
> I only fixed the links in the start.py

I think what we need is a variable that defines all up to `/apps/apps/home/start.ipynb`. For instance, for the local aiidalab, this variable should be something like `http://localhost:8888` for the server instance, it should be something like: `https://aiidalab.materialscloud.org/user/user.name@mail.com`.

csadorf:
> > thanks @yakutovicha
> > Since the "notebook root directory" seems to be used in so many places, does it perhaps make sense to define a variable for it and reuse that?
> > I only fixed the links in the start.py
> 
> I think what we need is a variable that defines all up to `/apps/apps/home/start.ipynb`. For instance, for the local aiidalab, this variable should be something like `http://localhost:8888` for the server instance, it should be something like: `https://aiidalab.materialscloud.org/user/user.name@mail.com`.

Why do you need the host information? We were able to link with relative links so far?

yakutovicha:
> Why do you need the host information? We were able to link with relative links so far?

Because that variable can be used unchanged for all the links. The current approach is to use the `path_to_root` variable that tells how to go from the current notebook to `<server>/apps/apps/`. Most often it is just equal to `..` because the notebooks is placed in the root folder of an app. But if one decides to place the notebook elsewhere, one needs to redefine this `path_to_root` variable.

csadorf:
> > Why do you need the host information? We were able to link with relative links so far?
> 
> Because that variable can be used unchanged for all the links. The current approach is to use the `path_to_root` variable that tells how to go from the current notebook to `<server>/apps/apps/`. Most often it is just equal to `..` because the notebooks are typically placed in the root folder of an app. But if one decides to place the notebook elsewhere, one needs to redefine this `path_to_root` variable.

So the reason that I am asking is because it is often non-trivial to determine this, especially in an automated way, and if we could avoid it would be better. For example, a server might be reachable from http://localhost:8888, http://127.0.0.1:8888, http://lvh.me:8888, but maybe also via https and so on. Maybe it is the right thing to do and many web applications require you to provide the hostname, but so far we got away without needing to specify it and that's actually kind of nice.

ltalirz:
I agree that avoiding absolute URLs is useful if you can.

Couldn't the AiiDAlab package provide a function that determines `path_to_root` automatically from the local file system?
I guess we know in which directory the jupyter notebook server is running, so one just needs to figure out where the current notebook is relative to that

yakutovicha:
> Couldn't the `aiidalab` package provide a little helper function that determines `path_to_root` automatically from the local file system?

I think it can be done fairly easy. Let me have a look.

yakutovicha:
> > Couldn't the `aiidalab` package provide a little helper function that determines `path_to_root` automatically from the local file system?
> 
> I think it can be done fairly easy. Let me have a look.

something like this?
```python
import os
from aiidalab.config import AIIDALAB_APPS
cwd = os.getcwd()
path_to_root = os.path.relpath(AIIDALAB_APPS, cwd)
```

ltalirz:
Apparently it's somewhat difficult to determine [in general](https://stackoverflow.com/a/52119628/1069467).
I think you approach will already work in most cases in the AiiDAlab - unless a notebook changes the working directory explicitly.

One way to make it robust against this is to save the working directory in the beginning.
It seems that this is actually stored in the `_dh` variable.

I'd suggest the logic:
 1. try to get it from `_dh`
 2. since this is an underscore variable that's not officially part of the API, if the variable does not exist, fall back to os.getcwd

See also discussion on https://github.com/ipython/ipython/issues/10123

yakutovicha:
> See also discussion on [ipython/ipython#10123](https://github.com/ipython/ipython/issues/10123)

After reading the thread I feel that we should not try to automate this.

PR #244, opened 2021-09-22
Fix computers management logic and links.

yakutovicha:


aiidalab_widgets_base/computers.py
```diff
@@ -369,13 +374,17 @@ def _configure_proxy(self, password, proxy_password):
...
+                if not success:
+                    return False, ""
 
             # Finally trying to connect.
             if self._send_pubkey(self.proxy_hostname, proxy_username, proxy_password):
                 return True, proxy_username + "@" + self.proxy_hostname
 
-            print(f"Could not send public key to {self.proxy_hostname} (proxy server).")
+            print(
+                f"Could not send public key to {self.proxy_hostname} (proxy server), sorry :-("
```
csadorf, line 386: I'm not really a fan of these kind of "cute" messages, but that's maybe just me... 😛

csadorf (approved):
Looks fine to me, did not test it.

aiidalab_widgets_base/computers.py
yakutovicha: ¯\\\_(ツ)_/¯

PR #246, opened 2021-09-30
Structure Viewer feature.

### LINKED ISSUE #241: add structure PK to the name of the download file
it is very useful to have automatically the PK of the structure in the name of the downloaded structure

https://github.com/aiidalab/aiidalab-widgets-base/blob/3c509a5fbfc3e6efbd1f05b650995349174c05ed/aiidalab_widgets_base/viewers.py#L551

yakutovicha:
fixes #241.

Add pk number to the downloaded structure if the structure is a stored AiiDA node.

### DIFF
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -546,10 +546,9 @@ class _StructureDataBaseViewer(ipw.VBox):
 
     def download(self, change=None):  # pylint: disable=unused-argument
         """Prepare a structure for downloading."""
-        pk = f"pk-{self.pk}" if self.pk else "not-stored"
+        suffix = f"-pk-{self.pk}" if self.pk else ""
         self._download(
             payload=self._prepare_payload(),
-            suffix = f"-pk-{self.pk}" if self.pk else ""
             filename=f"structure{suffix}.{self.file_format.value}",
         )
 
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -546,9 +546,10 @@ class _StructureDataBaseViewer(ipw.VBox):
 
     def download(self, change=None):  # pylint: disable=unused-argument
         """Prepare a structure for downloading."""
+        pk = f"pk-{self.pk}" if self.pk else "not-stored"
         self._download(
             payload=self._prepare_payload(),
-            filename="structure." + self.file_format.value,
+            filename=f"structure-{pk}.{self.file_format.value}",
         )
 
     @staticmethod
@@ -602,6 +603,7 @@ class StructureDataViewer(_StructureDataBaseViewer):
 
     structure = Union([Instance(Atoms), Instance(Node)], allow_none=True)
     displayed_structure = Instance(Atoms, allow_none=True, read_only=True)
+    pk = Int(allow_none=True)
 
     def __init__(self, structure=None, **kwargs):
         super().__init__(**kwargs)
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/viewers.py
```diff
@@ -546,9 +546,10 @@ def apply_selection(self, _=None):
 
     def download(self, change=None):  # pylint: disable=unused-argument
         """Prepare a structure for downloading."""
+        pk = f"pk-{self.pk}" if self.pk else "not-stored"
```
csadorf, line 549: I've adjusted this in the comment below. Further, this should probably only be executed when needed.
```suggestion
```

aiidalab_widgets_base/viewers.py
```diff
@@ -546,9 +546,10 @@ def apply_selection(self, _=None):
...
         self._download(
             payload=self._prepare_payload(),
-            filename="structure." + self.file_format.value,
+            filename=f"structure-{pk}.{self.file_format.value}",
```
csadorf, line 552: I don't think the "not-stored" suffix is needed:
```suggestion
            suffix = f"-{self.pk}" if self.pk else ""
            filename=f"structure{suffix}.{self.file_format.value}",
```
More importantly, I do not think we need the `self.pk` trait either:
```suggestion
            pk = getattr(self.structure, "pk", None)
            suffix = f"-{pk}" if pk is not None else ""
            filename=f"structure{suffix}.{self.file_format.value}",
```

aiidalab_widgets_base/viewers.py
```diff
@@ -602,6 +603,7 @@ class StructureDataViewer(_StructureDataBaseViewer):
 
     structure = Union([Instance(Atoms), Instance(Node)], allow_none=True)
     displayed_structure = Instance(Atoms, allow_none=True, read_only=True)
+    pk = Int(allow_none=True)
```
csadorf, line 606: I don't think we need this trait (see comment above).

csadorf (changes requested):

aiidalab_widgets_base/viewers.py
yakutovicha: Unfortunately, we do. See line 619 (validating structure). The structure always becomes an ASE object that does not contain the PK number.

aiidalab_widgets_base/viewers.py
yakutovicha: This wouldn't work, as `self.structure` is always an ASE object.

aiidalab_widgets_base/viewers.py
csadorf: Hence why I'm checking whether the structure trait has the pk attribute in the other code segment.

csadorf: Got it.

aiidalab_widgets_base/viewers.py
csadorf: Got it. Can you accept the first suggestion though?

yakutovicha: > Got it. Can you accept the first suggestion though?

Yes, but can you add back "pk" to the file name? This was a suggestion from @cpignedoli.

As far as I am aware, the term PK is widely used, so people are aware of what it means.

We should only make sure that people understand that the number in the file name is indeed the PK of the structure.

csadorf: Ah, yes, of course:
```suggestion
            suffix = f"-pk-{self.pk}" if self.pk else ""
            filename=f"structure{suffix}.{self.file_format.value}",
```

csadorf (approved):

PR #247, opened 2021-10-01
Add header to OptimadeQueryWidget.

### LINKED ISSUE #242: The "Select provider choice" of the OPTIMADE structure selector should be more visible
The "Select provider choice" can currently be easily overlooked (see screenshot). This could be easily remedied by adding an additional header right above the "Select provider" dropdown widget and renaming the dropdown hint to something else.

optimade-widget-select-provider

Proposed solution described here: https://github.com/CasperWA/voila-optimade-client/issues/351#issuecomment-923796290

csadorf:
Fixes #242.


yakutovicha (approved):
I would say we can merge this.

PR #255, opened 2021-10-06
Fix code setup.

### LINKED ISSUE #254: Setup a new code from DB should also create the computer automatically?
**Describe the bug**
When I pull from DB a code on a computer that is not available, the app crashes.

**To Reproduce**
1. Go to the setup a new code page
2. Pull the DB
3. Select a code on a computer you don't have
4. Click to setup the code.
5. See the error

**Expected behavior**
Either the button to setup the code is disabled and it's obvious which steps need to be done first, or (better?) also the computer is setup? (If a computer exists already, I don't know if a simple name match is enough though - maybe ask the user if one wants to reuse the same code, or setup another one e.g. "Eiger (2)" in case some settings are incorrect?

**Screenshots**
Screenshot 2021-10-04 at 19 25 04

yakutovicha:
fixes #254 

1. Make sure to not setup code if computer is not provided.
2. Gracefully handle situations when the requested code plugin
is not installed.

### DIFF
--- a/aiidalab_widgets_base/computers.py
+++ b/aiidalab_widgets_base/computers.py
@@ -908,8 +908,8 @@ class ComputerDropdown(ipw.VBox):
         if isinstance(computer, str):
             if computer in self.computers:
                 return self.computers[computer]
-            self.output.value = f"""No computer named '<span style="color:red">{computer}</span>'
-            was found in your AiiDA database."""
+            self.output.value = f"""Computer instance '<span style="color:red">{computer}</span>'
+            is not configured in your AiiDA profile."""
             return None
 
         if isinstance(computer, Computer):


notebooks/setup_code.ipynb
```diff
@@ -19,6 +19,27 @@
     "from aiidalab_widgets_base import AiiDACodeSetup, CodeDatabaseWidget"
    ]
   },
+  {
```
csadorf, line 22: Is there an actual change in this notebook or is this just random Jupyter shuffling?

aiidalab_widgets_base/computers.py
```diff
@@ -914,6 +914,6 @@ def _validate_selected_computer(self, change):
         if isinstance(computer, Computer):
             if computer.label in self.computers:
                 return computer
-            self.output.value = f"""The computer instance '<span style="color:red">{computer}</span>'
+            self.output.value = f"""The computer instance '<span style="color:red">{computer.label}</span>'
             supplied was not found in your AiiDA database."""
```
csadorf, line 918:
```suggestion
            self.output.value = f"""Computer instance '<span style="color:red">{computer.label}</span>'
            is not configured in your AiiDA profile."""
```

aiidalab_widgets_base/codes.py
```diff
@@ -257,6 +265,9 @@ def _setup_code(self, _=None):
             if not self.remote_abs_path:
                 print("You did not specify absolute path to the executable.")
                 return
+            if self.computer:
+                print("Please specify a computer from your database.")
```
csadorf, line 269:
```suggestion
                print("Please specify a computer that is configured in your AiiDA profile.")
```

aiidalab_widgets_base/codes.py
```diff
@@ -247,6 +247,14 @@ def __init__(self, path_to_root="../", **kwargs):
         ]
         super(AiiDACodeSetup, self).__init__(children, **kwargs)
 
+    @validate("input_plugin")
+    def _validate_input_plugin(self, proposal):
+        plugin = proposal["value"]
+        if plugin in self.inp_code_plugin.options:
+            return plugin
+        else:
+            return None
```
csadorf, line 256:
```suggestion
        return plugin if plugin in self.inp_code_plugin.options else None
```

csadorf (changes requested):
The computer part worked fine, but when I tried to setup a code for a non-installed plugin (I used eiger and used root to uninstall the aiida-quantumespresso plugin), nothing happened. No warning or error message or anything (see screenshot).
Screen Shot 2021-10-07 at 09 33 21

yakutovicha:
@csadorf, I tested things locally. Everything should work now. But please test it as well.

aiidalab_widgets_base/computers.py
```diff
@@ -910,10 +910,11 @@ def _validate_selected_computer(self, change):
                 return self.computers[computer]
             self.output.value = f"""No computer named '<span style="color:red">{computer}</span>'
```
csadorf, line 911: Make sure to adapt the error message to the slightly adjusted on lines 918+919.

csadorf (changes requested):
After uninstalling `aiida-quantumespresso` the "Code plugin" selection is now empty, however when clicking on "Setup code" nothing happens, just like before. I would suggest to disable the "Submit" button in case that the form is not correctly filled in.

As an aside, I did not refresh the page after uninstalling the package and the code was actually configured with the now uninstalled plugin.

csadorf:
@yakutovicha Should I test this again?

PR #257, opened 2021-10-13
Gracefully handling ELN-related issues.

yakutovicha:


aiidalab_widgets_base/elns.py
```diff
@@ -15,30 +15,48 @@
...
+        except NotImplementedError as err:
+            return None, str(err)
         eln.connect()
-        return eln
+        return eln, None
 
-    return None
+    return (
+        None,
+        "No ELN instance was provided, the default ELN instane is not configured either.",
```
kjappelbaum, line 58:
```suggestion
        "No ELN instance was provided, the default ELN instance is not configured either. Set a default ELN or select an ELN instance.",
```

aiidalab_widgets_base/elns.py
```diff
@@ -261,7 +279,14 @@ def check_connection(self, _=None):
 
     def display_eln_config(self, value=None):
         """Display ELN configuration specific to the selected type of ELN."""
-        eln_class = get_eln_connector(self.eln_types.value)
+        try:
+            eln_class = get_eln_connector(self.eln_types.value)
+        except NotImplementedError as err:
+            with self._output:
+                clear_output()
+                display(ipw.HTML("&#10060;" + str(err)))
```
kjappelbaum, line 287: Do we want to show a more user-friendly text also in this case? Something like "Unexpected error. There is no connector implemented for the ELN type you choose. Report his bug at ...."

aiidalab_widgets_base/elns.py
```diff
@@ -15,30 +15,48 @@
 
 
 def connect_to_eln(eln_instance=None, **kwargs):
+
+    # Assumming that the connection can only be established to the ELNs
+    # with the stored configuration.
     try:
         with open(ELN_CONFIG, "r") as file:
             config = json.load(file)
     except (FileNotFoundError, json.JSONDecodeError, KeyError):
```
kjappelbaum, line 24: Should the `write_to_config` have similar handling?

kjappelbaum: but the only way this would fail is if it, for some reason is not able to create and/or write this new file

aiidalab_widgets_base/elns.py
yakutovicha: > Might be actually a candidate for https://github.com/aiidalab/aiidalab-widgets-base/blob/develop/aiidalab_widgets_base/bug_report.py?

I would say this is mostly for the simulation apps where there can be quite unexpected errors. Here the possible errors are quite limited, so I would prefer to handle them as explicitly as possible.

aiidalab_widgets_base/elns.py
kjappelbaum: ok, good point. Then looks fine to me. I do not know what will happen if this directory does not exists (i think there is now no way I can imagine this to happen ...)

aiidalab_widgets_base/elns.py
```diff
@@ -15,30 +15,48 @@
 
 
 def connect_to_eln(eln_instance=None, **kwargs):
+
+    # Assumming that the connection can only be established to the ELNs
```
kjappelbaum, line 19:
```suggestion
    # assuming that the connection can only be established to the ELNs
```

kjappelbaum (dismissed):
nice, thanks!

PR #266, opened 2021-11-18
Reduce the maximum traceback size for the bug report feature.

csadorf:
The previous value could still lead to an excess of the maximum payload
size accepted by GitHub.


mbercx (approved):
Gave it a test run, and seems to fix the issue! So LGTM 🚀

PR #267, opened 2021-11-18
Update the list of disabled OPTIMADE providers.

csadorf:


CasperWA (approved):
This matches the lists in `CasperWA/voila-optimade-client` and `materialscloud-org/tools-optimade-client`.

PR #268, opened 2021-11-18
Computrational resources management.

yakutovicha:


aiidalab_widgets_base/databases.py
```diff
@@ -304,50 +305,47 @@ def _observe_proxy_command(self, _=None):
...
+    computer_config = traitlets.Dict()
+    computer_setup = traitlets.Dict()
+    code_config = traitlets.Dict()
 
     def __init__(self, **kwargs):
-        self.database = {}
 
+        self.database = requests.get(
+            "https://aiidateam.github.io/aiida-code-registry/database.json"
+        ).json()
```
csadorf, line 320: 1. Can we store the default value as a module constant and otherwise make this a class constructor parameter?
2. We should never perform blocking requests as part of the widget construction. This will block any application that uses this widget during startup and also cause a complete failure to load in case that the resource is not available. This should be executed non-blocking and also with an explicit well-behaving failure mode.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -1,43 +1,155 @@
...
+from aiida import plugins
+from IPython.display import clear_output, display
+
+from .databases import ComputationalResourcesDatabase
 
 STYLE = {"description_width": "200px"}
 
+class ComputationalResources(ipw.VBox):
+    """Code selection widget.
+    Attributes:
```
csadorf, line 13:
```suggestion
    """Code selection widget.
    
    Attributes:
```

csadorf (changes requested):
I have not tested this yet and this is also not a complete review, but I do have one change request that will need to be addressed.

csadorf:
@yakutovicha How do you want to proceed on this? Should I do a test and more thorough review or do you want to continue working on this first?

yakutovicha:
> @yakutovicha How do you want to proceed on this? Should I do a test and more thorough review or do you want to continue working on this first?

@csadorf I need to put some work into it before requesting your input. Didn't have time to continue the development last week.

csadorf:
@yakutovicha Please request my review when I should have a look at this. Thanks.

yakutovicha:
> @yakutovicha Please request my review when I should have a look at this. Thanks.

@csadorf please give it a try before reviewing. I wanted to make sure things work as wanted before diving into the code details.

yakutovicha:
Superseded by #279

Issue #262, opened 2021-11-02
NodesTreeWidget: find outputs not working for nested namespace


unkcpz:
I have the following outputs with some output nodes collected in the namespace:

```
Outputs                         PK    Type
------------------------------  ----  ------
convergence_cohesive_energy
    output_parameters           8916  Dict
delta_factor
    output_birch_murnaghan_fit  8937  Dict
    output_delta_factor         8943  Dict
    output_volume_energy        8935  Dict
pseudo_info                     7795  Dict
```

It will cause an issue when I want to represent the output as a NodesTreeWidget.
The exceptions is `AttributeError: 'AttributeDict' object has no attribute 'pk'`. 
It is because the `_find_outputs` class method did not detect when the output is a namespace the type `AttributeDict`.

danielhollas:
Oh brilliant, I have just hit this bug, glad that it is already reported. :blush:

unkcpz:
The fix only works for one layer of nest namespace hierarchy of outputs. If the namespace have more namespace it is still broken the work chain node view

PR #269, opened 2021-11-22
Node tree namespace

unkcpz:
fixes https://github.com/aiidalab/aiidalab-widgets-base/issues/262

Here I check the namespace in the output and make a `TreeNode` for it. I modified `AiidaOutputsTreeNode` by adding an attribute `self.namespace` to store the current namespace level.
Since the namespace and output have the same structure, that is to say, the output can be regarded as a special namespace in the top hierarchy without the property `namespace` being set (`nampspace=None`).

However, the namespace can also be nested where I didn't consider this but will be easy to add along with this PR. I suggest we keep one level nest namespace as this PR for the moment and get it merged, elaborate to a more comprehensive situation when we encounter the issue.

### DIFF
--- a/aiidalab_widgets_base/nodes.py
+++ b/aiidalab_widgets_base/nodes.py
@@ -107,17 +107,6 @@ class AiidaOutputsTreeNode(TreeNode):
         super().__init__(name=name, **kwargs)
 
 
-class AiidaNamespaceTreeNode(TreeNode):
-    icon = traitlets.Unicode("folder").tag(sync=True)
-    disabled = traitlets.Bool(True).tag(sync=True)
-
-    def __init__(self, name, parent_pk, namespace=None, **kwargs):
-        self.parent_pk = parent_pk
-        self.nodes_registry = dict()
-        self.namespace = namespace
-        super().__init__(name=name, **kwargs)
-
-
 class UnknownTypeTreeNode(AiidaNodeTreeNode):
     icon = traitlets.Unicode("file").tag(sync=True)
 
--- a/aiidalab_widgets_base/nodes.py
+++ b/aiidalab_widgets_base/nodes.py
@@ -3,6 +3,7 @@ import ipywidgets as ipw
 import traitlets
 from aiida.cmdline.utils.ascii_vis import calc_info
 from aiida.engine import ProcessState
+from aiida.common import AttributeDict
 from aiida.orm import (
     CalcFunctionNode,
     CalcJobNode,
@@ -102,6 +103,16 @@ class AiidaOutputsTreeNode(TreeNode):
     def __init__(self, name, parent_pk, **kwargs):
         self.parent_pk = parent_pk
         self.nodes_registry = dict()
+        super().__init__(name=name, **kwargs)
+        
+class AiidaNamespaceTreeNode(TreeNode):
+    icon = traitlets.Unicode("folder").tag(sync=True)
+    disabled = traitlets.Bool(True).tag(sync=True)
+
+    def __init__(self, name, parent_pk, namespace=None, **kwargs):
+        self.parent_pk = parent_pk
+        self.nodes_registry = dict()
+        self.namespace = namespace
         super().__init__(name=name, **kwargs)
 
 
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/nodes.py
```diff
@@ -207,16 +220,34 @@ def _find_called(cls, root):
...
+
+        output_attributs = {
+            key: outputs[key]
+            for key in outputs
+            if isinstance(outputs[key], AttributeDict)
+        }
+        for key in output_attributs.keys():
+            yield AiidaOutputsTreeNode(
+                name=key, parent_pk=root.parent_pk, namespace=key
+            )
```
csadorf, line 250: Do we need to make the distinction between `output_nodes` and `output_attributes` outside of the main loop on line 234? Maybe we just differentiate the yield based on the type directly in the loop?

unkcpz: Yes, I can make it outside the main loop, but then there is no way to sort the output_nodes by pks.

csadorf: You could sort on something like`key=lambda k: getattr(outputs[k], "pk", None)`?

unkcpz: Ha, true. Didn't think about that, I will make a try.

unkcpz:
Hi @csadorf sorry for the delay of this PR. I put the cases distinguish inside the loop and works as expected. 
Docs test failed, not sure what's going on with it.

aiidalab_widgets_base/nodes.py
```diff
@@ -207,16 +220,29 @@ def _find_called(cls, root):
...
+
+        for key in sorted(
+            output_nodes.keys(), key=lambda k: getattr(outputs[k], "pk", 0)
+        ):
+            node = output_nodes[key]
+
+            if isinstance(node, AttributeDict):
+                yield AiidaOutputsTreeNode(
+                    name=key, parent_pk=root.parent_pk, namespace=key
                 )
```
csadorf, line 239: This looks mostly good. What is the reason for not performing this conversion as part `._to_tree_node()`?

csadorf (commented):
Just a few more comments, but this is almost done.

aiidalab_widgets_base/nodes.py
unkcpz: Because after change `AiidaOutputsTreeNode` it no longer require have TreeNode object with `pk` as input, this is the reason cause this issue at first place. However  `._to_tree_node()` still try to use `pk` to initialize the TreeNode.

aiidalab_widgets_base/nodes.py
```diff
@@ -207,16 +220,29 @@ def _find_called(cls, root):
...
-                    output_node, name=f"{key}<{output_node.pk}>"
+        if root.namespace:
+            outputs = process_node.outputs[root.namespace]
+        else:
+            outputs = process_node.outputs
+
+        output_nodes = {key: outputs[key] for key in outputs}
+
+        for key in sorted(
+            output_nodes.keys(), key=lambda k: getattr(outputs[k], "pk", 0)
```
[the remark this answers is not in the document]
unkcpz: Changed, Thanks! @csadorf

aiidalab_widgets_base/nodes.py
csadorf: Sure, but that logic (type check or check for the `.pk` attribute) could have also been done inside of `._to_tree_node()`. Either way, I think this is fine and refactoring the sequence would not improve readability significantly or perhaps at all.

csadorf (approved):
Looks good!

I believe that the docs build failures might be related to a recent setuptools release which might have broken pip (https://github.com/pypa/setuptools/issues/3002). Either way, it would be clearly unrelated to this PR.

aiidalab_widgets_base/nodes.py
unkcpz: I guess it is possible to add an extra if-else in `to_tree_node` for nodes that are node the process type. But I think the current code is clear.

unkcpz:
@csadorf Thanks!

PR #281, opened 2022-01-17
Working version of trajectory visualizer based on bqplot

yakutovicha:


danielhollas:
@yakutovicha could you maybe explain what was the original motivation for this? I have actually based my TrajectoryData visualizer on this, but I have discarded the plotting. I think that makes sense, the base visualizer should probably be simple so that more complicated widgets can be built on top of it.

PR #282, opened 2022-01-17
Selection (#2)

cpignedoli:
* modified final selection for add

* modified final selection for modify atom


yakutovicha (approved):
Tested, seems to work fine. Thanks @cpignedoli

PR #285, opened 2022-01-23
Make StructureDataViewer more configurable

### LINKED ISSUE #283: StructureDataViewer: configure_view=False is broken
The following code does not work:

```pyhon
from ase.build import molecule
m = molecule('H2O')
m.center(vacuum=2.0)
CifData = DataFactory('cif')
s = CifData(ase=m)
vwr = viewer(s.store(), downloadable=True, configure_view=False)
display(vwr)
```
produces error:
```python
~/apps/aiidalab-widgets-base/aiidalab_widgets_base/viewers.py in _observe_selection(self, _)
    618     def _observe_selection(self, _=None):
    619         self.highlight_atoms(self.selection)
--> 620         self._selected_atoms.value = list_to_string_range(self.selection, shift=1)
    621 
    622     def apply_selection(self, _=None):

AttributeError: 'StructureDataViewer' object has no attribute '_selected_atoms'
```

I'll put up a PR with a proposed fix.

### LINKED ISSUE #284: Make StructureDataViewer more configurabe
I quite like the current StructureDataViewer widget, but I would like to simplify it's UI a it to make it more accessible to non-expert users. Also, I don't need some functionality for my use case.

I would like to be able to:

- Select which configuration tabs are displayed (e.g. the "Cell" tab is superfluous for gas phase molecules)
- hide the perspective/ortographic camera switcher

Because all this is done in the `_StructureDataBaseViewer:__init__`, I cannot easily override the behaviour. I'll make a PR with a proposed change if this seems like a reasonable request to y'all.

danielhollas:
- Fix configure_view=False option, only register observers
   when tabs are shown.
 - select which configuration tabs to display.
 - camera_default option
 - option to hide camera selection

Closes #284. Fixes #283. See these issues for more details/reasoning.

### DIFF
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -157,8 +157,14 @@ class DictViewer(ipw.VBox):
 class _StructureDataBaseViewer(ipw.VBox):
     """Base viewer class for AiiDA structure or trajectory objects.
 
-    :param configure_view: If True, add configuration tabs
-    :type configure_view: bool"""
+    :param configure_view: If True, add configuration tabs (deprecated)
+    :type configure_view: bool
+    :param configuration_tabs: List of configuration tabs (default: ["Selection", "Appearance", "Cell", "Download"])
+    :type configure_view: list
+    :param default_camera: default camera (orthographic|perspective), can be changed in the Appearance tab
+    :type default_camera: string
+
+    """
 
     selection = List(Int)
     selection_adv = Unicode()
@@ -168,44 +174,48 @@ class _StructureDataBaseViewer(ipw.VBox):
     DEFAULT_SELECTION_RADIUS = 6
     DEFAULT_SELECTION_COLOR = "green"
 
-    def __init__(self, configure_view=True, **kwargs):
+    def __init__(
+        self,
+        configure_view=True,
+        configuration_tabs=["Selection", "Appearance", "Cell", "Download"],
+        default_camera="orthographic",
+        **kwargs,
+    ):
         # Defining viewer box.
 
-        # 1. Nglviwer
+        # Nglviwer
         self._viewer = nglview.NGLWidget()
-        self._viewer.camera = "orthographic"
+        self._viewer.camera = default_camera
         self._viewer.observe(self._on_atom_click, names="picked")
         self._viewer.stage.set_parameters(mouse_preset="pymol")
 
-        # 2. Camera type.
-        camera_type = ipw.ToggleButtons(
-            options={"Orthographic": "orthographic", "Perspective": "perspective"},
-            description="Camera type:",
-            value="orthographic",
-            layout={"align_self": "flex-end"},
-            style={"button_width": "115.5px"},
-            orientation="vertical",
-        )
+        view_box = ipw.VBox([self._viewer])
 
-        def change_camera(change):
+        configuration_tabs_map = {
+            "Selection": self._selection_tab(),
+            "Appearance": self._appearance_tab(),
+            "Cell": self._cell_tab(),
+            "Download": self._download_tab(),
+        }
 
-            self._viewer.camera = change["new"]
-
-        camera_type.observe(change_camera, names="value")
-        view_box = ipw.VBox([self._viewer, camera_type])
+        if configure_view is not True:
+            warnings.warn(
+                "`configure_view` is deprecated, please use `configuration_tabs` instead.",
+                DeprecationWarning,
+            )
+            if not configure_view:
+                configuration_tabs.clear()
 
         # Constructing configuration box
-        if configure_view:
+        if len(configuration_tabs) != 0:
             configuration_box = ipw.Tab(
                 layout=ipw.Layout(flex="1 1 auto", width="auto")
             )
             configuration_box.children = [
-                self._selection_tab(),
-                self._appearance_tab(),
-                self._cell_tab(),
-                self._download_tab(),
+                configuration_tabs_map[tab_title] for tab_title in configuration_tabs
             ]
-            for i, title in enumerate(["Selection", "Appearance", "Cell", "Download"]):
+
+            for i, title in enumerate(configuration_tabs):
                 configuration_box.set_title(i, title)
             children = [ipw.HBox([view_box, configuration_box])]
             view_box.layout = {"width": "60%"}
@@ -299,11 +309,29 @@ class _StructureDataBaseViewer(ipw.VBox):
         link((background_color, "value"), (self._viewer, "background"))
         background_color.value = "white"
 
-        # 3. Center button.
-        center_button = ipw.Button(description="Center")
+        # 3. Camera switcher
+        camera_type = ipw.ToggleButtons(
+            options={"Orthographic": "orthographic", "Perspective": "perspective"},
+            description="Camera type:",
+            value=self._viewer.camera,
+            layout={"align_self": "flex-start"},
+            style={"button_width": "115.5px"},
+            orientation="vertical",
+        )
+
+        def change_camera(change):
+
+            self._viewer.camera = change["new"]
+
+        camera_type.observe(change_camera, names="value")
+
+        # 4. Center button.
+        center_button = ipw.Button(description="Center molecule")
         center_button.on_click(lambda c: self._viewer.center())
 
-        return ipw.VBox([supercell_selector, background_color, center_button])
+        return ipw.VBox(
+            [supercell_selector, background_color, camera_type, center_button]
+        )
 
     @observe("cell")
     def _observe_cell(self, _=None):
--- a/notebooks/aiida_datatypes_viewers.ipynb
+++ b/notebooks/aiida_datatypes_viewers.ipynb
@@ -69,7 +69,7 @@
    "source": [
     "CifData = DataFactory('cif')\n",
     "s = CifData(ase=m)\n",
-    "vwr = viewer(s.store(), downloadable=True)\n",
+    "vwr = viewer(s.store(), configuration_tabs=['Selection', 'Appearance', 'Cell', 'Download'])\n",
     "display(vwr)"
    ]
   },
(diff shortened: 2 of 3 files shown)


aiidalab_widgets_base/viewers.py
```diff
@@ -993,7 +1014,8 @@ def _observe_selection_adv(self, _=None):
 
     @observe("selection")
     def _observe_selection_2(self, _=None):
-        self.selection_info.value = self.create_selection_info()
+        if hasattr(self, "selection_info"):
```
danielhollas, line 1017: I am not happy about this, but since this observer is in `StructureDataViewer` class, I cannot register it inside `_selection_tab`. Ideas welcomed. :blush:

aiidalab_widgets_base/viewers.py
```diff
@@ -305,7 +326,6 @@ def change_supercell(_=None):
 
         return ipw.VBox([supercell_selector, background_color, center_button])
 
-    @observe("cell")
```
danielhollas, line 308: Instead of using the decorator, observers are registered manually only when the corresponding tab is displayed.

aiidalab_widgets_base/viewers.py
```diff
@@ -168,44 +168,46 @@ class _StructureDataBaseViewer(ipw.VBox):
     DEFAULT_SELECTION_RADIUS = 6
     DEFAULT_SELECTION_COLOR = "green"
 
-    def __init__(self, configure_view=True, **kwargs):
+    def __init__(
+        self,
+        configure_view=True,
+        configuration_tabs=["Selection", "Appearance", "Cell", "Download"],
+        default_camera="orthographic",
+        show_camera_selector=True,
```
yakutovicha, line 176: @danielhollas I was wondering, would it make sense for your use case to move the camera setting into another tab (along with "Selection", "Appearance")? Then we could manage its appearance just like everything else.

aiidalab_widgets_base/viewers.py
```diff
@@ -217,6 +219,23 @@ def change_camera(change):
 
         super().__init__(children, **kwargs)
 
+    def _camera_selection(self, default_camera):
```
yakutovicha, line 222: If that works for you, I would propose to make it just another tab and call it `_camera_tab(self, ..)`

yakutovicha (commented):
Thanks, @danielhollas - just a few comments, suggestions.

aiidalab_widgets_base/viewers.py
danielhollas: That seems fine. Actually, wouldn't it make sense if the camera toggle was part of the appearance tab? It seems a bit wasteful to have an extra tab just for this.

yakutovicha: > Actually, wouldn't it make sense if the camera toggle was part of the appearance tab? 

fine with me.

aiidalab_widgets_base/viewers.py
```diff
@@ -168,44 +168,46 @@ class _StructureDataBaseViewer(ipw.VBox):
     DEFAULT_SELECTION_RADIUS = 6
     DEFAULT_SELECTION_COLOR = "green"
 
-    def __init__(self, configure_view=True, **kwargs):
+    def __init__(
+        self,
+        configure_view=True,
```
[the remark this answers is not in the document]
danielhollas: @yakutovicha I was thinking about that as well but did not want to break existing code. I am not sure I understood your suggestion correctly, but below I changed:

```python
if configure_view:
     ...
```
to 
```python
if configure_view and len(configuration_tabs) != 0:
     ....
```
This makes `configure_view` redundant but still working in existing widgets.

notebooks/aiida_datatypes_viewers.ipynb
```diff
@@ -88,7 +88,7 @@
    "source": [
     "StructureData = DataFactory('structure')\n",
     "s = StructureData(ase=m)\n",
-    "vwr = viewer(s.store(), downloadable=True)\n",
+    "vwr = viewer(s.store())\n",
```
danielhollas, line 91: "downloadable" parameter does not make any difference for StructureViewerWidget.

danielhollas:
@yakutovicha thank you for the review! I made some changes based on your suggestions, and I also tried to update the docstring accordingly.

aiidalab_widgets_base/viewers.py
yakutovicha: I don't think it is a good idea. `cell` is the `_StructureDataBaseViewer` trait. And, as I wrote in another comment, all the sub-widgets should always exist (despite they are not always shown).

aiidalab_widgets_base/viewers.py
yakutovicha: As I wrote in the other comments, `selection_info` should always exist. So there is, probably, no need to add an `if` statement here.

aiidalab_widgets_base/viewers.py
```diff
@@ -168,44 +174,40 @@ class _StructureDataBaseViewer(ipw.VBox):
...
-
-            self._viewer.camera = change["new"]
-
-        camera_type.observe(change_camera, names="value")
-        view_box = ipw.VBox([self._viewer, camera_type])
+        view_box = ipw.VBox([self._viewer])
 
         # Constructing configuration box
-        if configure_view:
+        if configure_view and len(configuration_tabs) != 0:
```
yakutovicha, line 195: Maybe a deprecation warning would be a good addition, something like this:

```suggestion
        if configure_view is not None:
            warnings.warn("`configure_view` is deprecated, please use `configuration_tabs` instead.", DeprecationWarning)
        if configure_view and len(configuration_tabs) != 0:
```

yakutovicha (changes requested):
Thanks, @danielhollas.

I did another pass - see my comments/suggestions below.

Once they are addressed, I think we can go ahead with merging.

aiidalab_widgets_base/viewers.py
danielhollas: I am fine with deprecation warning, but the if condition is tricky because right now `configure_view=True` by default. 
If we make it None by default, then the if condition `if configure_view and len(configuration_tabs)` does not work.

danielhollas: I guess keeping the default to True and just inverting the condition like this will work, though it is a bit confusing:
```python
if not configure_view:
    warnings.warn("`configure_view` is deprecated, please use `configuration_tabs` instead.", DeprecationWarning)
```

aiidalab_widgets_base/viewers.py
danielhollas: Okay, Done.

aiidalab_widgets_base/viewers.py
danielhollas: @csadorf thanks, I went with your idea.

aiidalab_widgets_base/viewers.py
```diff
@@ -168,44 +174,40 @@ class _StructureDataBaseViewer(ipw.VBox):
...
             configuration_box = ipw.Tab(
                 layout=ipw.Layout(flex="1 1 auto", width="auto")
             )
+
+            configuration_tabs_map = {
```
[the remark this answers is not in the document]
danielhollas: Okay, I moved the dictionary and reverted all changes to observers.

danielhollas:
@yakutovicha thanks! I pushed a new version which I believe addresses all your comments.

csadorf:
@danielhollas I believe @yakutovicha might not be available for a review right now. How urgent is this PR? If urgent, I could look into finishing the review process instead.

danielhollas:
No worries, not super urgent.

Dne po 31. 1. 2022 3:58 odp. uživatel Carl Simon Adorf <
***@***.***> napsal:

aiidalab_widgets_base/viewers.py
```diff
@@ -168,44 +174,48 @@ class _StructureDataBaseViewer(ipw.VBox):
...
+            if configure_view is False:
+                configuration_tabs.clear()
 
         # Constructing configuration box
-        if configure_view:
+        if len(configuration_tabs) != 0:
             configuration_box = ipw.Tab(
                 layout=ipw.Layout(flex="1 1 auto", width="auto")
             )
             configuration_box.children = [
```
yakutovicha, line 214: What should happen if the user provided a non-existing tab name? Should it fail as it does now, or should there be a warning message?

I am fine with any approach. Just mentioning it here, as it is worth considering.

yakutovicha (changes requested):
Thanks, @danielhollas. I tested it, and it works.

Please replace the following line: 
https://github.com/aiidalab/aiidalab-widgets-base/blob/c0aee35119707f88a2ece7726ce618825cd7c871/aiidalab_widgets_base/structures.py#L95
with 
```python
self.viewer = StructureDataViewer(**kwargs)
```
and address the comment below. Once done - this is good to go from my side.

aiidalab_widgets_base/viewers.py
danielhollas: @yakutovicha good question. In general I am often unsure about how to handle / propagate errors. In this case I'd say this would be a programming error and it is okay to crash.

aiidalab_widgets_base/viewers.py
```diff
@@ -168,44 +174,48 @@ class _StructureDataBaseViewer(ipw.VBox):
...
+        }
 
-        camera_type.observe(change_camera, names="value")
-        view_box = ipw.VBox([self._viewer, camera_type])
+        if configure_view is not True:
+            warnings.warn(
+                "`configure_view` is deprecated, please use `configuration_tabs` instead.",
+                DeprecationWarning,
+            )
+            if configure_view is False:
```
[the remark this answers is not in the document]
danielhollas: Yep, done.

danielhollas:
Thanks @yakutovicha . All done I believe.

aiidalab_widgets_base/viewers.py
yakutovicha: ok, fine with me.

yakutovicha (approved):
All good!
Thank you very much, @danielhollas!

PR #290, opened 2022-02-02
SmilesWidget: PBC=False + better error handling

### LINKED ISSUE #270: SmilesWidget: Handle invalid SMILES gracefully
For example, this input: `CC(C)(C)*OO` generates this stack trace.

```python
KeyError                                  Traceback (most recent call last)
~/apps/aiidalab-widgets-base/aiidalab_widgets_base/structures.py in _on_button_pressed(self, change)
    728             self.SPINNER
    729         )  # font-size:20em;
--> 730         self.structure = self._mol_from_smiles(self.smiles.value)
    731         self.output.value = ""
    732 

~/apps/aiidalab-widgets-base/aiidalab_widgets_base/structures.py in _mol_from_smiles(self, smiles, steps)
    715         """Convert SMILES to ase structure try rdkit then pybel"""
    716         try:
--> 717             return self._rdkit_opt(smiles, steps)
    718         except ValueError:
    719             return self._pybel_opt(smiles, steps)

~/apps/aiidalab-widgets-base/aiidalab_widgets_base/structures.py in _rdkit_opt(self, smiles, steps)
    710         natoms = mol.GetNumAtoms()
    711         species = [mol.GetAtomWithIdx(j).GetSymbol() for j in range(natoms)]
--> 712         return self._make_ase(species, positions, smiles)
    713 
    714     def _mol_from_smiles(self, smiles, steps=10000):

~/apps/aiidalab-widgets-base/aiidalab_widgets_base/structures.py in _make_ase(self, species, positions, smiles)
    663         # Get the principal axes and realign the molecule along z-axis.
    664         positions = PCA(n_components=3).fit_transform(positions)
--> 665         atoms = Atoms(species, positions=positions, pbc=True)
    666         atoms.cell = np.ptp(atoms.positions, axis=0) + 10
    667         atoms.center()

/opt/conda/lib/python3.8/site-packages/ase/atoms.py in __init__(self, symbols, positions, numbers, tags, momenta, masses, magmoms, charges, scaled_positions, cell, pbc, celldisp, constraint, calculator, info, velocities)
    206                     'Use only one of "symbols" and "numbers".')
    207             else:
--> 208                 self.new_array('numbers', symbols2numbers(symbols), int)
    209 
    210         if cell is None:

/opt/conda/lib/python3.8/site-packages/ase/symbols.py in symbols2numbers(symbols)
     18     for s in symbols:
     19         if isinstance(s, str):
---> 20             numbers.append(atomic_numbers[s])
     21         else:
     22             numbers.append(int(s))

KeyError: '*'
```

Happy to take this on, but if anybody want to fix this go ahead. :-)

### LINKED ISSUE #289: SmilesWidget: PBC should be set to False
For some reason, the structures generated with SmilesWidget have PBC set to true.
I don't think that SMILES can be used to define crystal structures, and even if it did I think PBC=False is a more common default.

danielhollas:
- Make PBC=False by default, closes #289
 - Improve handling of invalid SMILES input, closes #270.

Turns out that the RDkit library doesn't have a very good error handling, they just print all errors and warnings to stderr.
I tried to improve things on our side to capture obviously errorneous SMILES input and print a generic error message.
We could improve this further by capturing stderr and displaying the original RDKit messages as per http://rdkit.blogspot.com/2016/03/capturing-error-information.html, but not sure it's worth it the additional code complexity and messing with stderr.

**Some examples with invalid SMILES:**

1. 
input1: `c1cc1`
stderr: `RDKit ERROR: [19:58:16] Can't kekulize mol.  Unkekulized atoms: 0 1 2` 
input2: `Ch`
stderr: `RDKit ERROR: [20:01:52] SMILES Parse Error: syntax error while parsing: Ch
RDKit ERROR: [20:01:52] SMILES Parse Error: Failed parsing SMILES 'Ch' for input: 'Ch'`

Both of these would result in:
```
ArgumentError: Python argument types in
    rdkit.Chem.rdmolops.AddHs(NoneType)
did not match C++ signature:
    AddHs(RDKit::ROMol mol, bool explicitOnly=False, bool addCoords=False, boost::python::api::object onlyOnAtoms=None, bool addResidueInfo=False)
```
With this PR, we just print "ERROR: Invalid SMILES code"

---
2.
input: `CC(C)(C)*OO`
For some reason, this one passes through rdkit even though it contains an invalid character `*`, and fails later in ASE, as described in #270. In this PR, we print "Invalid atom type".

---
3. 
input: `CS(C)(C)=O`
This is actually a valid SMILES string, as discussed in https://github.com/rdkit/rdkit/issues/200. However, the UFF force field that we use for minimization is missing parameters for sulphur in this particular hybridization, so the rdkit minimization fails with
```
~/apps/aiidalab-widgets-base/aiidalab_widgets_base/structures.py in _rdkit_opt(self, smiles, steps)
    706 
    707         AllChem.EmbedMolecule(mol, maxAttempts=20, randomSeed=42)
--> 708         AllChem.UFFOptimizeMolecule(mol, maxIters=steps)
    709         positions = mol.GetConformer().GetPositions()
    710         natoms = mol.GetNumAtoms()

RuntimeError: Pre-condition Violation
	bad params pointer
	Violation occurred on line 75 in file Code/ForceField/UFF/AngleBend.cpp
	Failed Expression: at2Params
	RDKIT: 2020.09.1
	BOOST: 1_74
```
With the new code, we abandon rdkit if we detect missing UFF parameters, and instead try OpenBabel instead, which in this case actually produces the desired molecule.

@cpignedoli would you mind taking a look as the original author of the widget. I just want to make sure that I didn't break some use case that worked before. I would be also interested to hear a bit about the original design / testing of the current code, i.e. for which SMILES did you find OpenBabel working better than RDKit? Thanks!

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -669,8 +669,8 @@ class SmilesWidget(ipw.VBox):
             atoms = Atoms(species, positions=positions, pbc=False)
         except KeyError:
             # https://github.com/aiidalab/aiidalab-widgets-base/issues/270
-            self.output.value = "ERROR: Invalid atom name"
-            return None
+            msg = "ERROR: Invalid atom name"
+            raise ValueError(msg)
         atoms.cell = np.ptp(atoms.positions, axis=0) + 10
         atoms.center()
         # We're attaching this info so that it
@@ -687,7 +687,8 @@ class SmilesWidget(ipw.VBox):
         obconversion = ob.OBConversion()
         obconversion.SetInFormat("smi")
         obmol = ob.OBMol()
-        obconversion.ReadString(obmol, smiles)
+        if not obconversion.ReadString(obmol, smiles):
+            raise ValueError("OpenBabel: Invalid SMILES")
 
         pbmol = pb.Molecule(obmol)
         pbmol.make3D(forcefield="uff", steps=50)
@@ -696,9 +697,13 @@ class SmilesWidget(ipw.VBox):
         pbmol.localopt(forcefield="mmff94", steps=100)
 
         f_f = pb._forcefields["uff"]  # pylint: disable=protected-access
-        f_f.Setup(pbmol.OBMol)
+        if not f_f.Setup(pbmol.OBMol):
+            raise ValueError("OpenBabel: Missing UFF parameters")
+
         f_f.ConjugateGradients(steps, 1.0e-9)
-        f_f.GetCoordinates(pbmol.OBMol)
+        if not f_f.GetCoordinates(pbmol.OBMol):
+            raise ValueError("OpenBabel: Optimization failed")
+
         species = [chemical_symbols[atm.atomicnum] for atm in pbmol.atoms]
         positions = np.asarray([atm.coords for atm in pbmol.atoms])
         return self._make_ase(species, positions, smiles)
@@ -711,30 +716,47 @@ class SmilesWidget(ipw.VBox):
         smiles = smiles.replace("[", "").replace("]", "")
         mol = Chem.MolFromSmiles(smiles)
         if mol is None:
-            # Something is seriously wrong with the SMILES code,
-            # just return None and don't attempt anything else.
-            self.output.value = "ERROR: Invalid SMILES string"
+            self.output.value = (
+                "WARNING: RDKit failed to parse SMILES, trying OpenBabel."
+            )
             return None
         mol = Chem.AddHs(mol)
 
-        AllChem.EmbedMolecule(mol, maxAttempts=20, randomSeed=42)
+        params = AllChem.ETKDG()
+        params.randomSeed = 42
+        params.maxAttempts = 20
+        conf_id = AllChem.EmbedMolecule(mol, params=params)
+        if conf_id == -1:
+            # useRandomCoords might be more robust for larger molecules
+            # https://sourceforge.net/p/rdkit/mailman/message/21776083/
+            params.useRandomCoords = True
+            conf_id = AllChem.EmbedMolecule(mol, params=params)
+        if conf_id == -1:
+            msg = "WARNING: RDkit embedding failed, trying OpenBabel"
+            self.output.value = msg
+            return None
+
         if not AllChem.UFFHasAllMoleculeParams(mol):
-            msg = "WARNING: Missing UFF parameters"
-            raise ValueError(msg)
+            msg = "WARNING: RDKit is missing UFF parameters, trying OpenBabel"
+            self.output.value = msg
+            return None
 
         AllChem.UFFOptimizeMolecule(mol, maxIters=steps)
-        positions = mol.GetConformer().GetPositions()
+        positions = mol.GetConformer(id=conf_id).GetPositions()
         natoms = mol.GetNumAtoms()
         species = [mol.GetAtomWithIdx(j).GetSymbol() for j in range(natoms)]
         return self._make_ase(species, positions, smiles)
 
-    def _mol_from_smiles(self, smiles, steps=10000):
+    def _mol_from_smiles(self, smiles, steps=1000):
         """Convert SMILES to ase structure try rdkit then pybel"""
         try:
-            return self._rdkit_opt(smiles, steps)
+            struct = self._rdkit_opt(smiles, steps)
+            if struct is None:
+                struct = self._pybel_opt(smiles, steps)
+            return struct
         except ValueError as e:
             self.output.value = str(e)
-            return self._pybel_opt(smiles, steps)
+            return None
 
     def _on_button_pressed(self, change):  # pylint: disable=unused-argument
         """Convert SMILES to ase structure when button is pressed."""
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -764,9 +764,7 @@ class SmilesWidget(ipw.VBox):
 
         if not self.smiles.value:
             return
-        spiner = "Screening possible conformers {}".format(
-            self.SPINNER
-        )  # font-size:20em;
+        spiner = f"Screening possible conformers {self.SPINNER}"  # font-size:20em;
         self.output.value = spiner
         self.structure = self._mol_from_smiles(self.smiles.value)
         # Don't overwrite possible error/warning messages
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -665,7 +665,12 @@ class SmilesWidget(ipw.VBox):
         """Create ase Atoms object."""
         # Get the principal axes and realign the molecule along z-axis.
         positions = PCA(n_components=3).fit_transform(positions)
-        atoms = Atoms(species, positions=positions, pbc=True)
+        try:
+            atoms = Atoms(species, positions=positions, pbc=False)
+        except KeyError:
+            # https://github.com/aiidalab/aiidalab-widgets-base/issues/270
+            self.output.value = "ERROR: Invalid atom name"
+            return None
         atoms.cell = np.ptp(atoms.positions, axis=0) + 10
         atoms.center()
         # We're attaching this info so that it
@@ -705,9 +710,18 @@ class SmilesWidget(ipw.VBox):
 
         smiles = smiles.replace("[", "").replace("]", "")
         mol = Chem.MolFromSmiles(smiles)
+        if mol is None:
+            # Something is seriously wrong with the SMILES code,
+            # just return None and don't attempt anything else.
+            self.output.value = "ERROR: Invalid SMILES string"
+            return None
         mol = Chem.AddHs(mol)
 
         AllChem.EmbedMolecule(mol, maxAttempts=20, randomSeed=42)
+        if not AllChem.UFFHasAllMoleculeParams(mol):
+            msg = "WARNING: Missing UFF parameters"
+            raise ValueError(msg)
+
         AllChem.UFFOptimizeMolecule(mol, maxIters=steps)
         positions = mol.GetConformer().GetPositions()
         natoms = mol.GetNumAtoms()
@@ -718,7 +732,8 @@ class SmilesWidget(ipw.VBox):
         """Convert SMILES to ase structure try rdkit then pybel"""
         try:
             return self._rdkit_opt(smiles, steps)
-        except ValueError:
+        except ValueError as e:
+            self.output.value = str(e)
             return self._pybel_opt(smiles, steps)
 
     def _on_button_pressed(self, change):  # pylint: disable=unused-argument
@@ -727,11 +742,14 @@ class SmilesWidget(ipw.VBox):
 
         if not self.smiles.value:
             return
-        self.output.value = "Screening possible conformers {}".format(
+        spiner = "Screening possible conformers {}".format(
             self.SPINNER
         )  # font-size:20em;
+        self.output.value = spiner
         self.structure = self._mol_from_smiles(self.smiles.value)
-        self.output.value = ""
+        # Don't overwrite possible error/warning messages
+        if self.output.value == spiner:
+            self.output.value = ""
 
     @default("structure")
     def _default_structure(self):


csadorf:
@danielhollas Once you think this is ready, please request my review if you want to merge this before the end of the week, thank you.

cpignedoli:
@danielhollas great you are improving this widget.
I will try to find the exmaples I had that were rendered properly with openbabel and not with rdikit
it could be is one of the following:
smi="C1(C(C(C2=C(C1=CC=C3C4=CC=C5)C3=C6C7=C2C8=CC9=C7C%10=C%11C(C%12=C%13C%14=C%15C(C(C%16=C%17C(C%18=C%19C%20=C%21C(C(C%22=CC=CC%23=C%22C%21=C(C%19=C%17C%24=C%25)C%25=C%23)=CC=C%26%27)=C%26C%28=C%20C(C(C(C%28=C(C%27=C/%29)NC%29=C/%30)=N/%31)=CC%31=C\C%32=CC=C(/C=C%33C=CC%30=N/%33)N%32)=C%18)=CC%34=C%16C%15=C(C%13=C%11C9=C%35)C%35=C%34)=C%24C=C%36%37)=C%36C%38=C%14C(C(C(C%38=C(C%37=C/%39)NC%39=C/%40)=N/%41)=CC%41=C\C%42=CC=C(/C=C%43C=CC%40=N/%43)N%42)=C%12)=CC%44=C%10C6=C4C5=C%44)=C(C8=C/%45)NC%45=C/%46)=N/%47)=CC%47=C\C%48=CC=C(/C=C%49C=CC%46=N/%49)N%48"
image


smi2="C12=CC=CC3=C1C(CC4=C3C5=C(C(C(C6=CC=CC7=C6C(C8)=CC=C7)=C8C=C9)=C9C=C%10)C%10=CC=C5C=C4)=CC=C2"
image


smi1="C=C1C(C=C(C=C2)C3=CC=C(C4=CC(C=C5)=C(C6=C5C=C(C7=CC=C(C8=CC9=C(C%10=C(C9=C)C=CC=C%10)C=C8C%11=C)C%11=C7)C=C6)C=C4C=C%12)C%12=C3)=C2C(C=C%13%14)=C1C=C%13C%15=CC=CC=C%15C%14=C"
image


smi3="c12c(c(c(c(c1c1=[C]C(=[C][C]=c1c1c2[C]=C([C]=[C]1)Br)Br)C1=[C][C]=[C][C]=[C]1)C1=[C][C]=[C][C]=[C]1)C1=[C][C]=[C][C]=[C]1)C1=[C][C]=[C][C]=[C]1"
image

cpignedoli:
another thing that should be fixed (for the moment I do not know how to deal with this) are smiles like this:

[H][C]1C2=CC=CC3=C2C(C4=C1C=CC5=C4C6=C(C7=C([C]([H])C8=CC=CC9=CC=CC7=C98)C=C%10)C%10=CC=C6C=C5)=CC=C3

that should reproduce the following:
image

This works: C12=CC=CC3=C1C([C]C4=C3C5=C(C(C(C6=CC=CC7=C6C([C]8)=CC=C7)=C8C=C9)=C9C=C%10)C%10=CC=C5C=C4)=CC=C2

image

danielhollas:
@cpignedoli thanks for all the examples! I am looking into it.

danielhollas:
> another thing that should be fixed (for the moment I do not know how to deal with this) are smiles like this:
Could you please provide summary formulas that you would expect for these SMILES?

The first one fails in RDKit even before the 3D embedding with:
> RDKit ERROR: [15:04:02] SMILES Parse Error: syntax error while parsing: HC1C2=CC=CC3=C2C(C4=C1C=CC5=C4C6=C(C7=C(CC8=CC=CC9=CC=CC7=C98)C=C%10)C%10=CC=C6C=C5)=CC=C3

Hard to say what goes wrong. OpenBabel seems to work and generates `C40H23`
image
Is this the correct structure?

The problem with the current OpenBabel code is that it apparently generates structures even from invalid SMILES. E.g. `c1cc1` get's a warning about failed kekulization but is generated anyway. 
For `CC(C)(C)*OO`, the invalid character `*` apparently gets substituted with hydragen and a super weird geometry is produced. So I am a bit torn between failing early with RDKit or allowing OpenBabel to proceed even if it generates garbage.

In any case, I have pushed a new commit. @cpignedoli do you want to try it out or should I let @csadorf  to review this?
btw: I have now also added some error handling to the OpenBabel code as well. Previously, invalid input such as 'Ch' would result in segfault and core dump.

aiidalab_widgets_base/structures.py
```diff
@@ -702,33 +712,63 @@ def _rdkit_opt(self, smiles, steps):
...
+
         AllChem.UFFOptimizeMolecule(mol, maxIters=steps)
-        positions = mol.GetConformer().GetPositions()
+        positions = mol.GetConformer(id=conf_id).GetPositions()
         natoms = mol.GetNumAtoms()
         species = [mol.GetAtomWithIdx(j).GetSymbol() for j in range(natoms)]
         return self._make_ase(species, positions, smiles)
 
-    def _mol_from_smiles(self, smiles, steps=10000):
+    def _mol_from_smiles(self, smiles, steps=1000):
```
danielhollas, line 747: 10000 steps seems excessive. Probably better to bail out early rather than letting user wait forever.

aiidalab_widgets_base/structures.py
```diff
@@ -688,9 +694,13 @@ def _pybel_opt(self, smiles, steps):
         pbmol.localopt(forcefield="mmff94", steps=100)
```
danielhollas, line 694: @cpignedoli could you please comment on this code? It seems we are first optimizing with UFF, then gaff, then mmff94, and finally UFF again?

aiidalab_widgets_base/structures.py
cpignedoli: the correct geometry for: C12=CC=CC3=C1C([C]C4=C3C5=C(C(C(C6=CC=CC7=C6C([C]8)=CC=C7)=C8C=C9)=C9C=C%10)C%10=CC=C5C=C4)=CC=C2

is:

image

the one that fails would have two additional H atoms

aiidalab_widgets_base/structures.py
```diff
@@ -702,33 +712,63 @@ def _rdkit_opt(self, smiles, steps):
 
         smiles = smiles.replace("[", "").replace("]", "")
```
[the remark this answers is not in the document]
cpignedoli: this was introduced to avoid problems we had with some smiles from experimentalists. Not sure what the real problem was @pgasparo should comment on this

aiidalab_widgets_base/structures.py
cpignedoli: also here @pgasparo should comment

yakutovicha:
guys, what is the status here? Is my input/review needed?

danielhollas:
@yakutovicha If you could do a first pass on the review that would be great! I have some open questions about the existing code but they are not blockers for this particular PR.

aiidalab_widgets_base/structures.py
```diff
@@ -688,9 +694,13 @@ def _pybel_opt(self, smiles, steps):
         pbmol.localopt(forcefield="mmff94", steps=100)
 
         f_f = pb._forcefields["uff"]  # pylint: disable=protected-access
-        f_f.Setup(pbmol.OBMol)
+        if not f_f.Setup(pbmol.OBMol):
+            raise ValueError("OpenBabel: Missing UFF parameters")
```
yakutovicha, line 698: Are we sure we need to raise an exception here? We might just want to put the message into `self.output`?

yakutovicha: The same is true for the similar things down below.

aiidalab_widgets_base/structures.py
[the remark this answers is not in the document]
yakutovicha: @cpignedoli, was it somehow related to the use of SMILES format for the GNR?

aiidalab_widgets_base/structures.py
```diff
@@ -702,33 +712,63 @@ def _rdkit_opt(self, smiles, steps):
...
     def _on_button_pressed(self, change):  # pylint: disable=unused-argument
         """Convert SMILES to ase structure when button is pressed."""
         self.output.value = ""
 
         if not self.smiles.value:
             return
-        self.output.value = "Screening possible conformers {}".format(
+        spiner = "Screening possible conformers {}".format(
             self.SPINNER
         )  # font-size:20em;
```
yakutovicha, line 766:
```suggestion
        spiner =f "Screening possible conformers {self.SPINNER}"  # font-size:20em;
```

yakutovicha (commented):
Ok, I gave the first pass, and up to now I have more questions than answers :sweat_smile:

- @cpignedoli can you please remind me of the scope of this widget? Does it suppose to work only with SMILES that adhere to the [openSMILES specification](http://opensmiles.org/opensmiles.html)?
- Related to the previous: I think another goal was to be able to handle GNRs. Would a SMILES that decodes a GNR adhere to the openSMILES specification?
- What is the SMILES specification that rdkit and openbabel are able to deal with?

Depending on the answers, we might want to go different routes: we might want to create a "canonical SMILES" widget that would be able to parse only the "standard" smiles. For GNR we might want to have another class derived from the previous one.

Those are just my open thoughts. I am not that familiar with the whole SMILES story, but I think those general questions above might be crucial to answering before going forward.

aiidalab_widgets_base/structures.py
cpignedoli: I do not remember where the .replace("[") was coming from. In any case the goal is to have a ROBUST app to convert smiles into a geometry NOT to have an app tuned for some non-standard case. So GrapheneNanoRibbons are a very particular case that we typically do not handle through SMILES. if the "[" was a trick to help with GNRs it shoudl then be removed.

aiidalab_widgets_base/structures.py
```diff
@@ -662,7 +662,12 @@ def _make_ase(self, species, positions, smiles):
         """Create ase Atoms object."""
         # Get the principal axes and realign the molecule along z-axis.
         positions = PCA(n_components=3).fit_transform(positions)
-        atoms = Atoms(species, positions=positions, pbc=True)
+        try:
+            atoms = Atoms(species, positions=positions, pbc=False)
+        except KeyError:
```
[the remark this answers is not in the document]
cpignedoli: the only thing I kow is
"We removed 720 768 duplicates by comparing reaction strings without atom mapping and an additional 780 reactions, because the SMILES string could not be canonicalized with RDKit,[34](https://pubs.rsc.org/en/content/articlelanding/2018/sc/c8sc02339e#cit34) as the explicit number of valence electrons for one of the atoms was greater than permitted." DOI	https://doi.org/10.1039/C8SC02339E

yakutovicha:
@danielhollas I would like to suggest something, let me know if you agree.

It is clear that this class needs some additional work, but we should not strive to do everything in a single PR, otherwise, we will never end.

It appears to me that many discussion points deserve their own issue and separate testing.

Could you maybe create separate issues for the open questions? This would allow us to move forward with what we already have in this PR.

danielhollas:
@yakutovicha indeed, the discussion here got a bit broader than I originally intended. I like your suggestion. I am also planning to invest some time to make the structure generation from SMILES as robust as possible (the goal is to actually generate multiple conformers, not just one structure) so I think the followup PRs will stem from that work.

Unfortunately, my dev machine just got packed and shipped away just an hour ago since we're moving to a different university, so it will unfortunately take some time till I am able to work on this again. :-(

aiidalab_widgets_base/structures.py
danielhollas: @cpignedoli Could you provide please provide a link if the GNR work was published.

aiidalab_widgets_base/structures.py
danielhollas: So I am actually catching `ValueError` on line 754 when calling this function (as it was kind of done in the original implementation), and appending the error to self.output when the exception is caught. Not sure what is preferable here.

yakutovicha:
> Unfortunately, my dev machine just got packed and shipped away just an hour ago since we're moving to a different university, so it will unfortunately take some time till I am able to work on this again. :-(

Thanks, @danielhollas and no problem at all, let me know when you need my input.

danielhollas:
Closing this. Basic error handling was merged in #321, and more substantial changes to improve robustness of RDKit generation will (hopefully) come in a separate PR (e.g. today I learned from another group that we might want to switch the UFF force field for MMFF94, or basically anything better. I am currently working on an extension that will do conformer sampling from a single SMILES string and optimize with semiempirical xTB method. If anybody is interested in this work, don't hesitate to contact me. :smile:

PR #293, opened 2022-02-21
Remove arguments from all the remaining `.super()` calls.

yakutovicha:
Otherwise, the objects fail to initialise.


yakutovicha:
mentioning @cpignedoli as it should fix the issue he recently encountered.

PR #299, opened 2022-03-12
NodesTreeWidget: _find_outputs parse nested namespace outputs

### LINKED ISSUE #262: NodesTreeWidget: find outputs not working for nested namespace

unkcpz:
fix #262

Previous PR #269 only works when the outputs have a single layer of
the nested namespace. We store the `parent_pk` in order to transversal nodes
level by level, however only with extra stored namespace parameter is not enough to access the
outputs in deep level. To solve this, `namespace` replace by `namespaces` which is a 
tuple to store the full path so that all the output nodes can be visited.

### DIFF
--- a/aiidalab_widgets_base/nodes.py
+++ b/aiidalab_widgets_base/nodes.py
@@ -216,15 +216,14 @@ class NodesTreeWidget(ipw.Output):
     def _find_outputs(cls, root):
         process_node = load_node(root.parent_pk)
 
-        outputs = process_node.outputs
-        # if output tree node is from namespace node
-        namespaces = root.namespaces
-        if namespaces:
-            outputs = functools.reduce(
-                lambda d, namespace: d[namespace], namespaces, process_node.outputs
-            )
+        # Gather outputs from node and its namespaces (if present):
+        outputs = functools.reduce(
+            lambda d, namespace: d[namespace],
+            root.namespaces or [],
+            process_node.outputs,
+        )
 
-        # convert aiida LinkManager to type dict
+        # Convert aiida.orm.LinkManager to dict
         output_nodes = {key: outputs[key] for key in outputs}
 
         for key in sorted(
@@ -236,7 +235,7 @@ class NodesTreeWidget(ipw.Output):
                 yield AiidaOutputsTreeNode(
                     name=key,
                     parent_pk=root.parent_pk,
-                    namespaces=(*namespaces, key),
+                    namespaces=(*root.namespaces, key),
                 )
 
             else:
--- a/aiidalab_widgets_base/nodes.py
+++ b/aiidalab_widgets_base/nodes.py
@@ -215,10 +215,11 @@ class NodesTreeWidget(ipw.Output):
     @classmethod
     def _find_outputs(cls, root):
         """
-        A generator for all namespace and output nodes from the output root.
-        If a namespace encountered it generate a AiidaOutputsTreeNode with
-        the namespaces path stored so it can be accessed by root node and namespaces tuple
-        in walk tree breadth-first search.
+        A generator for all (including nested) output nodes.
+        
+        Generates an AiidaOutputsTreeNode when encountering a namespace,
+        keeping track of the full namespace path to make it accessible via the
+        root node in form of a breadth-first search.
         """
         process_node = load_node(root.parent_pk)
 
--- a/aiidalab_widgets_base/nodes.py
+++ b/aiidalab_widgets_base/nodes.py
@@ -214,16 +214,22 @@ class NodesTreeWidget(ipw.Output):
 
     @classmethod
     def _find_outputs(cls, root):
+        """
+        A generator for all namespace and output nodes from the output root.
+        If a namespace encountered it generate a AiidaOutputsTreeNode with
+        the namespaces path stored so it can be accessed by root node and namespaces tuple
+        in walk tree breadth-first search.
+        """
         process_node = load_node(root.parent_pk)
 
-        # Gather outputs from node and its namespaces (if present):
+        # Gather outputs from node and its namespaces:
         outputs = functools.reduce(
-            lambda d, namespace: d[namespace],
+            lambda attr_dict, namespace: attr_dict[namespace],
             root.namespaces or [],
             process_node.outputs,
         )
 
-        # Convert aiida.orm.LinkManager to dict
+        # Convert aiida.orm.LinkManager or AttributDict (if namespace presented) to dict
         output_nodes = {key: outputs[key] for key in outputs}
 
         for key in sorted(
@@ -235,7 +241,7 @@ class NodesTreeWidget(ipw.Output):
                 yield AiidaOutputsTreeNode(
                     name=key,
                     parent_pk=root.parent_pk,
-                    namespaces=(*root.namespaces, key),
+                    namespaces=(*root.namespaces, key),  # attach nested namespace name
                 )
 
             else:
--- a/aiidalab_widgets_base/nodes.py
+++ b/aiidalab_widgets_base/nodes.py
@@ -1,4 +1,7 @@
 """Widgets to work with AiiDA nodes."""
+import functools
+from typing import Tuple
+
 import ipywidgets as ipw
 import traitlets
 from aiida.cmdline.utils.ascii_vis import calc_info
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/nodes.py
```diff
@@ -210,22 +215,30 @@ def _find_called(cls, root):
...
-        else:
-            outputs = process_node.outputs
 
+        outputs = process_node.outputs
+        # if output tree node is from namespace node
+        namespaces = root.namespaces
+        if namespaces:
+            outputs = functools.reduce(
+                lambda d, namespace: d[namespace], namespaces, process_node.outputs
+            )
```
csadorf, line 225: I find this segment harder to understand now. Would the following refactoring be correct?
```suggestion
        # Gather outputs from node and its namespaces (if present):
        outputs = functools.reduce(
            lambda d, namespace: d[namespace], root.namespaces or [], process_node.outputs
        )
```

aiidalab_widgets_base/nodes.py
```diff
@@ -210,22 +215,30 @@ def _find_called(cls, root):
...
+
+        # convert aiida LinkManager to type dict
```
csadorf, line 227:
```suggestion
        # Convert aiida.orm.LinkManager to dict
```

aiidalab_widgets_base/nodes.py
```diff
@@ -210,22 +215,30 @@ def _find_called(cls, root):
...
         ):
             node = output_nodes[key]
-
             if isinstance(node, AttributeDict):
+                # for namespace tree node attach label and continue recursively
                 yield AiidaOutputsTreeNode(
-                    name=key, parent_pk=root.parent_pk, namespace=key
+                    name=key,
+                    parent_pk=root.parent_pk,
+                    namespaces=(*namespaces, key),
```
csadorf, line 239: In combination with above suggestion:
```suggestion
                    namespaces=(*root.namespaces, key),
```

csadorf (commented):
This looks sound to me, but I think I would prefer to have a better understanding of the namespaces. Can you point me to the relevant section in the AiiDA documentation?

unkcpz:
@csadorf Thanks a lot! Indeed, the code is more clear with your suggestion, updated.

> Can you point me to the relevant section in the AiiDA documentation?

The introduction of `namespace`  is in the process part of documentation https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/processes/usage.html?highlight=namespace#ports-and-port-namespaces 

It is then more detail described in the 'Exposing output and input' part https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/write_workflows.html?highlight=namespace#exposing-inputs-and-outputs since this is the most common way plugin developers use namespace concept implicitly, such as in the `QeWorkChain` in aiidalab-qe.

aiidalab_widgets_base/nodes.py
```diff
@@ -210,22 +215,29 @@ def _find_called(cls, root):
...
 
+        # Gather outputs from node and its namespaces (if present):
+        outputs = functools.reduce(
+            lambda d, namespace: d[namespace],
+            root.namespaces or [],
+            process_node.outputs,
+        )
+
+        # Convert aiida.orm.LinkManager to dict
         output_nodes = {key: outputs[key] for key in outputs}
```
csadorf, line 227: What would happen if namespaces have outputs with the same key?

aiidalab_widgets_base/nodes.py
```diff
@@ -210,22 +215,29 @@ def _find_called(cls, root):
...
             if isinstance(node, AttributeDict):
+                # for namespace tree node attach label and continue recursively
                 yield AiidaOutputsTreeNode(
-                    name=key, parent_pk=root.parent_pk, namespace=key
+                    name=key,
+                    parent_pk=root.parent_pk,
+                    namespaces=(*root.namespaces, key),
                 )
+
             else:
```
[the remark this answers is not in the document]
unkcpz: I don't think this change is correct. The `else` branch is for the condition that the node is not a namespace but a true data node. While the `if node.pk not in root.nodes_registry` is for the caching tree nodes.

aiidalab_widgets_base/nodes.py
unkcpz: This should already be validated by aiida, it is not allowed to have duplicate keys under one namespace.

aiidalab_widgets_base/nodes.py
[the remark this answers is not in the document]
csadorf: Yes, sorry about that.

aiidalab_widgets_base/nodes.py
unkcpz: >  there can be no situation where outputs from different namespaces could share the same key ?

For the different namespaces, the keys can be the same. But the 'functools.reduce' unfold the namespace to the innermost nested specific namespace. For one namespace, keys can't be the same.

aiidalab_widgets_base/nodes.py
```diff
@@ -210,22 +215,29 @@ def _find_called(cls, root):
...
-            outputs = process_node.outputs[root.namespace]
-        else:
-            outputs = process_node.outputs
 
+        # Gather outputs from node and its namespaces (if present):
+        outputs = functools.reduce(
+            lambda d, namespace: d[namespace],
+            root.namespaces or [],
+            process_node.outputs,
+        )
```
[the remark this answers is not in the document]
unkcpz: No worries, I admit this is kind of confused, every time I look at it again I need to go through all the code and try to clean my mind.

Assuming we have the following outputs of a process node `wc_node`:
```
ns00
    ns10
        outnode_a
        outnode_b
ns01
        outnode_a
        outnode_c
```
If we want to access `outnode_a` node, it should be `wc_node.outputs['ns00']['ns01']['outnode_a']`. In the scope (iteration), the `root.namespaces` in this layer equal to `[`ns00`, `ns01`]`.

> Either way, this function probably requires a bit more commenting since it is somewhat intransparent.
Indeed, I will add it if you think the code and explanation above are clear enough.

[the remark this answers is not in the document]
unkcpz: `d[namespace]` in the `reduce` make sure that the `d` accumulate the change for every nested `namespace`.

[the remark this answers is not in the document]
csadorf: From offline discussion: The code appears to be correct, however it will require a bit more documentation to make it easier to follow. After that I'm happy to approve and merge.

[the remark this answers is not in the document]
unkcpz: Hi @csadorf, I add docstring and comment to make (try to🤣) it more clear of the logic. Please have a look.

aiidalab_widgets_base/nodes.py
```diff
@@ -209,23 +214,36 @@ def _find_called(cls, root):
 
     @classmethod
     def _find_outputs(cls, root):
+        """
+        A generator for all namespace and output nodes from the output root.
+        If a namespace encountered it generate a AiidaOutputsTreeNode with
+        the namespaces path stored so it can be accessed by root node and namespaces tuple
+        in walk tree breadth-first search.
```
csadorf, line 221:
```suggestion
        A generator for all (including nested) output nodes.
        
        Generates an AiidaOutputsTreeNode when encountering a namespace,
        keeping track of the full namespace path to make it accessible via the
        root node in form of a breadth-first search.
```

csadorf (approved):
One suggestion to reformulate the doc-string a bit, but otherwise good to merge. Thanks a lot! 👍

unkcpz:
@csadorf Thanks! Docstring updated and ready to be merged 😄 please approve once more.

csadorf (approved):

PR #305, opened 2022-04-14
Fix `AiidaComputerSetup` widget.

yakutovicha:
The widget wasn't working correctly as during the computer configuration
the value for the ssh port was of type "str". The issue is fixed
by converting the value type to "int".

Additionally, the `_configure_computer` function got slightly simplified by using
the `Computer.configure()` method provided by AiiDA.

### DIFF
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -697,22 +697,26 @@ class AiidaComputerSetup(ipw.VBox):
 
         super().__init__(children, **kwargs)
 
-    def _configure_computer(self):
-        """Create AuthInfo."""
+    def _configure_computer(self, computer):
+        """Configure the computer"""
         sshcfg = parse_sshconfig(self.hostname.value)
         authparams = {
-            "compress": True,
+            "port": int(sshcfg.get("port", 22)),
+            "look_for_keys": True,
             "key_filename": os.path.expanduser(
                 sshcfg.get("identityfile", ["~/.ssh/id_rsa"])[0]
             ),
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/computational_resources.py
```diff
@@ -728,13 +732,8 @@ def _configure_computer(self):
...
         elif "proxyjump" in sshcfg:
             authparams["proxy_jump"] = sshcfg["proxyjump"]
-        aiidauser = orm.User.objects.get_default()
 
-        authinfo = orm.AuthInfo(
-            computer=orm.Computer.objects.get(label=self.label.value), user=aiidauser
-        )
-        authinfo.set_auth_params(authparams)
-        authinfo.store()
+        computer.configure(**authparams)
```
csadorf, line 736: So this function does what the previous code did w.r.t to `authinfo`?

csadorf (commented):
It seems that the PR description does not really match/motivate the proposed change here. Can you clarify that? Otherwise it looks good to me.

aiidalab_widgets_base/computational_resources.py
yakutovicha: Well, here I incorporated a function provided by AiiDA that does exactly the same. Just a little simplification.

I will now modify the PR description to better describe what has happened here.

yakutovicha: done.

csadorf: Thanks!

csadorf (approved):
Good to go! 🚢

PR #306, opened 2022-04-14
ComputationalResourcesWidget refresh things after every update

### LINKED ISSUE #301: ComputationalResourcesWidget: computer and code setting need refresh to see the change
The expected behavior of computer and code setting with `Setup new code` -> `Detailed setup` is when a new computer setup, in the next code setup substep, the newly added computer appears in the dropdown list. This requires updating the computers dropdown list when `setup` button is clicked. 

The same issue for the codes list when a code is newly added, have to refresh the page to see the new code.

yakutovicha:
fixes #301
In this PR the ComputationalResourcesWidget's behaviour is improved:
 - After each computer setup the code widget is refreshed.
 - After each code setup, the list of available codes is refreshed.

Additionally, there were a few changes to achieve the things mentioned above:
- `SshComputerSetup.on_setup_ssh()`, `AiidaComputerSetup.on_setup_computer()`,
`AiidaCodeSetup.on_setup_code()` are now returning `False` in case of a failure, and `True`
in case of a success.
- `AiidaCodeSetup.refresh()` method is implemented. It updates the content of the widget with
the latest changes.
- Simplify `ssh_copy_id` function that raises an `RuntimeError` in case the connection to a
remote computer couldn't  be established. The exception is then handled by the
`SshComputerSetup.on_setup_ssh()` method.

### DIFF
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -143,9 +143,8 @@ class ComputationalResourcesWidget(ipw.VBox):
         with self.hold_trait_notifications():
             success = self.ssh_computer_setup.on_setup_ssh()
             if success:
-                success = self.aiida_computer_setup.on_setup_computer()
-            if success:
-                self.aiida_code_setup.on_setup_code()
+                if self.aiida_computer_setup.on_setup_computer():
+                    self.aiida_code_setup.on_setup_code()
 
     def _get_codes(self):
         """Query the list of available codes."""
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -265,7 +265,7 @@ class SshComputerSetup(ipw.VBox):
         SshConnectionState, allow_none=True, default_value=None
     )
     SSH_POSSIBLE_RESPONSES = [
-        # order matters! the index will return by pexpect and compare 
+        # order matters! the index will return by pexpect and compare
         # with SshConnectionState
         "[Pp]assword:",  # 0
         "Now try logging into",  # 1
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -265,7 +265,7 @@ class SshComputerSetup(ipw.VBox):
         SshConnectionState, allow_none=True, default_value=None
     )
     SSH_POSSIBLE_RESPONSES = [
-        # order matters! the index will return by pexpect and compare
+        # order matters! the index will return by pexpect and compare 
         # with SshConnectionState
         "[Pp]assword:",  # 0
         "Now try logging into",  # 1
@@ -546,8 +546,6 @@ class SshComputerSetup(ipw.VBox):
         elif self.ssh_connection_state is SshConnectionState.do_you_want_to_continue:
             self._ssh_connection_process.sendline("yes")
 
-        self.ssh_connection_state = SshConnectionState.waiting_for_input
-
     def _handle_ssh_password(self):
         """Send a password to a remote computer."""
         message = (
@@ -565,6 +563,8 @@ class SshComputerSetup(ipw.VBox):
             self._ssh_password.disabled = False
             self._continue_with_password_button.disabled = False
             self._ssh_connection_message = message
+
+        self.ssh_connection_state = SshConnectionState.waiting_for_input
 
     def _on_verification_mode_change(self, change):
         """which verification mode is chosen."""
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -141,8 +141,7 @@ class ComputationalResourcesWidget(ipw.VBox):
     def quick_setup(self, _=None):
         """Go through all the setup steps automatically."""
         with self.hold_trait_notifications():
-            success = self.ssh_computer_setup.on_setup_ssh()
-            if success:
+            if self.ssh_computer_setup.on_setup_ssh():
                 if self.aiida_computer_setup.on_setup_computer():
                     self.aiida_code_setup.on_setup_code()
 
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -99,11 +99,11 @@ class ComputationalResourcesWidget(ipw.VBox):
             (self.comp_resources_database, "code_setup"),
             (self.aiida_code_setup, "code_setup"),
         )
-        self.aiida_code_setup.on_setup_code_success.append(self.refresh)
+        self.aiida_code_setup.on_setup_code_success(self.refresh)
 
         # After a successfull computer setup the codes widget should be refreshed.
         # E.g. the list of available computers needs to be updated.
-        self.aiida_computer_setup.on_setup_computer_success.append(
+        self.aiida_computer_setup.on_setup_computer_success(
             self.aiida_code_setup.refresh
         )
 
@@ -565,7 +565,7 @@ class AiidaComputerSetup(ipw.VBox):
 
     def __init__(self, **kwargs):
 
-        self.on_setup_computer_success = []
+        self._on_setup_computer_success = []
 
         # List of widgets to be displayed.
         self.label = ipw.Text(
@@ -742,7 +742,7 @@ class AiidaComputerSetup(ipw.VBox):
             try:
                 computer = orm.Computer.objects.get(label=self.label.value)
                 print(f"A computer called {self.label.value} already exists.")
-                for function in self.on_setup_computer_success:
+                for function in self._on_setup_computer_success:
                     function()
                 return True
             except common.NotExistent:
@@ -780,10 +780,13 @@ class AiidaComputerSetup(ipw.VBox):
                 return False
 
             if self._configure_computer(computer):
-                for function in self.on_setup_computer_success:
+                for function in self._on_setup_computer_success:
                     function()
             print(f"Computer<{computer.pk}> {computer.label} created")
             return True
+
+    def on_setup_computer_success(self, function):
+        self._on_setup_computer_success.append(function)
 
     def test(self, _=None):
         with self._test_out:
@@ -833,7 +836,7 @@ class AiidaCodeSetup(ipw.VBox):
 
     def __init__(self, path_to_root="../", **kwargs):
 
-        self.on_setup_code_success = []
+        self._on_setup_code_success = []
 
         # Code label.
         self.label = ipw.Text(
@@ -956,12 +959,15 @@ class AiidaCodeSetup(ipw.VBox):
                 print(f"Unable to store the Code: {exception}")
                 return False
 
-            for function in self.on_setup_code_success:
+            for function in self._on_setup_code_success:
                 function()
 
             print(f"Code<{code.pk}> {code.full_label} created")
 
             return True
+
+    def on_setup_code_success(self, function):
+        self._on_setup_code_success.append(function)
 
     def _reset(self):
         self.label.value = ""
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -16,6 +16,7 @@ from aiida.transports.plugins.ssh import parse_sshconfig
 from IPython.display import clear_output, display
 
 from .databases import ComputationalResourcesDatabaseWidget
+from .utils import StatusHTML
 
 STYLE = {"description_width": "180px"}
 LAYOUT = {"width": "400px"}
@@ -54,7 +55,7 @@ class ComputationalResourcesWidget(ipw.VBox):
         description (str): Description to display before the dropdown.
         """
         self.output = ipw.HTML()
-
+        self.setup_message = StatusHTML()
         self.code_select_dropdown = ipw.Dropdown(
             description=description, disabled=True, value=None
         )
@@ -84,11 +85,20 @@ class ComputationalResourcesWidget(ipw.VBox):
 
         self.ssh_computer_setup = SshComputerSetup()
         ipw.dlink(
+            (self.ssh_computer_setup, "message"),
+            (self.setup_message, "message"),
+        )
+
+        ipw.dlink(
             (self.comp_resources_database, "ssh_config"),
             (self.ssh_computer_setup, "ssh_config"),
         )
 
         self.aiida_computer_setup = AiidaComputerSetup()
+        ipw.dlink(
+            (self.aiida_computer_setup, "message"),
+            (self.setup_message, "message"),
+        )
         ipw.dlink(
             (self.comp_resources_database, "computer_setup"),
             (self.aiida_computer_setup, "computer_setup"),
@@ -96,6 +106,10 @@ class ComputationalResourcesWidget(ipw.VBox):
 
         # Set up AiiDA code.
         self.aiida_code_setup = AiidaCodeSetup()
+        ipw.dlink(
+            (self.aiida_code_setup, "message"),
+            (self.setup_message, "message"),
+        )
         ipw.dlink(
             (self.comp_resources_database, "code_setup"),
             (self.aiida_code_setup, "code_setup"),
@@ -115,8 +129,6 @@ class ComputationalResourcesWidget(ipw.VBox):
             children=[
                 self.ssh_computer_setup.username,
                 quick_setup_button,
-                self.ssh_computer_setup.setup_ssh_out,
-                self.aiida_computer_setup.setup_compupter_out,
                 self.aiida_code_setup.setup_code_out,
             ]
         )
@@ -142,7 +154,7 @@ class ComputationalResourcesWidget(ipw.VBox):
     def quick_setup(self, _=None):
         """Go through all the setup steps automatically."""
         with self.hold_trait_notifications():
-            self.ssh_computer_setup.on_setup_ssh()
+            self.ssh_computer_setup._on_setup_ssh_button_pressed()
             if self.aiida_computer_setup.on_setup_computer():
                 self.aiida_code_setup.on_setup_code()
 
@@ -222,6 +234,8 @@ class ComputationalResourcesWidget(ipw.VBox):
                         """Please select the computer/code from a database to pre-fill the fields below."""
                     ),
                     self.comp_resources_database,
+                    self.setup_message,
+                    self.ssh_computer_setup.password_box,
                     self.output_tab,
                 )
             else:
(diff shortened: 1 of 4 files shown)


yakutovicha:
@unkcpz I am not sure if you got notified about this PR, but maybe you can also quickly test it to make sure it resolves all the #301 issue.

unkcpz:
@yakutovicha thanks! I will do it.

> I am not sure if you got notified about this PR

I only turn the notification of the repo when being mentioned.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -129,16 +139,13 @@ def __init__(self, description="Select code:", path_to_root="../", **kwargs):
...
-            self.aiida_computer_setup.on_setup_computer(on_success=setup_code)
-
+        """Go through all the setup steps automatically."""
         with self.hold_trait_notifications():
-            self.ssh_computer_setup.on_setup_ssh(on_success=setup_code_and_computer)
+            success = self.ssh_computer_setup.on_setup_ssh()
+            if success:
+                success = self.aiida_computer_setup.on_setup_computer()
+            if success:
+                self.aiida_code_setup.on_setup_code()
```
csadorf, line 148:
```suggestion
            if success:
                if self.aiida_computer_setup.on_setup_computer():
                    self.aiida_code_setup.on_setup_code()
```
However, I think it would be better to use an exception to indicate a failure state.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -385,19 +392,19 @@ def on_setup_ssh(self, _=None, on_success=None):
             # If hostname & username are not provided - do not do anything.
             if self.hostname.value == "":  # check hostname
                 print("Please specify the computer hostname.")
-                return
+                return False
```
csadorf, line 395: See also comment above. I think we should use a `RuntimeError` to indicate the error state.
Add the following exception to the module:
```
class _SshConfigError(RuntimeError):
    pass
```

csadorf (changes requested):
One strong and one weak change request, otherwise good.

aiidalab_widgets_base/computational_resources.py
yakutovicha: I think raising an exception won't work here as I am also calling the `on_setup_ssh()` method by ipywidgets.

The other possibility would be to have some other function that is raising a `RuntimeError` in case of an issue. I would then call this new function inside `on_setup_ssh()` and handle the exception there.

Maybe, keeping it as-is is still better.

what do you think?

aiidalab_widgets_base/computational_resources.py
```diff
@@ -837,6 +834,8 @@ class AiidaCodeSetup(ipw.VBox):
 
     def __init__(self, path_to_root="../", **kwargs):
 
+        self.on_setup_code_success = []
```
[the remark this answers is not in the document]
yakutovicha: ok.

unkcpz:
Hi @yakutovicha, I test it by setup a new computer the auto refresh works fine. But a minor issue could be fixed in this PR.  

However, an exception encountered when `Setup new code` -> password-less SSH -> fill fields for computer `imxgesrv1.epfl.ch` -> using password -> if password is empty or wrong:

```
/opt/conda/lib/python3.7/site-packages/traitlets/traitlets.py in _notify_observers(self, event)
   1264                 c = getattr(self, c.name)
   1265 
-> 1266             c(event)
   1267 
   1268     def _add_notifiers(self, handler, name, type):

~/.local/lib/python3.7/site-packages/aiidalab_widgets_base/utils/__init__.py in next_i(change)
    104             def next_i(change):
    105                 try:
--> 106                     i.send(change.new)
    107                 except StopIteration:
    108                     widget.unobserve(next_i, attribute)

~/.local/lib/python3.7/site-packages/aiidalab_widgets_base/computational_resources.py in ssh_copy_id()
    475                     else:
    476                         raise RuntimeError(
--> 477                             f"Failed\nUnknown problem.\n{child.before}\n{child.after}"
    478                         )
    479                 child.close()

RuntimeError: Failed
Unknown problem.
b' \r\njyu@ela.cscs.ch: Permission denied (publickey,password).\r\r\n'
<class 'pexpect.exceptions.EOF'>
```

I think better to give a prompt message and allow user to reinput the password.

unkcpz:
I also test set password-less SSH setting with private key and get error I report an issue https://github.com/aiidalab/aiidalab-qe/issues/226. Not necessary to be fixed here.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -129,16 +139,12 @@ def __init__(self, description="Select code:", path_to_root="../", **kwargs):
...
-        def setup_code_and_computer():
-            self.aiida_computer_setup.on_setup_computer(on_success=setup_code)
-
+        """Go through all the setup steps automatically."""
         with self.hold_trait_notifications():
-            self.ssh_computer_setup.on_setup_ssh(on_success=setup_code_and_computer)
+            success = self.ssh_computer_setup.on_setup_ssh()
+            if success:
+                if self.aiida_computer_setup.on_setup_computer():
+                    self.aiida_code_setup.on_setup_code()
```
csadorf, line 147: Why stop there? 😆 
```suggestion
            if self.ssh_computer_setup.on_setup_ssh():
                if self.aiida_computer_setup.on_setup_computer():
                    self.aiida_code_setup.on_setup_code()
```

aiidalab_widgets_base/computational_resources.py
yakutovicha: I was referring to this:

https://github.com/aiidalab/aiidalab-widgets-base/blob/419449b387b15206e3191f868b1f57098ecf51cf/aiidalab_widgets_base/computational_resources.py#L302-L304

`Setup ssh` button calls this function, so raising an exception in it doesn't look like a good option.

yakutovicha:
@unkcpz, I think the issue you reported earlier should be fixed. I followed the suggestion of @csadorf and introduced a trait that represents the current state of the connection. I tested things locally, everything seems to be alright.

unkcpz:
@yakutovicha thanks! I will test it again and review it late today.

unkcpz:
Hi @yakutovicha, sorry for the late. I still got the exception when I input the wrong password but the error message is now a little bit different which stem from other part of the code:

```
RuntimeError                              Traceback (most recent call last)
/opt/conda/lib/python3.7/site-packages/ipywidgets/widgets/widget.py in _handle_msg(self, msg)
    674                 if 'buffer_paths' in data:
    675                     _put_buffers(state, data['buffer_paths'], msg['buffers'])
--> 676                 self.set_state(state)
    677 
    678         # Handle a state request.

/opt/conda/lib/python3.7/site-packages/ipywidgets/widgets/widget.py in set_state(self, sync_data)
    543                     from_json = self.trait_metadata(name, 'from_json',
    544                                                     self._trait_from_json)
--> 545                     self.set_trait(name, from_json(sync_data[name], self))
    546 
    547     def send(self, content, buffers=None):

/opt/conda/lib/python3.7/contextlib.py in __exit__(self, type, value, traceback)
    117         if type is None:
    118             try:
--> 119                 next(self.gen)
    120             except StopIteration:
    121                 return False

/opt/conda/lib/python3.7/site-packages/traitlets/traitlets.py in hold_trait_notifications(self)
   1214                 for changes in cache.values():
   1215                     for change in changes:
-> 1216                         self.notify_change(change)
   1217 
   1218     def _notify_trait(self, name, old_value, new_value):

/opt/conda/lib/python3.7/site-packages/ipywidgets/widgets/widget.py in notify_change(self, change)
    604                 # Send new state to front-end
    605                 self.send_state(key=name)
--> 606         super(Widget, self).notify_change(change)
    607 
    608     def __repr__(self):

/opt/conda/lib/python3.7/site-packages/traitlets/traitlets.py in notify_change(self, change)
   1227     def notify_change(self, change):
   1228         """Notify observers of a change event"""
-> 1229         return self._notify_observers(change)
   1230 
   1231     def _notify_observers(self, event):

/opt/conda/lib/python3.7/site-packages/traitlets/traitlets.py in _notify_observers(self, event)
   1264                 c = getattr(self, c.name)
   1265 
-> 1266             c(event)
   1267 
   1268     def _add_notifiers(self, handler, name, type):

~/.local/lib/python3.7/site-packages/aiidalab_widgets_base/utils/__init__.py in next_i(change)

~/.local/lib/python3.7/site-packages/aiidalab_widgets_base/computational_resources.py in ssh_copy_id()
    475         self._ssh_connection_message = None
    476         self._ssh_connection_process = None
--> 477 
    478     @traitlets.observe("ssh_connection_state")
    479     def _observe_ssh_connnection_state(self, _=None):

RuntimeError: Failed
Unknown problem.
b' \r\njyu@imxgesrv1.epfl.ch: Permission denied (publickey,password).\r\r\n'
<class 'pexpect.exceptions.EOF'>
```

unkcpz (changes requested):

unkcpz:
thanks! The logic makes sense and is tested fine. Just minor change requests.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -225,12 +231,41 @@ def _setup_new_code(self, _=None):
...
+
 class SshComputerSetup(ipw.VBox):
     """Setup password-free access to a computer."""
 
     ssh_config = traitlets.Dict()
+    ssh_connection_state = traitlets.UseEnum(
+        SshConnectionState, allow_none=True, default_value=None
+    )
+    SSH_POSSIBLE_RESPONSES = [
+        "assword:",  # 0
```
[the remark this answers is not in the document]
yakutovicha: this wouldn't work, unfortunately. Sometimes SSH says "Password", sometimes "ela.cscs.ch's password"

unkcpz (commented):
@yakutovicha I think you just have a discussion with Simon, I'll let him to approve this.

yakutovicha:
@csadorf, @unkcpz - as we've discussed, I leave this PR up to you to play around with and do further improvements, if necessary.

aiidalab_widgets_base/utils/__init__.py
```diff
@@ -1,10 +1,16 @@
 """Some utility functions used acrross the repository."""
-from functools import wraps
 
+# +
```
unkcpz, line 3: This can be removed and the same below.

aiidalab_widgets_base/utils/__init__.py
```diff
@@ -131,3 +107,47 @@ def screen_to_vector(self, move_vector):
     @property
     def inverse_matrix(self):
         return np.linalg.inv(self.matrix)
+
+
+# +
```
unkcpz, line 112: Same here.

unkcpz (commented):
@csadorf We can go ahead and try to finish this? I add a quick commit for the `#` mentioned to be removed

aiidalab_widgets_base/computational_resources.py
```diff
@@ -225,12 +245,56 @@ def _setup_new_code(self, _=None):
...
+class SshConnectionState(enum.Enum):
+    waiting_for_input = -1
+    enter_password = 0
+    success = 1
+    keys_already_present = 2
+    do_you_want_to_continue = 3
+    no_keys = 4
+    unknown_hostname = 5
+    connection_refused = 6
+    end_of_file = 7
```
csadorf, line 257: We can probably use [`auto`](https://docs.python.org/3.7/library/enum.html#using-automatic-values) here, right?

aiidalab_widgets_base/computational_resources.py
```diff
@@ -375,126 +437,131 @@ def _write_ssh_config(self, private_key_abs_fname=None):
...
+
+    def _send_password(self, _=None):
+        self._ssh_password.disabled = True
+        self._continue_with_password_button.disabled = True
+        self._ssh_connection_process.sendline(self._ssh_password.value)
+
+    @traitlets.observe("ssh_connection_state")
+    def _observe_ssh_connnection_state(self, _=None):
+        """Observe the ssh connection state and act according to the changes."""
+        if self.ssh_connection_state == SshConnectionState.waiting_for_input:
```
csadorf, line 515: [Enums should be compared by identity](https://docs.python.org/3.7/library/enum.html#using-automatic-values).

csadorf (commented):
No obvious issues, although I would think that the code structure could be significantly introduced by separating all business logic related functions (SSH-setup, computer-setup, code-setup) into separate modules. Right now everything is mixed with the widget implementation. In general, it is better to try to separate these two things as much as possible (roughly following a [Model-view-controller](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) pattern).

aiidalab_widgets_base/computational_resources.py
```diff
@@ -375,126 +437,131 @@ def _write_ssh_config(self, private_key_abs_fname=None):
...
             try:
-                ssh_copy_id()
-            except StopIteration:
-                print(f"Unsuccessful attempt to connect to {self.hostname.value}.")
-                return
+                self.ssh_connection_state = self._ssh_connection_process.expect(
+                    self.SSH_POSSIBLE_RESPONSES,
+                    timeout=timeout,
+                )
+            except pexpect.TIMEOUT:
```
csadorf, line 485: This means that if the user did not provide a password within 30 seconds, the process times out and one needs to start all over again. That's a bit of a heavy-handed approach for synchronization between the widget threads and this thread, but I think it is acceptable.

I think there are probably better approaches to this, but I would recommend to move forward with this for now.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -375,126 +437,131 @@ def _write_ssh_config(self, private_key_abs_fname=None):
...
+        # Copy public key on the remote computer.
+        ssh_connection_thread = threading.Thread(target=self._ssh_copy_id)
+        ssh_connection_thread.start()
+
+    def _ssh_copy_id(self):
+        """Run the ssh-copy-id command and follow it until it is completed."""
+        timeout = 30
+        self.password_message = f"Sending public key to {self.hostname.value}... "
+        self._ssh_connection_process = pexpect.spawn(
+            f"ssh-copy-id {self.hostname.value}"
```
csadorf, line 477: I assume it is the case since that's how we are using it here, but does ssh-copy-id automatically copy the public key to the jump hosts as well?

aiidalab_widgets_base/computational_resources.py
```diff
@@ -225,12 +245,56 @@ def _setup_new_code(self, _=None):
                 }
 
 
+class SshConnectionState(enum.Enum):
```
csadorf, line 248: I had not noticed that the `SshConnectionState` and the `SshPossibleRepsonses` from below are intimately linked via their order. They must therefore be part of the same data structure to avoid getting out of sync. Or at the very least there must be a comment that makes this intimate connection clear.

csadorf (changes requested):

aiidalab_widgets_base/computational_resources.py
```diff
@@ -230,12 +231,41 @@ def _setup_new_code(self, _=None):
...
+
 class SshComputerSetup(ipw.VBox):
     """Setup password-free access to a computer."""
 
     ssh_config = traitlets.Dict()
+    ssh_connection_state = traitlets.UseEnum(
+        SshConnectionState, allow_none=True, default_value=None
+    )
+    SSH_POSSIBLE_RESPONSES = [
+        "assword:",  # 0
```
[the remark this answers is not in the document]
unkcpz: The string pass to `expect` will be transfer to a compiled regex. Therefore I put `[Pp]assword:` here.

aiidalab_widgets_base/computational_resources.py
unkcpz: As we discussed offline, this needs to correspond to the index of `SSH_POSSIBLE_RESPONSES`.

aiidalab_widgets_base/computational_resources.py
unkcpz: Not true. The `enter_password` state (0) is not is the cases that will break the loop, so it will keep on asking for input. I think this also the main purpose of the 'while True' loop.

aiidalab_widgets_base/computational_resources.py
unkcpz: This is tested and yes, it will copy key to jump host as well.

aiidalab_widgets_base/computational_resources.py
unkcpz: I add a commend for `SshPossibleRepsonses` and convert idx return from `spawn.expect` to `Enum` class and compare it by identity.

unkcpz:
@csadorf thanks for the review. I have implemented most of the ideas as we discussed. Except below.

> could be significantly introduced by separating all business logic related functions (SSH-setup, computer-setup, code-setup) into separate modules. 

There are modules that contain legacy implementation of `ComputerSetup` and `CodeSetup` and will deprecate for next major release. 
As for decoupling to MVC model, I think it is not very easy. It requires lots of work to refactor properly since traitlets and object variable of view part entangled with backend model. Will leave it in the future.

aiidalab_widgets_base/computational_resources.py
csadorf: The `expect()` function waits on any of the responses in the `SshConnectionState` enum, that includes the password request. The loop is broken in two cases:
1. The `expect()` function times out.
2. We encountered one of the "terminal states" listed below the `try: ... except: ...` block.

This means that once we encountered the "enter_password" state, the loop will continue and wait for the next state. In case that the user does not enter the password within the timeout period (30s), the `pexpect.TIMEOUT` will be raised which triggers the disabling of the password field and a break out of the loop.

unkcpz: Yes, I see your point now. Let's keep it 30 s, I think it should be enough for inputting a password.

csadorf (changes requested):
Changes are good, however I found one more issue.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -512,35 +515,35 @@ def _send_password(self, _=None):
...
             )
             return
-        if self.ssh_connection_state == SshConnectionState.enter_password:
+        if self.ssh_connection_state is SshConnectionState.enter_password:
             self._handle_ssh_password()
-        elif self.ssh_connection_state == SshConnectionState.do_you_want_to_continue:
+        elif self.ssh_connection_state is SshConnectionState.do_you_want_to_continue:
             self._ssh_connection_process.sendline("yes")
 
         self.ssh_connection_state = SshConnectionState.waiting_for_input
```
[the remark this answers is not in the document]
unkcpz: Hi @csadorf, I think this `waiting_for_input` is needed to bookkeep a 'holding' state when setting password. 
If we don't have this modify of `ssh_connection_state` when I click `Setup ssh` twice the setting process will stuck, since the `ssh_connection_state` end up at `nter_password` state.

Since for my understanding `waiting_for_input` is waiting for password input, I move it to `ssh_handle`, and seems works fine.

csadorf (approved):
🚢  😃

unkcpz (approved):

unkcpz:
Thanks! merged. 🎉

PR #310, opened 2022-05-15
Add decorator to check whether operation allowed for editor

### LINKED ISSUE #309: Structure editor failed if operated without select atoms
Redirected from issue mentioned https://github.com/aiidalab/aiidalab-qe/issues/217

If not select atoms and process operations that involves using `ase.atom` will lead to `IndexError`. 

```
IndexError                                Traceback (most recent call last)
~/.local/lib/python3.7/site-packages/aiidalab_widgets_base/structures.py in translate_to_xyz(self, _)
   1059 
   1060         # The action.
-> 1061         geo_center = np.average(self.structure[self.selection].get_positions(), axis=0)
   1062         atoms.positions[self.selection] += self.str2vec(self.dxyz.value) - geo_center
   1063 

/opt/conda/lib/python3.7/site-packages/ase/atoms.py in __getitem__(self, i)
   1073         atoms.arrays = {}
   1074         for name, a in self.arrays.items():
-> 1075             atoms.arrays[name] = a[i].copy()
   1076 
   1077         atoms.constraints = conadd

IndexError: arrays used as indices must be of integer (or boolean) type
```

unkcpz:
Fixes #309 

The decorators are added to check editor is allowed to process the
following operation. If it is forbidden, a warning message will raised
to ask user to do the necessage operation rather than failed and raise
issue.

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -734,6 +734,30 @@ class SmilesWidget(ipw.VBox):
     def _default_structure(self):
         return None
 
+def valid_operator(operator):
+    """check if the operation is capabal to process.
+    If not valid the decorated method should not been execute
+    but pop up a warning message ask for doing the prerequisites"""
+    def inner(ref, *args, **kwargs):
+
+        if not ref.structure:
+            ref._operation_blocker_messages.value = """
+            <div class="alert alert-info">
+            <strong>Please choose structure to process the operation.</strong>
+            </div>
+            """
+            return
+        elif not ref.selection:
+            ref._operation_blocker_messages.value = """
+            <div class="alert alert-info">
+            <strong>Please choose atoms to process the operation.</strong>
+            </div>
+            """
+            return
+        else:
+            return operator(ref, *args, **kwargs)
+
+    return inner
 
 class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attributes
     """Widget that allows for the basic structure editing."""
@@ -741,6 +765,7 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
     structure = Instance(Atoms, allow_none=True)
     selection = List(Int)
     camera_orientation = List()
+    message = Unicode(allow_none=True)
 
     def __init__(self, title=""):
 
@@ -884,6 +909,8 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
             style={"description_width": "initial"},
         )
 
+        self._operation_blocker_messages = ipw.HTML()
+
         super().__init__(
             children=[
                 ipw.HTML(
@@ -965,6 +992,7 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
                     layout={"margin": "0px 0px 0px 20px"},
                 ),
                 ipw.HBox([btn_remove], layout={"margin": "0px 0px 0px 20px"}),
+                self._operation_blocker_messages,
             ]
         )
 
@@ -1007,6 +1035,7 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
         if self.autoclear_selection.value:
             self.selection = list()
 
+    @valid_operator
     def def_axis_p2(self, _=None):
         """Define the second point of axis."""
         com = (
@@ -1028,6 +1057,7 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
             )
             self.axis_p2.value = self.vec2str(versor.tolist())
 
+    @valid_operator
     def translate_dr(self, _=None):
         """Translate by dr along the selected vector."""
         atoms = self.structure.copy()
@@ -1041,6 +1071,7 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
         self.structure = atoms
         self.selection = selection
 
+    @valid_operator
     def translate_dxdydz(self, _=None):
         """Translate by the selected XYZ delta."""
         selection = self.selection
@@ -1052,6 +1083,7 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
         self.structure = atoms
         self.selection = selection
 
+    @valid_operator
     def translate_to_xyz(self, _=None):
         """Translate to the selected XYZ position."""
         selection = self.selection
@@ -1064,6 +1096,7 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
         self.structure = atoms
         self.selection = selection
 
+    @valid_operator
     def rotate(self, _=None):
         """Rotate atoms around selected point in space and vector."""
 
@@ -1080,6 +1113,7 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
         self.structure = atoms
         self.selection = selection
 
+    @valid_operator
     def mirror(self, _=None, norm=None, point=None):
         """Mirror atoms on the plane perpendicular to the action vector."""
 
@@ -1112,6 +1146,7 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
         self.structure = atoms
         self.selection = selection
 
+    @valid_operator
     def mirror_3p(self, _=None):
         """Mirror atoms on the plane containing action vector and action point."""
         pt1 = self.str2vec(self.axis_p2.value)
@@ -1121,6 +1156,7 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
         normal = normal / np.linalg.norm(normal)
         self.mirror(norm=normal, point=pt3)
 
+    @valid_operator
     def align(self, _=None):
         """Rotate atoms to align action vector with XYZ vector."""
 
@@ -1139,6 +1175,7 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
         self.structure = atoms
         self.selection = selection
 
+    @valid_operator
     def mod_element(self, _=None):
         """Modify selected atoms into the given element."""
         atoms = self.structure.copy()
@@ -1171,6 +1208,7 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
         self.structure = atoms
         self.selection = new_selection
 
+    @valid_operator
     def copy_sel(self, _=None):
         """Copy selected atoms and shift by 1.0 A along X-axis."""
         atoms = self.structure.copy()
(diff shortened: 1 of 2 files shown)


aiidalab_widgets_base/structures.py
```diff
@@ -735,12 +735,40 @@ def _default_structure(self):
         return None
 
 
+def valid_operator(operator):
+    """check if the operation is capabal to process.
```
yakutovicha, line 739:
```suggestion
    """Check if the operation can process.
```

aiidalab_widgets_base/structures.py
```diff
@@ -884,6 +912,8 @@ def disable_element(_=None):
             style={"description_width": "initial"},
         )
 
+        self._operation_blocker_messages = ipw.HTML()
```
yakutovicha, line 915: Maybe use `StatusHTML` instead.

https://github.com/aiidalab/aiidalab-widgets-base/blob/3378ccbe4458b4cb56d222d6acdc72f31d7298b9/aiidalab_widgets_base/utils/__init__.py#L141

aiidalab_widgets_base/structures.py
```diff
@@ -735,12 +735,40 @@ def _default_structure(self):
         return None
 
 
+def valid_operator(operator):
+    """check if the operation is capabal to process.
+    If not valid the decorated method should not been execute
+    but pop up a warning message ask for doing the prerequisites"""
```
yakutovicha, line 741:
```suggestion
    If not valid, the decorated method should not be executed but
    pop up a warning message asking for doing the prerequisites."""
```

aiidalab_widgets_base/structures.py
```diff
@@ -1112,6 +1149,7 @@ def mirror(self, _=None, norm=None, point=None):
         self.structure = atoms
         self.selection = selection
 
+    @valid_operator
```
yakutovicha, line 1152:
```suggestion
```

Here it is probably not needed, as `self.mirror` function is called further.

aiidalab_widgets_base/structures.py
```diff
@@ -735,12 +735,40 @@ def _default_structure(self):
...
+    """check if the operation is capabal to process.
+    If not valid the decorated method should not been execute
+    but pop up a warning message ask for doing the prerequisites"""
+
+    def inner(ref, *args, **kwargs):
+
+        if not ref.structure:
+            ref._operation_blocker_messages.value = """
+            <div class="alert alert-info">
+            <strong>Please choose structure to process the operation.</strong>
```
yakutovicha, line 748:
```suggestion
            <strong>Please choose a structure first.</strong>
```

aiidalab_widgets_base/structures.py
```diff
@@ -735,12 +735,40 @@ def _default_structure(self):
...
+            </div>
+            """
+            return
+        elif not ref.selection:
+            ref._operation_blocker_messages.value = """
+            <div class="alert alert-info">
+            <strong>Please choose atoms to process the operation.</strong>
```
yakutovicha, line 755:
```suggestion
            <strong>Please select atoms first.</strong>
```

yakutovicha (changes requested):
Thanks a lot @unkcpz!

I made the first pass and proposed some changes.

Another thing that I would like to take care here is the following. In many operators there is this pattern:

```python
        atoms = self.structure.copy()
        selection = self.selection
        ...
        self.structure = atoms
        self.selection = selection
```

Maybe we can also decorate this part of the code? As an option, we could name the decorator something like `@register_operator` and add an argument `update_structure`. When `update_structure` is True (default), then the code above is executed. If it is `False`, then only warning messages are added.

What do you think?

unkcpz:
> Maybe we can also decorate this part of the code? 

hi @yakutovicha, Please check `mirror` and decorator for my attempt for this implementation. It works (extra argument `update_structure` is very easy), but the tricky thing here is the entanglement of `structure` traitlets. 
In terms of `selection`, the tmp variable is needed since the viewer will reset it to empty, which is the only place related to the selection list. However, for `self.structure`, I have to create a `self.atoms` (or better to be called `self.tmp_structure`) to store the structure and restore after operation is done. It involves too much and I think it is the only way with the current code. Let me know if you are good we this implementation, I will continue along with this strategy.

yakutovicha (commented):
@unkcpz this is how I see it (very simplified):

```python
def register_operator(operator):
    def inner(ref, *argts, **kwargs):
        ref.structure, ref.selection = operator(ref, *args, **kwargs,
                                           atoms=ref.structure.copy(),
                                           selection=ref.selection.copy())
    return inner

...

    @register_operator
    def remove(self, _=None, atoms=None, selection=None):
        """Remove selected atoms."""
        del [atoms[selection]]
        return atoms, selection
```

unkcpz:
@yakutovicha thanks! Yes, it is much more simplified, I didn't think about that return value from the decorator.
But it seems `StatusHTML` has some issues. When it disappears, it can not be triggered and show the message again. Do you have any idea why?

aiidalab_widgets_base/structures.py
```diff
@@ -735,6 +736,45 @@ def _default_structure(self):
         return None
 
 
+def register_operator(operator):
```
csadorf, line 739:
```suggestion
def _register_operator(operator):
```

I would recommend to make this explicitly private.

aiidalab_widgets_base/utils/__init__.py
```diff
@@ -128,13 +129,13 @@ def _clear_value(self):
         else:
             self.value = ""
 
-    def show_temporary_message(self, value, clear_after=3):
+    def show_temporary_message(self, value):
```
csadorf, line 132: ⚠️ This changes the API of this class!

If we really need the class variable, why not use this as a default here?

aiidalab_widgets_base/structures.py
```diff
@@ -1028,73 +1078,58 @@ def def_perpendicular_to_screen(self, _=None):
...
-
-        selection = self.selection
 
         atoms.positions[self.selection] += np.array(
             self.action_vector * self.displacement.value
         )
 
-        self.structure = atoms
-        self.selection = selection
+        return atoms, selection
```
csadorf, line 1089: I am worried that the magic introduced with `@register_operator` might be obfuscating the actual behavior here too much. I understand having a guard against operating on empty sets, etc., but do we really have to also manipulate the state in the hidden wrapper function?

I think it would be much better to make explicit that the class state is changed here:
```suggestion
        self.atoms, self.selection = atoms, selection
```
If there is need for additional, checks, you could provide some utility function if needed:
```suggestion
        self.atoms, self.selection = self._update_atoms_selection(atoms, selection)
```

Alternatively, you could make the operator function truly stateless (either as `@staticmethod` or `@classmethod`) and then pass in the input, and return the output state.

csadorf (changes requested):

aiidalab_widgets_base/structures.py
yakutovicha: I would prefer to keep it public as I would use it in the other editors.

aiidalab_widgets_base/utils/__init__.py
unkcpz: Since I want to set this value when I create an instance of `StatusHTML`. Otherwise how I can set the timer to rather than 3 secs?

unkcpz: > You can set the timer when you call that function.

I think the problem is, that `show_temporary_message` is called in the callback function. 

```
    @traitlets.observe("message")
    def _observe_message(self, change):
        self.show_temporary_message(change["new"])
```

csadorf: > Otherwise, why not follow my suggestion and use the class variable as a default value for the function argument?

What about my other suggestion?

unkcpz: As the class variable still not have a chance to set it when creating the instance?

unkcpz: @yakutovicha Could you look at the use of the `StatusHTML` widget here and let me know how to fix the problem that the message does not show again?

yakutovicha: Yep, this is done in c9214504b4e80a7a8b4d346bc5d695e08ea14520. Also, I made some changes in the `StatusHTML` widget itself, but those will be incorporated in the other PR.

yakutovicha: Not entirely. Please see other changes (including yours) here: https://github.com/aiidalab/aiidalab-widgets-base/pull/313

csadorf:
@unkcpz Please re-request my review once this is ready to merge. Thank you.

aiidalab_widgets_base/structures.py
```diff
@@ -735,6 +738,40 @@ def _default_structure(self):
...
+        if not ref.structure:
+            ref._status_message.message = """
+            <div class="alert alert-info">
+            <strong>Please choose a structure first.</strong>
+            </div>
+            """
+        elif not ref.selection:
+            ref._status_message.message = """
+            <div class="alert alert-info">
+            <strong>Please select atoms first.</strong>
```
csadorf, line 760:
```suggestion
            <strong>Please make a selection first.</strong>
```
They don't necessarily have to be atoms, do they?

aiidalab_widgets_base/structures.py
```diff
@@ -1009,14 +1049,21 @@ def def_axis_p1(self, _=None):
 
     def def_axis_p2(self, _=None):
```
csadorf, line 1050: Why is the first argument always ignored? Could we just drop that as part of our operator decorator?

aiidalab_widgets_base/utils/__init__.py
yakutovicha: addressed in #313

aiidalab_widgets_base/structures.py
yakutovicha: > Why is the first argument always ignored?

It is because the button widgets provide this argument when calling the function. It has no useful information, so we ignore it.

> Could we just drop that as part of our operator decorator?

I think the answer depends a bit on how we design the decorator/decorators. So we should probably answer [this](https://github.com/aiidalab/aiidalab-widgets-base/pull/310#discussion_r884730170) one first.

unkcpz:
The docs build test failed, it was all fine. any idea how to fix it?

aiidalab_widgets_base/structures.py
csadorf: My point is that we could potentially have "structures" that are not made up of "atoms", e.g., molecular nano structures. So by avoiding that term, it's just as correct, but less restrictive.

unkcpz: > My point is that we could potentially have "structures" that are not made up of "atoms", e.g., molecular nano structures

But do we support these kind structures in this widget?

csadorf: I withdraw my suggestion.

aiidalab_widgets_base/structures.py
csadorf: Please move forward as you see fit.

unkcpz:
Hi @csadorf, can this be merged? Anything else I can add?

csadorf (approved):

csadorf:
> Hi @csadorf, can this be merged? Anything else I can add?

Please go ahead! 😃  Nice work! 👍

unkcpz:
@csadorf thanks! It needs @yakutovicha to approve to get it merged?

csadorf:
> @csadorf thanks! It needs @yakutovicha to approve to get it merged?

I'd recommend that.

yakutovicha (approved):

yakutovicha:
No objections from my side, please go ahead @unkcpz.

Thanks a lot!

PR #313, opened 2022-05-24
Work on the `StatusHTML` widget.

yakutovicha:
- Introduce a `new_line` attribute that allows modifying the new line
symbol depending on the string type.
- Reset the `message` trait every time its value has been added to the
stack. This allows re-displaying the warning message even if the text
hasn't changed.
- Set the `clear_after` attribute at the moment of object construction.

### DIFF
--- a/aiidalab_widgets_base/utils/__init__.py
+++ b/aiidalab_widgets_base/utils/__init__.py
@@ -131,8 +131,8 @@ class _StatusWidgetMixin(traitlets.HasTraits):
             self.value = ""
 
     def show_temporary_message(self, value, clear_after=None):
-        clear_after = clear_after or self._clear_after
         """Show a temporary message and clear it after the given interval."""
+        clear_after = clear_after or self._clear_after
         if value:
             self._message_stack.append(value)
             self.value = self.new_line.join(self._message_stack)
--- a/aiidalab_widgets_base/utils/__init__.py
+++ b/aiidalab_widgets_base/utils/__init__.py
@@ -140,7 +140,7 @@ class _StatusWidgetMixin(traitlets.HasTraits):
             # Start new timer that will clear the value after the specified interval.
             self._clear_timer = threading.Timer(self._clear_after, self._clear_value)
             self._clear_timer.start()
-            self.message = ""
+            self.message = None
 
 
 class StatusHTML(_StatusWidgetMixin, ipw.HTML):
--- a/aiidalab_widgets_base/utils/__init__.py
+++ b/aiidalab_widgets_base/utils/__init__.py
@@ -130,7 +130,8 @@ class _StatusWidgetMixin(traitlets.HasTraits):
         else:
             self.value = ""
 
-    def show_temporary_message(self, value):
+    def show_temporary_message(self, value, clear_after=None):
+        clear_after = clear_after or self._clear_after
         """Show a temporary message and clear it after the given interval."""
         if value:
             self._message_stack.append(value)
--- a/aiidalab_widgets_base/utils/__init__.py
+++ b/aiidalab_widgets_base/utils/__init__.py
@@ -114,9 +114,11 @@ class _StatusWidgetMixin(traitlets.HasTraits):
     """
 
     message = traitlets.Unicode(default_value="", allow_none=True)
+    new_line = "\n"
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, clear_after=3, *args, **kwargs):
         self._clear_timer = None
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/utils/__init__.py
```diff
@@ -114,33 +114,39 @@ class _StatusWidgetMixin(traitlets.HasTraits):
...
+            self._message_stack.append(value)
+            self.value = self.new_line.join(self._message_stack)
 
-        # Start new timer that will clear the value after the specified interval.
-        self._clear_timer = threading.Timer(clear_after, self._clear_value)
-        self._clear_timer.start()
+            # Start new timer that will clear the value after the specified interval.
+            self._clear_timer = threading.Timer(self._clear_after, self._clear_value)
+            self._clear_timer.start()
+            self.message = ""
```
yakutovicha, line 142: I believe @csadorf won't like this, but that is the only way I could find to let the widget accept the same warning message.

aiidalab_widgets_base/utils/__init__.py
```diff
@@ -114,33 +114,39 @@ class _StatusWidgetMixin(traitlets.HasTraits):
...
         """Set widget .value to be an empty string."""
         if self._message_stack:
             self._message_stack.pop(0)
-            self.value = "<br>".join(self._message_stack)
+            self.value = self.new_line.join(self._message_stack)
         else:
             self.value = ""
 
-    def show_temporary_message(self, value, clear_after=3):
+    def show_temporary_message(self, value):
```
csadorf, line 133: Let's please avoid the incompatible change and use the class variable as default value (as suggested multiple times before):
```suggestion
    def show_temporary_message(self, value, clear_after=None):
        clear_after = clear_after or self._clear_after
```

aiidalab_widgets_base/utils/__init__.py
csadorf: Indeed, I suspect that this could lead to some weird behavior since this function is triggered as part of the trait callback and thus would trigger the callback directly again. Maybe we can clear it together with the `.value` trait instead?

csadorf:
> And another question, do we really need both `self.message` and `self.value`? If I understand correctly, `value` is for set the instant message, and message is to keep it and show?

This class is supposed to be used as a mixin, where the `.message` trait is the one that we use to trigger showing a temporary message, and the `.value` trait is from the parent class that actually displays the message. Using traits like this is a bit ugly, which might explain some of the oddities that @yakutovicha is seeing.

aiidalab_widgets_base/utils/__init__.py
```diff
@@ -130,7 +130,8 @@ def _clear_value(self):
         else:
             self.value = ""
 
-    def show_temporary_message(self, value):
+    def show_temporary_message(self, value, clear_after=None):
+        clear_after = clear_after or self._clear_after
         """Show a temporary message and clear it after the given interval."""
```
csadorf, line 135: You need to switch line 134 and 135.

yakutovicha: sure.

aiidalab_widgets_base/utils/__init__.py
yakutovicha: I thought a bit about this.

> Maybe we can clear it together with the `.value` trait instead?

I am not sure what benefits it gives. Can you name any? It looks to me that the solution you propose makes a delayed trait reset. The mechanism that prevents `self.show_temporary_message()` from doing anything remains the same.

At the same time, we already have some code examples that modify the trait in the observer function and so far we didn't run into issues with such an approach.

My preferred solution would be to change the trait value but not notify the observing functions. I looked for a while and didn't find ways to temporary block the notifications. Maybe there are good reasons for it to not exist.

csadorf: The purpose of the `message` trait is to be able to trigger showing a temporary message via a trait _change_. That's a bit of a weird mechanism anyways, and maybe the `.message` trait should not exist in the first place. My main issue is that `.show_temporary_message()` both _observes and changes_ the `.message` trait. That's typically a red flag to me and could lead infinite loops. However, like you said, we are also clearing it in the `._clear_value()` function, just on a delay and we are protecting this function with the truthy check on line 136, so should be fine.

The one suggestion I have is to maybe use `None` as the explicit _null_ value since that's allowed:
```suggestion
            self.message = None
```

csadorf (approved):

csadorf (approved):

PR #315, opened 2022-06-01
Bumpver: optimade client to 2022.4.20

unkcpz:
This will partially solve the issue of https://github.com/aiidalab/aiidalab-qe/issues/188 and this is the version that used for current materials cloud optimadeclient.


unkcpz:
@csadorf yes, I run install and test on my local machine. This library is also the version used in materials cloud so should be safe. Only a warning show in notebook mode:
```
/home/aiida/.local/lib/python3.7/site-packages/optimade/__init__.py:18: DeprecationWarning: v0.17 of the `optimade` package will be the last to support Python 3.7. Please upgrade to Python 3.8+ to use v0.18 and later versions of `optimade`.
  DeprecationWarning,
```

csadorf (approved):

PR #320, opened 2022-06-22
Fix: Structures from SmilesWidget should have PBC=false

### LINKED ISSUE #289: SmilesWidget: PBC should be set to False

danielhollas:
Extracted from #290 so that it could be merged.

Closes #289


yakutovicha (approved):
LGTM!

PR #321, opened 2022-06-23
SmilesWidget: Add basic error handling

danielhollas:
Extracted from #290 as a more conservative subset of the changes there.

I am also proposing some small style changes. Moving the Generate button next to the input field since there's plenty of space, and the space below can be used for notification area. I also changed the button style to 'primary' since it _is_ the primary button of this plugin (and I like the blue color better :-)). Feel free to push back on that though as it might not be consistent with the colors in other parts of the widgets. I have also added tooltips to hopefully improve accessibility.

I was also wondering whether we should be adding classes to some of the key UI elements, e.g.

```python
self.create_structure_btn.add_class("btn-generate-from-smiles")
```
This would for example make them easier targets in selenium tests.

BEFORE:
image

AFTER:
image

## Testing

1.  
input1: `c1cc1`
stderr: `RDKit ERROR: [19:58:16] Can't kekulize mol.  Unkekulized atoms: 0 1 2` 
input2: `Ch`
stderr: `RDKit ERROR: [20:01:52] SMILES Parse Error: syntax error while parsing: Ch
RDKit ERROR: [20:01:52] SMILES Parse Error: Failed parsing SMILES 'Ch' for input: 'Ch'`

Both of these would result in:
```
ArgumentError: Python argument types in
    rdkit.Chem.rdmolops.AddHs(NoneType)
did not match C++ signature:
    AddHs(RDKit::ROMol mol, bool explicitOnly=False, bool addCoords=False, boost::python::api::object onlyOnAtoms=None, bool addResidueInfo=False)
```
With this PR, we just print "ERROR: Invalid SMILES code"

2.  input: `CS(C)(C)=O`
This is actually a valid SMILES string, as discussed in https://github.com/rdkit/rdkit/issues/200. However, the UFF force field that we use for minimization is missing parameters for sulphur in this particular hybridization, so the rdkit minimization fails with
```
~/apps/aiidalab-widgets-base/aiidalab_widgets_base/structures.py in _rdkit_opt(self, smiles, steps)
    706 
    707         AllChem.EmbedMolecule(mol, maxAttempts=20, randomSeed=42)
--> 708         AllChem.UFFOptimizeMolecule(mol, maxIters=steps)
    709         positions = mol.GetConformer().GetPositions()
    710         natoms = mol.GetNumAtoms()

RuntimeError: Pre-condition Violation
	bad params pointer
	Violation occurred on line 75 in file Code/ForceField/UFF/AngleBend.cpp
	Failed Expression: at2Params
	RDKIT: 2020.09.1
	BOOST: 1_74
```
In the new code, if there are missing UFF parameters we skip the UFF optimization. 

3. input: `CC(C)(C)*OO`
For some reason, this one passes through rdkit even though it contains an invalid character `*`, and fails later in ASE, as described in #270. This PR does not solve this problem. We should probably treat this as an RDKit bug and open an isssue there.

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -705,9 +705,18 @@ class SmilesWidget(ipw.VBox):
 
         smiles = smiles.replace("[", "").replace("]", "")
         mol = Chem.MolFromSmiles(smiles)
+        if mol is None:
+            # Something is seriously wrong with the SMILES code,
+            # just return None and don't attempt anything else.
+            self.output.value = "ERROR: Invalid SMILES string"
+            return None
         mol = Chem.AddHs(mol)
 
         AllChem.EmbedMolecule(mol, maxAttempts=20, randomSeed=42)
+        if not AllChem.UFFHasAllMoleculeParams(mol):
+            msg = "WARNING: Missing UFF parameters"
+            raise ValueError(msg)
+
         AllChem.UFFOptimizeMolecule(mol, maxIters=steps)
         positions = mol.GetConformer().GetPositions()
         natoms = mol.GetNumAtoms()
@@ -718,7 +727,8 @@ class SmilesWidget(ipw.VBox):
         """Convert SMILES to ase structure try rdkit then pybel"""
         try:
             return self._rdkit_opt(smiles, steps)
-        except ValueError:
+        except ValueError as e:
+            self.output.value = str(e)
             return self._pybel_opt(smiles, steps)
 
     def _on_button_pressed(self, change):  # pylint: disable=unused-argument
@@ -727,11 +737,14 @@ class SmilesWidget(ipw.VBox):
 
         if not self.smiles.value:
             return
-        self.output.value = "Screening possible conformers {}".format(
+        spiner = "Screening possible conformers {}".format(
             self.SPINNER
         )  # font-size:20em;
+        self.output.value = spiner
         self.structure = self._mol_from_smiles(self.smiles.value)
-        self.output.value = ""
+        # Don't overwrite possible error/warning messages
+        if self.output.value == spiner:
+            self.output.value = ""
 
     @default("structure")
     def _default_structure(self):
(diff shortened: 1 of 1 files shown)


csadorf:
@danielhollas Please let us know when this is ready for review. Just as a note, I won't have time to review this today since I am not familiar enough with the particular widget and @yakutovicha is not available until next week. So it will probably have to wait until early next week if that's alright.

danielhollas:
@yakutovicha you can take a look, thanks! I only kept this as a draft to do a bit more testing.

aiidalab_widgets_base/structures.py
```diff
@@ -703,22 +703,31 @@ def _rdkit_opt(self, smiles, steps):
...
+        if mol is None:
+            # Something is seriously wrong with the SMILES code,
+            # just return None and don't attempt anything else.
+            self.output.value = "RDKit ERROR: Invalid SMILES string"
+            return None
         mol = Chem.AddHs(mol)
 
         AllChem.EmbedMolecule(mol, maxAttempts=20, randomSeed=42)
+        if not AllChem.UFFHasAllMoleculeParams(mol):
+            raise ValueError("RDKit ERROR: Missing UFF parameters")
```
danielhollas, line 716: Alternatively, we could just skip the UFF optimization in this case.

danielhollas:
@yakutovicha this is now ready for review. Thanks!

yakutovicha:
> @yakutovicha this is now ready for review. Thanks!

Hey, @danielhollas sorry for being slow in replying. This week is a bit crazy for me. I will have a look as soon as I can.

yakutovicha (approved):
Thanks, @danielhollas, code looks good also things work as expected.

I had some fun playing with the [examples from Wikipedia](https://en.wikipedia.org/wiki/Simplified_molecular-input_line-entry_system#Examples)

danielhollas:
@yakutovicha awesome, thank you! :heart: Thanks for the link, I somehow missed that table. I have an early version of the SmilesWidget that does multiple conformer sampling, will have to try it out on these as well. :-)

PR #324, opened 2022-07-06
Fixed bug in setting private key

daniel-sintef:
name is an attribute not a method


yakutovicha (approved):
good catch, thanks a lot @daniel-sintef!

PR #325, opened 2022-07-11
Update optimade database provider list

unkcpz:
It has to be updated here for optimade database providers if we don't want to set the list explicitly on QeApp. ~~I think the best solution is just to have the default list set on `optimade-client` as const variable and use it for both here and materials cloud optimade client.~~ I make PR https://github.com/CasperWA/voila-optimade-client/pull/449 to voila-optimade-client tool for the list. This PR relies on it.

### DIFF
--- a/aiidalab_widgets_base/databases.py
+++ b/aiidalab_widgets_base/databases.py
@@ -144,7 +144,7 @@ class OptimadeQueryWidget(ipw.VBox):
 
     :param embedded: Whether or not to show extra database and provider information.
         When set to `True`, the extra information will be hidden, this is useful
-        in situations where the widget is used in a Tab or similar, e.g., for the 
+        in situations where the widget is used in a Tab or similar, e.g., for the
         class :class:`aiidalab_widgets_base.structures.StructureManagerWidget`.
     :type embedded: bool
     :param title: Title used for Tab header if employed in
--- a/aiidalab_widgets_base/databases.py
+++ b/aiidalab_widgets_base/databases.py
@@ -144,8 +144,8 @@ class OptimadeQueryWidget(ipw.VBox):
 
     :param embedded: Whether or not to show extra database and provider information.
         When set to `True`, the extra information will be hidden, this is useful
-        in situations where the widget is used in a Tab or similar, e.g., can be used
-        for the :class:`aiidalab_widgets_base.structures.StructureManagerWidget`.
+        in situations where the widget is used in a Tab or similar, e.g., for the 
+        class :class:`aiidalab_widgets_base.structures.StructureManagerWidget`.
     :type embedded: bool
     :param title: Title used for Tab header if employed in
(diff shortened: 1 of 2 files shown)


unkcpz:
New optimade-client is released with the list const variable accessible, this PR is ready to go.

aiidalab_widgets_base/databases.py
```diff
@@ -190,6 +173,7 @@ def __init__(
             database_limit=kwargs.pop("database_limit", None),
             disable_providers=kwargs.pop("disable_providers", self._disable_providers),
             skip_databases=kwargs.pop("skip_databases", self._skip_databases),
+            skip_provider=kwargs.pop("skip_databases", self._skip_providers),
```
CasperWA, line 176:
```suggestion
            skip_providers=kwargs.pop("skip_databases", self._skip_providers),
```

aiidalab_widgets_base/databases.py
```diff
@@ -138,8 +144,8 @@ class OptimadeQueryWidget(ipw.VBox):
 
     :param embedded: Whether or not to show extra database and provider information.
         When set to `True`, the extra information will be hidden, this is useful
-        in situations where the widget is used in a Tab or similar, e.g., for the
-        :class:`aiidalab_widgets_base.structures.StructureManagerWidget`.
+        in situations where the widget is used in a Tab or similar, e.g., can be used
+        for the :class:`aiidalab_widgets_base.structures.StructureManagerWidget`.
```
CasperWA, line 148:
```suggestion
        in situations where the widget is used in a Tab or similar, e.g., for the
        class :class:`aiidalab_widgets_base.structures.StructureManagerWidget`.
```

CasperWA (changes requested):

aiidalab_widgets_base/databases.py
unkcpz: Thanks! changed.

setup.cfg
```diff
@@ -33,7 +33,7 @@ install_requires =
     nbconvert<6
     nglview~=3.0
     numpy~=1.17
-    optimade-client==2022.4.20
+    optimade-client~=2022.9.19
```
CasperWA, line 36: Note, this can be dangerous, as the versioning of `optimade-client` is not following semantic versioning, but rather, calendar versioning. Hence, major changes may occur from one version to another, without a way to control that through this version number.

I think I would keep the `==` operator here, instead of using the `~=` operator, and simply update the dependency whenever a new release occurs for `optimade-client`, making sure it doesn't break stuff.

yakutovicha (commented):
I would suggest that @CasperWA approves the PR. I am not an expert here, but it looks good to me.

setup.cfg
unkcpz: Thanks! Pinning the version makes sense. Changed.

CasperWA (approved):

PR #327, opened 2022-07-12
convert to primitive cell or to conventional cell

unkcpz:
Add to structure editor the cell convertor from or to primitive/conventional cell.

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -847,6 +847,7 @@ def _register_operator(operator):
             )
 
     return inner
+    
 
 
 class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attributes
@@ -859,6 +860,12 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
     def __init__(self, title=""):
 
         self.title = title
+        # cell transfor opration widget
+        primitive_cell = ipw.Button(
+            description="Convert to primitive cell", layout={"width": "initial"},
+        )
+        primitive_cell.on_click(self.def_primitive_cell)
+
         # Define action vector.
         self.axis_p1 = ipw.Text(
             description="Starting point", value="0 0 0", layout={"width": "initial"}
@@ -1002,6 +1009,14 @@ class BasicStructureEditor(ipw.VBox):  # pylint: disable=too-many-instance-attri
 
         super().__init__(
             children=[
+                ipw.HTML(
+                    "<b>Cell transformation:</b>",
+                    layout={"margin": "20px 0px 10px 0px"},
+                ),
+                ipw.HBox([
+                        primitive_cell,
+                    ],
+                ),
                 ipw.HTML(
                     "<b>Action vector and point:</b>",
                     layout={"margin": "20px 0px 10px 0px"},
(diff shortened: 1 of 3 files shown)


unkcpz:
Hi @yakutovicha, can you have a look at this PR?

unkcpz:
Tested on QeApp.

aiidalab_widgets_base/structures.py
```diff
@@ -812,7 +775,7 @@ def _default_structure(self):
         return None
 
 
-def _register_operator(operator):
+def _register_structure(operator):
```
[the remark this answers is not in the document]
unkcpz: True, I always have this bad habit to add something else 'by the way' without a clear independent PR. Will be careful in the future.

unkcpz:
@yakutovicha I think this PR is ready.

aiidalab_widgets_base/structures.py
```diff
@@ -842,15 +869,88 @@ def inner(ref, *args, **kwargs):
...
+        self.structure = atoms
+
+    @_register_structure
+    def def_conventional_cell(self, _=None, atoms=None):
+        atoms = self._to_standardize_cell(atoms, to_primitive=False)
+
+        self.structure = atoms
+
+    @staticmethod
+    def _to_standardize_cell(
```
yakutovicha, line 927: I think it should be called `to_standard_cell` or `stardardize_cell`.

aiidalab_widgets_base/structures.py
```diff
@@ -842,15 +869,88 @@ def inner(ref, *args, **kwargs):
...
+            layout={"width": "initial"},
+        )
+        conventional_cell.on_click(self.def_conventional_cell)
+
+        super().__init__(
+            children=[
+                ipw.HTML(
+                    "<b>Cell transformation:</b>",
+                    layout={"margin": "20px 0px 10px 0px"},
+                ),
```
yakutovicha, line 904: I would argue this is a bit redundant. "Cell transformation" can also be set as a title.

yakutovicha (changes requested):
Hi @unkcpz. I tested it, looks good. Just a few minor suggestions.

superstar54:
`Cell Transformation` is a more general concept. If we use this term, it is better to include the general transformation matrix. Please have a look at `make_supercell` method here. https://wiki.fysik.dtu.dk/ase/ase/build/tools.html#ase.build.make_supercell
Also mentioned in #364

aiidalab_widgets_base/structures.py
unkcpz: Fair point, I use `to_standard_cell`.

aiidalab_widgets_base/structures.py
unkcpz: That's true, I use "Cell transform" for the default tab title and remove this HTML widget title.

unkcpz:
@superstar54 Thanks for the advice. It is a very good suggestion, but I think it is better to be implemented it in a separate PR. Because it involves some design choice such as what we expect as the interface to input the transformation matrix that is worth having more discussion. I convert the item in your issue to a separate issue #365.

aiidalab_widgets_base/structures.py
yakutovicha: I think the function is now called `to_standardize_cell`. I am confused :)

unkcpz: Ah, my bad. Can not focus on the so early train. I change it to `to_standard_cell`.

yakutovicha (approved):
LGTM!

PR #328, opened 2022-07-12
Parse symmetry info and show in cell tab

unkcpz:
- [x] ~~Move the cell tab to front~~ Can be set when create widget.
- [x] parse symmetry info using `spglib`
- [x] ~~Using `ase.Atom` instead of `ase.Cell` in `StructureDataViewer` as active traitslet talk to other widget.~~
- [x] Missing unit symbol 'Angstrom'
- [x] ~~drop `ase.Cell` for structure viewer, use only `ase.Atoms` as trait.~~
- [x] When atom selected, automatically change to 'selection' tab.

### DIFF
--- a/aiidalab_widgets_base/utils/__init__.py
+++ b/aiidalab_widgets_base/utils/__init__.py
@@ -164,10 +164,10 @@ class StatusHTML(_StatusWidgetMixin, ipw.HTML):
         self.show_temporary_message(change["new"])
 
 
-def ase2spglib(cell: Atoms):
+def ase2spglib(ase_structure: Atoms):
     """convert ase Atoms instance to spglib cell"""
-    lattice = cell.get_cell()
-    positions = cell.get_scaled_positions()
-    numbers = cell.get_atomic_numbers()
+    lattice = ase_structure.get_cell()
+    positions = ase_structure.get_scaled_positions()
+    numbers = ase_structure.get_atomic_numbers()
 
     return (lattice, positions, numbers)
--- a/aiidalab_widgets_base/utils/__init__.py
+++ b/aiidalab_widgets_base/utils/__init__.py
@@ -5,12 +5,8 @@ import ipywidgets as ipw
 import more_itertools as mit
 import numpy as np
 import traitlets
-from aiida.plugins import DataFactory
+from ase import Atoms
 from ase.io import read
-
-CifData = DataFactory("cif")  # pylint: disable=invalid-name
-StructureData = DataFactory("structure")  # pylint: disable=invalid-name
-TrajectoryData = DataFactory("array.trajectory")  # pylint: disable=invalid-name
 
 
 def valid_arguments(arguments, valid_args):
@@ -166,3 +162,12 @@ class StatusHTML(_StatusWidgetMixin, ipw.HTML):
     @traitlets.observe("message")
     def _observe_message(self, change):
         self.show_temporary_message(change["new"])
+
+
+def ase2spglib(cell: Atoms):
+    """convert ase Atoms instance to spglib cell"""
+    lattice = cell.get_cell()
+    positions = cell.get_scaled_positions()
+    numbers = cell.get_atomic_numbers()
+
+    return (lattice, positions, numbers)
--- a/aiidalab_widgets_base/utils/__init__.py
+++ b/aiidalab_widgets_base/utils/__init__.py
@@ -6,8 +6,8 @@ import ipywidgets as ipw
 import more_itertools as mit
 import numpy as np
 import traitlets
-from ase import Atoms
 from aiida.plugins import DataFactory
+from ase import Atoms
 from ase.io import read
 
 CifData = DataFactory("cif")  # pylint: disable=invalid-name
--- a/aiidalab_widgets_base/utils/__init__.py
+++ b/aiidalab_widgets_base/utils/__init__.py
@@ -1,5 +1,6 @@
 """Some utility functions used acrross the repository."""
 import threading
+from typing import Any, Tuple
 
 import ipywidgets as ipw
 import more_itertools as mit
@@ -164,8 +165,11 @@ class StatusHTML(_StatusWidgetMixin, ipw.HTML):
         self.show_temporary_message(change["new"])
 
 
-def ase2spglib(ase_structure: Atoms):
-    """convert ase Atoms instance to spglib cell"""
+def ase2spglib(ase_structure: Atoms) -> Tuple[Any, Any, Any]:
+    """
+    Convert ase Atoms instance to spglib cell in the format defined at
+    https://spglib.github.io/spglib/python-spglib.html#crystal-structure-cell
+    """
     lattice = ase_structure.get_cell()
     positions = ase_structure.get_scaled_positions()
     numbers = ase_structure.get_atomic_numbers()
--- a/aiidalab_widgets_base/utils/__init__.py
+++ b/aiidalab_widgets_base/utils/__init__.py
@@ -7,7 +7,12 @@ import more_itertools as mit
 import numpy as np
 import traitlets
 from ase import Atoms
+from aiida.plugins import DataFactory
 from ase.io import read
+
+CifData = DataFactory("cif")  # pylint: disable=invalid-name
+StructureData = DataFactory("structure")  # pylint: disable=invalid-name
+TrajectoryData = DataFactory("array.trajectory")  # pylint: disable=invalid-name
 
 
 def valid_arguments(arguments, valid_args):
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -14,8 +14,7 @@ import traitlets
 from aiida.cmdline.utils.common import get_workchain_report
 from aiida.orm import Node
 from ase import Atoms, neighborlist
-
-# from ase.cell import Cell
+from ase.cell import Cell
 from IPython.display import clear_output, display
 from matplotlib.colors import to_rgb
 from numpy.linalg import norm
@@ -157,7 +156,7 @@ class DictViewer(ipw.VBox):
 
 
 class _StructureDataBaseViewer(ipw.VBox):
-    """Base viewer class for AiiDA structure or trajectory objects.
+    """Base viewer class for AiiDA structure or ase structure objects.
 
     :param configure_view: If True, add configuration tabs (deprecated)
     :type configure_view: bool
@@ -171,7 +170,8 @@ class _StructureDataBaseViewer(ipw.VBox):
     selection = List(Int)
     selection_adv = Unicode()
     supercell = List(Int)
-    cell = Instance(Atoms, allow_none=True)
+    cell = Instance(Cell, allow_none=True)  # Will be deprecated
+    ase_structure = Instance(Atoms, allow_none=True)
     DEFAULT_SELECTION_OPACITY = 0.2
     DEFAULT_SELECTION_RADIUS = 6
     DEFAULT_SELECTION_COLOR = "green"
@@ -229,9 +229,10 @@ class _StructureDataBaseViewer(ipw.VBox):
 
         super().__init__(children, **kwargs)
 
-    @observe("cell")
-    def _observe_cell(self, _=None):
-        if self.cell:
+    @observe("ase_structure")
+    def _observe_ase_structure(self, _=None):
+        if self.ase_structure:
+            self.cell = self.ase_structure.cell
             self.cell_a.value = "<i><b>a</b></i>: {:.4f} {:.4f} {:.4f}".format(
                 *self.cell.array[0]
             )
@@ -256,9 +257,9 @@ class _StructureDataBaseViewer(ipw.VBox):
             self.cell_beta.value = f"&beta;: {self.cell.angles()[1]:.4f}"
             self.cell_gamma.value = f"&gamma;: {self.cell.angles()[2]:.4f}"
 
-            spglib_cell = ase2spglib(self.cell)
+            spglib_structure = ase2spglib(self.ase_structure)
             symmetry_dataset = spglib.get_symmetry_dataset(
-                spglib_cell, symprec=1e-5, angle_tolerance=1.0
+                spglib_structure, symprec=1e-5, angle_tolerance=1.0
             )
 
             self.cell_spacegroup.value = f"Spacegroup: {symmetry_dataset['international']} (No.{symmetry_dataset['number']})"
@@ -276,6 +277,9 @@ class _StructureDataBaseViewer(ipw.VBox):
             self.cell_beta.value = "&beta;:"
             self.cell_gamma.value = "&gamma;:"
 
+            self.cell_spacegroup.value = "Spacegroup: "
+            self.cell_hall.value = "Hall: "
+
     def _cell_tab(self):
         self.cell_a = ipw.HTML()
         self.cell_b = ipw.HTML()
@@ -292,7 +296,7 @@ class _StructureDataBaseViewer(ipw.VBox):
         self.cell_spacegroup = ipw.HTML()
         self.cell_hall = ipw.HTML()
 
-        self._observe_cell()
+        self._observe_ase_structure()
 
         return ipw.VBox(
             [
@@ -332,8 +336,8 @@ class _StructureDataBaseViewer(ipw.VBox):
                                 ipw.HTML("Symmetry infomation:"),
                                 self.cell_spacegroup,
                                 self.cell_hall,
-                                self.cell_gamma,
-                            ]
+                            ],
+                            layout={"margin": "0 0 0 50px"},
                         ),
                     ]
                 ),
@@ -784,10 +788,10 @@ class StructureDataViewer(_StructureDataBaseViewer):
         # Remove the current structure(s) from the viewer.
         if change["new"] is not None:
             self.set_trait("displayed_structure", change["new"].repeat(self.supercell))
-            self.set_trait("cell", change["new"].cell)
+            self.set_trait("ase_structure", change["new"])
         else:
             self.set_trait("displayed_structure", None)
-            self.set_trait("cell", None)
+            self.set_trait("ase_structure", None)
 
     @observe("displayed_structure")
     def _update_structure_viewer(self, change):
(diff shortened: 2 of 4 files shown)


unkcpz:
fixes https://github.com/aiidalab/aiidalab-qe/issues/244

yakutovicha:
@unkcpz I see there are still some non-completed tasks for this PR. Whenever this is ready please assign relevant people for the review.

unkcpz:
There is still one thing probably good to have. When the structure is not chosen or uploaded the `StructureDataViewer` is hidden. If you all agree with this, since I am on this part of code, I can quickly add this feature to this PR.

unkcpz:
The result of the widget after these change would be:
image

csadorf:
@unkcpz Please re-request my review when this is ready, thank you!

unkcpz:
@csadorf Yes, it is ready to review.

aiidalab_widgets_base/utils/__init__.py
```diff
@@ -153,3 +154,12 @@ class StatusHTML(_StatusWidgetMixin, ipw.HTML):
     @traitlets.observe("message")
     def _observe_message(self, change):
         self.show_temporary_message(change["new"])
+
+
+def ase2spglib(ase_structure: Atoms):
+    """convert ase Atoms instance to spglib cell"""
```
csadorf, line 160:
```suggestion
    """Convert ase Atoms instance to spglib cell."""
```

aiidalab_widgets_base/utils/__init__.py
```diff
@@ -153,3 +154,12 @@ class StatusHTML(_StatusWidgetMixin, ipw.HTML):
     @traitlets.observe("message")
     def _observe_message(self, change):
         self.show_temporary_message(change["new"])
+
+
+def ase2spglib(ase_structure: Atoms):
```
csadorf, line 159: Can you declare the return type here?

aiidalab_widgets_base/utils/__init__.py
```diff
@@ -153,3 +154,12 @@ class StatusHTML(_StatusWidgetMixin, ipw.HTML):
...
+    """convert ase Atoms instance to spglib cell"""
+    lattice = ase_structure.get_cell()
+    positions = ase_structure.get_scaled_positions()
+    numbers = ase_structure.get_atomic_numbers()
+
+    return (lattice, positions, numbers)
```
csadorf, line 165: Is this the definition of an spglib cell? Can we maybe add a link to the corresponding docs to the doc-string?

aiidalab_widgets_base/viewers.py
```diff
@@ -383,10 +397,14 @@ def _cell_tab(self):
...
 
-        self._observe_cell()
+        self.cell_spacegroup = ipw.HTML()
+        self.cell_hall = ipw.HTML()
+
+        self._observe_ase_structure()
 
         return ipw.VBox(
             [
+                ipw.HTML("Length unit: Angstrom"),
```
csadorf, line 407: Afaik the unit, unlike the name, is typically not capitalized.
```suggestion
                ipw.HTML("Length unit: angstrom"),
```

aiidalab_widgets_base/viewers.py
```diff
@@ -408,12 +426,24 @@ def _cell_tab(self):
...
+                            [
+                                ipw.HTML("Angles:"),
+                                self.cell_alpha,
+                                self.cell_beta,
+                                self.cell_gamma,
+                            ]
+                        ),
+                        ipw.VBox(
+                            [
+                                ipw.HTML("Symmetry infomation:"),
```
csadorf, line 441:
```suggestion
                                ipw.HTML("Symmetry information:"),
```

aiidalab_widgets_base/viewers.py
```diff
@@ -645,6 +675,10 @@ def _observe_selection(self, _=None):
         self.highlight_atoms(self.selection)
         self._selected_atoms.value = list_to_string_range(self.selection, shift=1)
 
+        # if atom selected from nglview, shift to selection tab
```
csadorf, line 678:
```suggestion
        # if atom is selected from nglview, shift to selection tab
```

aiidalab_widgets_base/viewers.py
```diff
@@ -167,15 +168,16 @@ class _StructureDataBaseViewer(ipw.VBox):
     selection = List(Int)
     selection_adv = Unicode()
     supercell = List(Int)
-    cell = Instance(Cell, allow_none=True)
+    cell = Instance(Cell, allow_none=True)  # Will be deprecated
```
csadorf, line 171: 1. Do we need to add a deprecation warning then?
2. What is the motivation for this?

aiidalab_widgets_base/viewers.py
```diff
@@ -153,11 +154,11 @@ def __init__(self, parameter, downloadable=True, **kwargs):
 
 
 class _StructureDataBaseViewer(ipw.VBox):
-    """Base viewer class for AiiDA structure or trajectory objects.
+    """Base viewer class for AiiDA structure or ase structure objects.
 
     :param configure_view: If True, add configuration tabs (deprecated)
     :type configure_view: bool
-    :param configuration_tabs: List of configuration tabs (default: ["Selection", "Appearance", "Cell", "Download"])
+    :param configuration_tabs: List of configuration tabs (default: ["Cell", "Selection", "Cell", "Download"])
```
csadorf, line 161: What are the allowed values here? Should we maybe provide an `Enum` of valid values?

aiidalab_widgets_base/viewers.py
```diff
@@ -167,15 +168,16 @@ class _StructureDataBaseViewer(ipw.VBox):
...
+    ase_structure = Instance(Atoms, allow_none=True)
     DEFAULT_SELECTION_OPACITY = 0.2
     DEFAULT_SELECTION_RADIUS = 6
     DEFAULT_SELECTION_COLOR = "green"
 
     def __init__(
         self,
         configure_view=True,
-        configuration_tabs=["Selection", "Appearance", "Cell", "Download"],
+        configuration_tabs=["Cell", "Selection", "Appearance", "Download"],
```
csadorf, line 180: [One should not use mutable default values in Python.](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/)

csadorf (changes requested):

aiidalab_widgets_base/viewers.py
unkcpz: per 1: Maybe (only if other app use this widget and directly use `cell` traitslet, @yakutovicha if the apps from empa using this directly). 

per 2: The motivation to replace `cell` with `ase_structure` is to read the symmetry information from the traitlets . From my point of view, it makes more sense to store and use full structure.

aiidalab_widgets_base/viewers.py
unkcpz: These are four allowed values. I think `Enum` makes it more robust, will give it a try.

csadorf: "Cell" appears twice.

unkcpz: As we discussed in the meeting, not necessary change to `Enum` the scope of these options are all only used in this function for mapping.

aiidalab_widgets_base/utils/__init__.py
unkcpz: Changed.

aiidalab_widgets_base/utils/__init__.py
unkcpz: Link https://spglib.github.io/spglib/python-spglib.html#crystal-structure-cell added to docstring.

aiidalab_widgets_base/viewers.py
csadorf: I'll wait for @yakutovicha to comment on this.

yakutovicha: The `structure` trait of the `StructureDataViewer` child class serves the purpose of `ase_structure`. You can extract all the information from it.

> per 2: The motivation to replace `cell` with `ase_structure` is to read the symmetry information from the traitlets . From my point of view, it makes more sense to store and use full structure.

We actually thought of `_StructureDataBaseViewer` as a base class that could be used also for the trajectory viewers. I would, therefore, be very careful in changing this part.

I would at least do this particular change in another PR so that we could carefully consider all the pros and cons.

yakutovicha: Just to mention that @danielhollas is currently preparing things for the trajectory viewer, so we might want to revisit the design of `_StructureDataBaseViewer` class.

I can create a separate issue to collect the ideas. What do you think @unkcpz and @danielhollas?

aiidalab_widgets_base/viewers.py
```diff
@@ -331,9 +337,10 @@ def change_camera(change):
             [supercell_selector, background_color, camera_type, center_button]
         )
 
-    @observe("cell")
-    def _observe_cell(self, _=None):
-        if self.cell:
+    @observe("ase_structure")
+    def _observe_ase_structure(self, _=None):
+        if self.ase_structure:
```
danielhollas, line 342: @unkcpz  In this logic, please take into account non-periodic systems by checking the pbc attribute of the ase object.

aiidalab_widgets_base/viewers.py
unkcpz: Thanks @yakutovicha please open an issue for trajectory. 
I will revert this part in the pr. Basically I need the ase.Atom only for parse the symmetry info.

unkcpz: Look at the implementation of this PR again, if I revert back, there is no essential changes left.
I think if we call this class a 'StructureData' viewer, the positions info should be included with is not in the original implementation, which makes not much sense to me. 
@danielhollas Could you have a look at this implementation to see if this will conflict with your trajectory feature, if not, it is much easier to keep the base traitslet type change here. 
Therefore, I'll remove the `cell` traitslet if it is fine for you.

aiidalab_widgets_base/viewers.py
unkcpz: Good point! @danielhollas 

I add a check and only parse and attach symmetry when pbc shows it is 3-D

danielhollas: @unkcpz it might also be nice to automatically hide the Cell tab in that case. Not sure how much work that would be so perhaps in another PR?

aiidalab_widgets_base/viewers.py
```diff
@@ -759,10 +796,10 @@ def _observe_structure(self, change):
         # Remove the current structure(s) from the viewer.
         if change["new"] is not None:
             self.set_trait("displayed_structure", change["new"].repeat(self.supercell))
-            self.set_trait("cell", change["new"].cell)
+            self.set_trait("ase_structure", change["new"])
```
danielhollas, line 799: I don't think this works in general case because the `structure` trait can be AiiDA node (StructureData or CifData). You would need to move this to the `displayed_structure` observer, but as mentioned in the other comment, I think it is easier to just move the `displayed_structure` trait to the parent class.

danielhollas (changes requested):
@unkcpz I left a couple comments, let me know if my suggestions make sense.

aiidalab_widgets_base/viewers.py
unkcpz: Probably can update the tab widget but then need to bring it back when 3D structure is selected again, seems a bit tedious to make it work correctly, I'll take a look. 
Or we can make the widget support  the 2D and 1D with big cell set like we have for [optimade-client]?(https://optimadeclient.materialscloud.io/). From where if you search structure in MC2D database. Not sure I remember how we do that, I'll check it.

unkcpz:
Thanks for the review! @danielhollas 
I think I am a bit confused with the design here. Can we have a short meeting to discuss it (we can meet next aiidalab meeting or quickly set a meeting if you are available)? 
The catch here is the `_StructureDataBaseViewer` has no positions info but I need to put the symmetry info in the `cell` tab which needs to use cell + position info. We need first to clarify the variable names used for our class since 'cell' sometimes only refers to lattice but in some libraries refers to the whole structure.

aiidalab_widgets_base/viewers.py
yakutovicha: > Look at the implementation of this PR again, if I revert back, there is no essential changes left.

Not really, I guess the purpose of this PR is to have the symmetry information displayed. This can stay.

>the positions info should be included with is not in the original implementation

sorry, I don't understand this, could you maybe rephrase?


> Just a correction to this, the structure trait can be either Atoms or AiiDA node, whereas `displayed_structure` is guaranteed to be the ASE Atoms object. So that's why I think it would make sense to move here.

No, it **_accepts_** both Atoms and AiiDA node, but then they get immediately converted into an ASE object:

https://github.com/aiidalab/aiidalab-widgets-base/blob/cabc5df932f41fcaf9907ac1ff2f0216ba20a700/aiidalab_widgets_base/viewers.py#L737-L738

> Regardless, I am fine with removing the cell traitlet. Since this is inside the private class, other things should not be using it anyway no?

It will also stay as a public member of the derived classes, we still need to be careful about removing it.

[P.S.]
Overall, I would like to stress that generally, I agree - we might change the classes to better satisfy our needs. We might do it in a separate PR (as I was suggesting) or just bite the bullet and do everything here. 

I try to be practical, that is why I suggest merging things that are clear and addressing the rest in a separate PR. If you want to do everything here - that is also fine with me. But keep in mind that it will cause some delays and long difficult-to-follow discussions.

yakutovicha:
> Thanks for the review! @danielhollas I think I am a bit confused with the design here. Can we have a short meeting to discuss it (we can meet next aiidalab meeting or quickly set a meeting if you are available)? The catch here is the `_StructureDataBaseViewer` has no positions info but I need to put the symmetry info in the `cell` tab which needs to use cell + position info. We need first to clarify the variable names used for our class since 'cell' sometimes only refers to lattice but in some libraries refers to the whole structure.

I agree, we need to meet to clear things out.

Just as a brief mention: the idea to keep the `structure` trait outside of `_StructureDataBaseViewer` is coming from the generality principle. We wanted to reuse the same base class for the trajectory, where the trait would probably be called `trajectory`.

danielhollas:
Sure, let's have a quick meeting. The next AiiDAlab meeting is after the Psi-k conference and I cannot attend unfortunately since I am at another conference in Rennes. Could we do a short meeting today or tomorrow afternoon?

@yakutovicha to clarify my suggestion, I would move the `displayed_structure` to the base class. That way there is a clearer separation of concern, since the base class is then in charge of what is displayed. The structure (or trajectory) traits would then stay in the concrete viewers. StructureDataViewer would then simply pass the structure info from structure trait to displayed_structure trait (converting to ase object from AiiDA node object), while the TrajectoryDataViewer would be passing the individual structure based on some selection widget.

yakutovicha:
> Could we do a short meeting today or tomorrow afternoon?

I am available both today and tomorrow. @unkcpz what about you?

unkcpz:
I have time today 3pm-3:30pm and tomorrow after 4pm. If you can find time, let me know. Thanks!

danielhollas:
@unkcpz @yakutovicha I am available now but tomorrow works as well.

yakutovicha:
> @unkcpz @yakutovicha I am available now but tomorrow works as well.

@danielhollas I've just sent you the link via email.

aiidalab_widgets_base/viewers.py
danielhollas:
```suggestion
        if self.ase_structure and any(self.ase_structure.pbc):
```

unkcpz:
Hi @yakutovicha @danielhollas, I adapt the changes without touching the `cell` trait. I hope this is what we agreed on.

danielhollas (dismissed):
@unkcpz thanks, the code looks good to me. I have some suggestions / questions, but those could be done in subsequent PR if needed.

I did not actually test this, I am leaving that to somebody with more solid state background.

aiidalab_widgets_base/viewers.py
```diff
@@ -335,7 +340,8 @@ def change_camera(change):
 
     @observe("cell")
     def _observe_cell(self, _=None):
-        if self.cell:
+        # only update cell info when it is a 3D structure.
+        if self.cell and all(self.structure.pbc):
```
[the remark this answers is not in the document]
unkcpz: Thanks for mentioning this. Actually, I made some effort to implement that but failed. The reason is that the code either hides the tab or adds a message inside the cell tab makes more sense to put in `_cell_tab` method, but in there the `self.structure` trait is not being set yet. 
Adding that too this PR will bring more coupling. But you are right, if it is okay with you, I'll open an issue for this and leave it to the future after we have decoupled the base structure data view.

aiidalab_widgets_base/viewers.py
```diff
@@ -335,7 +340,8 @@ def change_camera(change):
 
     @observe("cell")
     def _observe_cell(self, _=None):
-        if self.cell:
+        # only update cell info when it is a 3D structure.
```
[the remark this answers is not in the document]
unkcpz: There are two reasons that make this not easy to show. First, if it is a 2D structure, the spglib does not work well with fining the symmetry for 2D structure which has an [different set of space group](https://www.iucr.org/education/pamphlets/21/supplement). Secondly, I have no idea what to show for the cell information, do we still show a, b, c with c set to a large value as mimicking the vacuum layer, or don't show the c and beta in the tab.

unkcpz (commented):
Thanks! @danielhollas

aiidalab_widgets_base/viewers.py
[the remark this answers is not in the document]
unkcpz: Test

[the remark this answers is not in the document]
yakutovicha: @unkcpz did you open the issue? If yes, please mention it here for the record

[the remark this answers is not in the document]
unkcpz: I just open the issue https://github.com/aiidalab/aiidalab-widgets-base/issues/351 for this.

yakutovicha (dismissed):
All good also from my side, thanks @unkcpz!

unkcpz:
Okay, a bit late than the PR of @danielhollas (3 min 🥺), introduce a conflict here. I resolve the conflict and rebase it and should be fine now. Nothing is charged so I'll merge it without your approval.

PR #335, opened 2022-07-31
FEATURE: WorkChainNodeViewerWidget

danielhollas:
As a developer, I want to quickly inspect why a workchain exited with non-zero status or excepted, directly in the ProcessNodesTreeWidget. 

I have implemented a simple WorkChainNodeViewerWidget that essentially prints out the output of `verdi process report` command. I strip out the dates to make the output less noisy, and I only output reports from the selected workchain itself, not from its descendants, as it could get very big for large workflows.

Here's how it looks:

obrazek

### DIFF
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -2,6 +2,7 @@
 # pylint: disable=no-self-use
 
 import base64
+import re
 import warnings
 from copy import deepcopy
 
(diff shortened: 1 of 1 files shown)


yakutovicha (approved):
I think it is a very good addition, thanks a lot @danielhollas!

danielhollas:
Thanks  @yakutovicha!

PR #336, opened 2022-07-31
NodesTreeWidget: Red icon for failed processes

danielhollas:
This has bugged me for a while. We were displaying green icon for processes that finished but had non-zero exit status.

BEFORE:
obrazek

AFTER:
obrazek

### DIFF
--- a/aiidalab_widgets_base/nodes.py
+++ b/aiidalab_widgets_base/nodes.py
@@ -286,7 +286,7 @@ class NodesTreeWidget(ipw.Output):
                 else process_node.process_state
             )
             tree_node.icon_style = self.PROCESS_STATE_STYLE.get(
-                process_state, PROCESS_STATE_STYLE_DEFAULT
+                process_state, self.PROCESS_STATE_STYLE_DEFAULT
             )
 
     def update(self, _=None):
--- a/aiidalab_widgets_base/nodes.py
+++ b/aiidalab_widgets_base/nodes.py
@@ -278,8 +278,12 @@ class NodesTreeWidget(ipw.Output):
         if isinstance(tree_node, AiidaProcessNodeTreeNode):
             process_node = load_node(tree_node.pk)
             tree_node.name = calc_info(process_node)
-            tree_node.icon_style = self.PROCESS_STATE_STYLE.get(
-                process_node.process_state, self.PROCESS_STATE_STYLE_DEFAULT
+            tree_node.icon_style = (
+                self.PROCESS_STATE_STYLE.get(
+                    process_node.process_state, self.PROCESS_STATE_STYLE_DEFAULT
+                )
+                if not process_node.is_failed
+                else "danger"
             )
 
     def update(self, _=None):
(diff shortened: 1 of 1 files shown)


yakutovicha:
I let @csadorf review it, as he mainly contributed to this part of the code.

csadorf (commented):
Generally no objections, but it seems a bit odd that we need to special case this. Do you know what the value of `process_node.process_state` is in case that the `process_node.is_failed` condition is True?

danielhollas:
It is unfortunate, but the ProcessState is FINISHED in this case so there's no way to distinguish this condition based on this. (Note that also the method 'is_finished' returns true if the process finished with non-zero exit status).

I am pretty sure I saw some bugs related to this in the aiidalab-qe app as well, wil laubmit PRs when I get to it.

aiidalab_widgets_base/nodes.py
```diff
@@ -278,8 +278,12 @@ def _update_tree_node(self, tree_node):
...
             tree_node.name = calc_info(process_node)
-            tree_node.icon_style = self.PROCESS_STATE_STYLE.get(
-                process_node.process_state, self.PROCESS_STATE_STYLE_DEFAULT
+            tree_node.icon_style = (
+                self.PROCESS_STATE_STYLE.get(
+                    process_node.process_state, self.PROCESS_STATE_STYLE_DEFAULT
+                )
+                if not process_node.is_failed
+                else "danger"
             )
```
csadorf, line 287: So the one thing that bugs me is the hard-coded `"danger"` and I think the easiest way to avoid this is to simply determine the state first:

```suggestion
            # Override the process state in case that the process node has failed:
            # (This could be refactored with structural pattern matching with py>=3.10.)
            process_state = ProcessState.EXCEPTED if process_node.is_failed else process_node.process_state
            tree_node.icon_style = self.PROCESS_STATE_STYLE.get(process_state, PROCESS_STATE_STYLE_DEFAULT)
```
_The line breaks are probably wrong, but the pre-commit hooks should fix it._

csadorf:
@danielhollas Sorry, looks like I introduced a bug with my suggestion: https://results.pre-commit.ci/run/github/135423835/1659623008.H_oLFNBBSs-SpAAAzru51A

aiidalab_widgets_base/nodes.py
```diff
@@ -278,8 +278,15 @@ def _update_tree_node(self, tree_node):
...
+            # Override the process state in case that the process node has failed:
+            # (This could be refactored with structural pattern matching with py>=3.10.)
+            process_state = (
+                ProcessState.EXCEPTED
+                if process_node.is_failed
+                else process_node.process_state
+            )
             tree_node.icon_style = self.PROCESS_STATE_STYLE.get(
-                process_node.process_state, self.PROCESS_STATE_STYLE_DEFAULT
+                process_state, PROCESS_STATE_STYLE_DEFAULT
```
danielhollas, line 289:
```suggestion
                process_state, self.PROCESS_STATE_STYLE_DEFAULT
```

danielhollas:
@csadorf thanks, I've fixed the code and tested locally so this should be good to go.

csadorf (approved):

csadorf:
@danielhollas Thanks a lot!

PR #339, opened 2022-08-13
StructureUploadWidget: Handle invalid files

danielhollas:
Added more error handling to StructureUploadWidget, using an HTML widget to output errors instead of print statements. 
Partly extracted from #332, but no support for TrajectoryData is added here, although I tried to refactor a bit in preparation for it.

Besides the added error handling, there are two functional changes that deserve discussion:

1. Fail if multiple structures are detected in the file, instead of just picking the first one. To me this seems like a safer approach (until we implement opt-in support for TrajectoryData).
2. I changed the signature of the `utils.get_ase_from_file` function so that it returns all read structures. It's up to the caller to decide what to do (see point 1). It's also up to the caller to catch any exceptions). I don't know if functions in `utils/` are considered part of the public API though, perhaps I should create a new function and deprecate the old one?

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -414,7 +414,9 @@ class StructureUploadWidget(ipw.VBox):
             try:
                 return CifData(file=io.BytesIO(content))
             except Exception as e:
-                self._status_message.message = f"Could not parse CIF file {fname}: {e}"
+                self._status_message.message = f"""
+                    <div class="alert alert-danger">Could not parse CIF file {fname}: {e}</div>
+                    """
                 return None
 
         with tempfile.NamedTemporaryFile(suffix=suffix) as temp_file:
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -411,12 +411,10 @@ class StructureUploadWidget(ipw.VBox):
     def _read_structure(self, fname, content):
         suffix = "".join(pathlib.Path(fname).suffixes)
         if suffix == ".cif":
-            from CifFile import StarError
-
             try:
                 return CifData(file=io.BytesIO(content))
-            except StarError:
-                self._status_message.message = f"Could not parse CIF file {fname}"
+            except Exception as e:
+                self._status_message.message = f"Could not parse CIF file {fname}: {e}"
                 return None
 
         with tempfile.NamedTemporaryFile(suffix=suffix) as temp_file:
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -4,6 +4,7 @@
 import datetime
 import functools
 import io
+import pathlib
 import tempfile
 from collections import OrderedDict
 
@@ -368,16 +369,23 @@ class StructureUploadWidget(ipw.VBox):
         Supported structure formats
         </a>"""
         )
+        self.output = ipw.HTML("")
         self.file_upload.observe(self._on_file_upload, names="value")
-        super().__init__(children=[self.file_upload, supported_formats])
+        super().__init__(children=[self.file_upload, supported_formats, self.output])
 
     def _validate_and_fix_ase_cell(self, ase_structure, vacuum_ang=10.0):
         """
         Checks if the ase Atoms object has a cell set,
         otherwise sets it to bounding box plus specified "vacuum" space
         """
+        if not ase_structure:
+            return None
+
         cell = ase_structure.cell
 
+        # TODO: Since AiiDA 2.0, zero cell is possible if PBC=false
+        # so we should honor that here and do not put artificial cell
+        # around gas phase molecules.
         if (
             np.linalg.norm(cell[0]) < 0.1
             or np.linalg.norm(cell[1]) < 0.1
@@ -393,19 +401,42 @@ class StructureUploadWidget(ipw.VBox):
 
     def _on_file_upload(self, change=None):
         """When file upload button is pressed."""
+        self.output.value = ""
         for fname, item in change["new"].items():
-            frmt = fname.split(".")[-1]
-            if frmt == "cif":
-                self.structure = CifData(file=io.BytesIO(item["content"]))
-            else:
-                with tempfile.NamedTemporaryFile(suffix=f".{frmt}") as temp_file:
-                    temp_file.write(item["content"])
-                    temp_file.flush()
-                    self.structure = self._validate_and_fix_ase_cell(
-                        get_ase_from_file(temp_file.name)
-                    )
+            self.structure = self._read_structure(fname, item["content"])
             self.file_upload.value.clear()
             break
+
+    def _read_structure(self, fname, content):
+        suffix = "".join(pathlib.Path(fname).suffixes)
+        if suffix == ".cif":
+            from CifFile import StarError
+
+            try:
+                return CifData(file=io.BytesIO(content))
+            except StarError:
+                self.output.value = f"Could not parse CIF file {fname}"
+                return None
+
+        with tempfile.NamedTemporaryFile(suffix=suffix) as temp_file:
+            temp_file.write(content)
+            temp_file.flush()
+            try:
+                structures = get_ase_from_file(temp_file.name)
+            except ValueError as e:
+                self.output.value = f"ERROR: {e}"
+                return None
+            except KeyError:
+                self.output.value = f"ERROR: Could not parse file {fname}"
+                return None
+
+            if len(structures) > 1:
+                self.output.value = (
+                    f"ERROR: More than one structure found in file {fname}"
+                )
+                return None
+
+            return self._validate_and_fix_ase_cell(structures[0])
 
 
 class StructureExamplesWidget(ipw.VBox):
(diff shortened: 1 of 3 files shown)


aiidalab_widgets_base/utils/__init__.py
```diff
@@ -36,15 +36,8 @@ def get_ase_from_file(fname, format=None):  # pylint: disable=redefined-builtin
...
-        return False
-    if len(traj) > 1:
-        print(
-            "Warning: Uploaded file {} contained more than one structure. Selecting the first one.".format(
-                fname
-            )
-        )
-    return traj[0]
+        raise ValueError(f"Could not read any information from the file {fname}")
+    return traj
```
danielhollas, line 40: Functional change here, returning all structures and let the caller decide what to do. (in preparation for TrajectoryData support).

aiidalab_widgets_base/structures.py
```diff
@@ -437,7 +468,7 @@ def _on_select_structure(self, change):  # pylint: disable=unused-argument
         """When structure is selected."""
 
         self.structure = (
-            get_ase_from_file(self._select_structure.value)
+            get_ase_from_file(self._select_structure.value)[0]
```
danielhollas, line 471: I am purposefully omitting any error handling here, since it's up to the programmer to provide valid example files.

csadorf:
@yakutovicha Are you available to review?

aiidalab_widgets_base/structures.py
```diff
@@ -368,16 +369,23 @@ def __init__(self, title="", description="Upload Structure"):
         Supported structure formats
         </a>"""
         )
+        self.output = ipw.HTML("")
```
yakutovicha, line 372: I would say, suggest to use `StatusHTML` instead https://github.com/aiidalab/aiidalab-widgets-base/blob/3897957e33dda90fbbcc0eb8f8b4905bd6d31731/aiidalab_widgets_base/utils/__init__.py#L146

It shows a message for some time, and then it disappears.

yakutovicha (changes requested):
Overall, looks good. Please do the small change I requested. Once it is done, I will test the things.

aiidalab_widgets_base/structures.py
```diff
@@ -393,20 +401,43 @@ def _validate_and_fix_ase_cell(self, ase_structure, vacuum_ang=10.0):
...
+            try:
+                structures = get_ase_from_file(temp_file.name)
+            except ValueError as e:
+                self.output.value = f"ERROR: {e}"
+                return None
+            except KeyError:
+                self.output.value = f"ERROR: Could not parse file {fname}"
+                return None
+
+            if len(structures) > 1:
```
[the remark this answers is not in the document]
yakutovicha: agreed.

aiidalab_widgets_base/utils/__init__.py
yakutovicha: looks good to me.

aiidalab_widgets_base/structures.py
```diff
@@ -393,20 +401,43 @@ def _validate_and_fix_ase_cell(self, ase_structure, vacuum_ang=10.0):
...
-                        get_ase_from_file(temp_file.name)
-                    )
+            self.structure = self._read_structure(fname, item["content"])
             self.file_upload.value.clear()
             break
 
+    def _read_structure(self, fname, content):
+        suffix = "".join(pathlib.Path(fname).suffixes)
+        if suffix == ".cif":
+            from CifFile import StarError
```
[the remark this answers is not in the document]
yakutovicha: Do you mean just catching an `Exception`? I think it is not suggested.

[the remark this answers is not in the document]
danielhollas: Yeah, I essentially meant a naked except
```python
try:
   ...
except:
   ...
```
which I guess is mostly equivalent to catching the base exception class. Catching a specific exception is nicer, but I don't quite like suddenly importing something from what seems to be a transitive dependency.

aiidalab_widgets_base/structures.py
danielhollas: @yakutovicha thank you. I've changed the code to use the widget.

I must say I am not a big fan of automatically disappearing error messages though. Depending on the size of the message, you might need more time to read _and understand_ it, but if the time is too long, then it might not get cleared before the user takes some other action.

Also, if the message should be transient, it would be nicer to use some kind of hover tooltip, otherwise the UI is jumping up and down. But that's a discussion for another issue I guess. :-)

aiidalab_widgets_base/structures.py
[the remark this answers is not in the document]
yakutovicha: @csadorf what do you think about this?

aiidalab_widgets_base/structures.py
```diff
@@ -394,19 +404,51 @@ def _validate_and_fix_ase_cell(self, ase_structure, vacuum_ang=10.0):
...
+                    </div>
+                    """
+                return None
+
+            if len(structures) > 1:
+                self._status_message.message = f"""
+                    <div class="alert alert-info">
+                    <strong>ERROR: More than one structure found in file {fname}</strong>
+                    </div>
+                    """
```
yakutovicha, line 447:
```suggestion
                self._status_message.message = f"""
                    <div class="alert alert-danger">ERROR: More than one structure found in file {fname}</div>
                    """
```

aiidalab_widgets_base/structures.py
```diff
@@ -394,19 +404,51 @@ def _validate_and_fix_ase_cell(self, ase_structure, vacuum_ang=10.0):
...
+                    <strong>ERROR: {e}</strong>
+                    </div>
+                    """
+                return None
+            except KeyError:
+                self._status_message.message = f"""
+                    <div class="alert alert-info">
+                    <strong>ERROR: Could not parse file {fname}</strong>
+                    </div>
+                    """
```
yakutovicha, line 439:
```suggestion
                self._status_message.message = f"""
                    <div class="alert alert-danger">ERROR: Could not parse file {fname}</div>
                    """
```

aiidalab_widgets_base/structures.py
```diff
@@ -394,19 +404,51 @@ def _validate_and_fix_ase_cell(self, ase_structure, vacuum_ang=10.0):
...
+            temp_file.write(content)
+            temp_file.flush()
+            try:
+                structures = get_ase_from_file(temp_file.name)
+            except ValueError as e:
+                self._status_message.message = f"""
+                    <div class="alert alert-info">
+                    <strong>ERROR: {e}</strong>
+                    </div>
+                    """
```
yakutovicha, line 432:
```suggestion
                self._status_message.message = f"""
                    <div class="alert alert-danger">ERROR: {e}</div>
                    """
```

I would suggest making it alert-danger and don't use bold text. It is already bold enough :)

Same for the occurrences below.

yakutovicha (commented):
I tested it locally, all works great, thanks @danielhollas!

Just a couple of suggestions for the Error messages format. We should also clarify the `try-except` issue. Honestly, I would be also fine with keeping it as a bare exception. But let's see what @csadorf says. I am going to ping him again.

aiidalab_widgets_base/structures.py
yakutovicha: To me, it looked quite okay, tbh.

aiidalab_widgets_base/structures.py
```diff
@@ -393,20 +401,43 @@ def _validate_and_fix_ase_cell(self, ase_structure, vacuum_ang=10.0):
...
+
+        with tempfile.NamedTemporaryFile(suffix=suffix) as temp_file:
+            temp_file.write(content)
+            temp_file.flush()
+            try:
+                structures = get_ase_from_file(temp_file.name)
+            except ValueError as e:
+                self.output.value = f"ERROR: {e}"
+                return None
+            except KeyError:
```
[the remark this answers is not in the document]
yakutovicha: tested, all good.

aiidalab_widgets_base/structures.py
```diff
@@ -393,20 +401,43 @@ def _validate_and_fix_ase_cell(self, ase_structure, vacuum_ang=10.0):
...
+            except StarError:
+                self.output.value = f"Could not parse CIF file {fname}"
+                return None
+
+        with tempfile.NamedTemporaryFile(suffix=suffix) as temp_file:
+            temp_file.write(content)
+            temp_file.flush()
+            try:
+                structures = get_ase_from_file(temp_file.name)
+            except ValueError as e:
```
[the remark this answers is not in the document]
yakutovicha: tested, all good.

aiidalab_widgets_base/structures.py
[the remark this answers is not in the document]
csadorf: I agree with @danielhollas that we should not use this specific exception class since it is declared in a transitive dependency.

Catching a more specific exception class here is very difficult, because aiida does not explicitly define which Exceptions can be raised so there is no explicit API contract regarding this.

It seems to me that the errors thrown by pycifrw (transitive) are indeed `StarError` which is derived from `Exception` (should probably be a `Runtime` or `ValueError`). The alternative ways to load Cifs via ase (also transitive) can result in a variety of different exceptions which all appear to be either `Value` or `RuntimeError`s.

In conclusion, I think we have to catch `Exception` here, but we should not use a bare `except:` clause. The reason for this is explained [here](https://stackoverflow.com/a/54948581).

danielhollas:
@csadorf @yakutovicha thanks for the review and comments. I've implemented your suggestions so this should be good to go.

yakutovicha (approved):
All good, thanks a lot @danielhollas!

PR #340, opened 2022-08-15
StructureBrowserWidget: Customizable query types

danielhollas:
By default, StructureBrowserWidget returns CifData and StructureData nodes. In this PR, we add the class constructor argument  `structure_query_type` to make this customizable. The programmer can decide to only search for CifData or StructureData, or in my particular use case to extend the search for TrajectoryData.

Partly extracted from #332. 

@yakutovicha we've discussed that we could create a general NodeBrowserWidget and deprecate the StructureBrowserWidget. However, upon closer inspection, the StructureBrowserWidget contains a lot of code that only make sense for structure-like nodes. Creating a generalized browser widget might still be useful in the future, but I suspect in most cases the widget would still need to be specialized to make it useful to a particular use case. The `ProcessListWidget` is one such example. Here I simply submit a minimal PR to support my use case.

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -8,7 +8,6 @@ import pathlib
 import tempfile
 from collections import OrderedDict
 
-# ASE imports
 import ase
 import ipywidgets as ipw
 import numpy as np
@@ -29,12 +28,9 @@ from ase.data import chemical_symbols, covalent_radii
 from sklearn.decomposition import PCA
 from traitlets import Instance, Int, List, Unicode, Union, default, dlink, link, observe
 
-from aiidalab_widgets_base.utils import StatusHTML
-
-from .data import LigandSelectorWidget
-
 # Local imports
-from .utils import get_ase_from_file
+from .data import LigandSelectorWidget
+from .utils import StatusHTML, get_ase_from_file, get_formula
 from .viewers import StructureDataViewer
 
 CifData = DataFactory("cif")  # pylint: disable=invalid-name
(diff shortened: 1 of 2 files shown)


danielhollas:
@yakutovicha I know you're super busy so feel free to ignore this but just wanted to ask if you could take a look on this? 

Essentially I am trying to focus on releasing an alpha version of my app (which depends a bit on this). But if this is merged only after AWB-2.0 release, then I will need to migrate to aiida-2.0. I will need to do that eventually of course, but was hoping to postpone the migration for later.

yakutovicha:
> @yakutovicha I know you're super busy so feel free to ignore this but just wanted to ask if you could take a look on this?
> 
> Essentially I am trying to focus on releasing an alpha version of my app (which depends a bit on this). But if this is merged only after AWB-2.0 release, then I will need to migrate to aiida-2.0. I will need to do that eventually of course, but was hoping to postpone the migration for later.

Hey @danielhollas, I am going to review this today as a part of my Great Review Day :D

aiidalab_widgets_base/structures.py
```diff
@@ -448,15 +444,24 @@ def _default_structure(self):
...
-    def __init__(self, title=""):
+    def __init__(self, title="", query_types=None):
         self.title = title
 
         # Structure objects we want to query for.
-        self.query_structure_type = (DataFactory("structure"), DataFactory("cif"))
+        if query_types:
+            self.query_structure_type = query_types
+        else:
+            self.query_structure_type = (DataFactory("structure"), DataFactory("cif"))
```
yakutovicha, line 464: I am correcting myself here since I did it this way initially. But both StructureData and CifData are imported earlier, so we could just reuse them:
```suggestion
            self.query_structure_type = (StructureData, CifData)
```

aiidalab_widgets_base/utils/__init__.py
```diff
@@ -88,6 +93,21 @@ def string_range_to_list(strng, shift=-1):
     return singles, True
 
 
+def get_formula(data_node):
+    """A wrapper for getting a molecular formula out of the AiiDA Data node"""
+    if isinstance(data_node, TrajectoryData):
+        # TrajectoryData can only hold structures with the same chemical formula,
+        # so this approach is sound.
```
yakutovicha, line 100: Are you sure? If one runs Grand Canonical MC the number of particles does change.

yakutovicha (changes requested):
Hey @danielhollas, thanks a lot!

All good, just one minor change request and one question.

aiidalab_widgets_base/utils/__init__.py
danielhollas: I tried creating TrajectoryData from two different StructureData nodes with different chemical composition and got an exception, but feel free to verify. (tested with AiiDA 1.6 in AiiDAlab image)

yakutovicha: Ok, if that is the constraint they have, maybe we keep it like this for the moment. But please mention this in the comment above. Just to avoid that people will have to re-discover this in the future. Thanks!

danielhollas: To me the comment says that already. :-) Do you have a specific wording in mind?

yakutovicha: You are right, sorry, all good.

danielhollas:
@yakutovicha thanks for the review. I've rebased on master, commited your suggestion and did a quick test. This is good to go from my side.

PR #341, opened 2022-08-16
ComputationalResourceWidget: Hide comp_resources_database if it is empty

danielhollas:
As a developer I'd like to switch from the deprecated CodeDropdown to ComputerResourceWidget. 
The new widget is nice, but since I am not using any of the Swiss computer clusters, it is more noisy then what I had before without the added convenience.

Regardless of my use case. there's no point in displaying an empty widgets if the selected `input_plugin` is not available in any of the computers in the database (such as the case for ORCA program). In this case all the dropdowns are empty.

**Before:**
image

**After:**
image

It's not clear to me how the Quick Setup works and whether it makes sense to display it in this scenario.

### DIFF
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -57,7 +57,10 @@ class ComputationalResourcesWidget(ipw.VBox):
         self.output = ipw.HTML()
         self.setup_message = StatusHTML()
         self.code_select_dropdown = ipw.Dropdown(
-            description=description, disabled=True, value=None
+            description=description,
+            disabled=True,
+            value=None,
+            style={"description_width": "initial"},
         )
         traitlets.link((self, "codes"), (self.code_select_dropdown, "options"))
         traitlets.link((self.code_select_dropdown, "value"), (self, "value"))
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -128,7 +128,7 @@ class ComputationalResourcesWidget(ipw.VBox):
         # Quick setup.
         quick_setup_button = ipw.Button(description="Quick Setup")
         quick_setup_button.on_click(self.quick_setup)
-        quick_setup = ipw.VBox(
+        self.quick_setup = ipw.VBox(
             children=[
                 self.ssh_computer_setup.username,
                 quick_setup_button,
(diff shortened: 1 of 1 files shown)


danielhollas:
@unkcpz since you are removing the CodeDropdown widget in #344, perhaps you could take a look at this PR. Thanks! (I don't have enough privileges on this repo to request you as a reviewer). This is not urgent though so take your time.

CC @yakutovicha

unkcpz:
@danielhollas No problem, I add myself as reviewer.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -229,15 +232,19 @@ def _setup_new_code(self, _=None):
...
                 }
-                display(
-                    ipw.HTML(
-                        """Please select the computer/code from a database to pre-fill the fields below."""
-                    ),
-                    self.comp_resources_database,
-                    self.setup_message,
-                    self.ssh_computer_setup.password_box,
-                    self.output_tab,
+                children = (
```
unkcpz, line 235: I assume the `children` here need to be a list?

aiidalab_widgets_base/computational_resources.py
```diff
@@ -57,7 +57,10 @@ def __init__(self, description="Select code:", path_to_root="../", **kwargs):
         self.output = ipw.HTML()
         self.setup_message = StatusHTML()
         self.code_select_dropdown = ipw.Dropdown(
-            description=description, disabled=True, value=None
+            description=description,
+            disabled=True,
+            value=None,
+            style={"description_width": "initial"},
```
unkcpz, line 63: This is nice.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -229,15 +232,19 @@ def _setup_new_code(self, _=None):
...
+                        ipw.HTML(
+                            """Please select the computer/code from a database to pre-fill the fields below."""
+                        ),
+                        self.comp_resources_database,
+                        self.ssh_computer_setup.password_box,
+                    ]
+                    if self.comp_resources_database.database
+                    else []
                 )
+                children += [self.setup_message, self.output_tab]
```
unkcpz, line 246: Not sure about this message and output also for the 'detailed setup', if that is the case, I think we need to decouple it from here. But not necessary in this PR.

unkcpz (changes requested):
thanks! @danielhollas 
Just a minor request for code clarity.

Do you test setup the code using the `Detailed setup` after this change? I suggest doing it at least once.

yakutovicha (changes requested):
If no computers are available in the database, Quick Setup should also be hidden, IMO. Because it relies entirely on the prefilled information from the database.

aiidalab_widgets_base/computational_resources.py
danielhollas: Yes, variable `children` is a list. (the opening parenthesis here is just for multiple line splitting)

aiidalab_widgets_base/computational_resources.py
danielhollas: @unkcpz not sure what you mean, but `self.setup_message` is used to display status messages and errors from Detailed Setup as well.

danielhollas:
@unkcpz @yakutovicha thanks for the reviews!

> If no computers are available in the database, Quick Setup should also be hidden, IMO. Because it relies entirely on the prefilled information from the database.

Great, thanks for the info, I've hidden it. Here's how it looks now.
image

> Do you test setup the code using the Detailed setup after this change? I suggest doing it at least once.

I did a quick test and it seemed to work. :crossed_fingers:

aiidalab_widgets_base/computational_resources.py
unkcpz: Nevermind my comment, I misunderstood about this code, I thought it is for quicksetup only.

unkcpz (approved):
@danielhollas thanks a lot! Looks all good to me.

yakutovicha (approved):
@danielhollas all good, thanks a lot!

PR #344, opened 2022-08-29
Migrate widgets to 2.x

unkcpz:
- [x] `computer_resource` widget
- [x] Drop dep aiida-core <=1.x
- [x] Remove all deprecate widgets
- [ ] ~~bump AWB version to `2.0.0`~~
- [x] computer/code setup widget support `aiida-code-registry` for both 1.x and 2.x core datatype.
- [x] Add tests to test computer/code setup from code registry failed for incompatibility.

Blocking the issue: https://github.com/aiidalab/aiidalab-qe/issues/279

If we implement this change means it is not compatible with 1.x anymore. 
I think we need to set more strict dependency for aiida-core and bump the version of this PR to `2.0.0`. 

@yakutovicha @csadorf, any thoughts?


csadorf:
> If we implement this change means it is not compatible with 1.x anymore. I think we need to set more strict dependency for aiida-core and bump the version of this PR to `2.0.0`.
> 
> @yakutovicha @csadorf, any thoughts?

Yes, if this change requires aiida-core >= 2.0.0 then this needs to be reflected in the declared dependencies.

yakutovicha:
> @yakutovicha @csadorf, any thoughts?

if those changes are enough, I am very happy :)

unkcpz:
We are going to deprecate some widgets such as `computer.py::SshComputerSetup`, I will also remove them.
But always smoke tests are very useful for such changes, is that possible to have the test framework ready for this PR? @csadorf. Let me know how to add it and I'll add the tests.

unkcpz:
> if those changes are enough, I am very happy :)

Sadly, I this may not enough, we need to have a way to separate [aiida-code-registry](https://github.com/aiidateam/aiida-code-registry) to support both 1.x and 2.x config files.

unkcpz:
Tests are all passed. But the computer/code setup configs from https://github.com/aiidateam/aiida-code-registry not working anymore. 
One possible way is we also release `aiida-core-registry` with a tag for instance `1.x` and make the config database support select data by the tag.

csadorf:
> Tests are all passed. But the computer/code setup configs from https://github.com/aiidateam/aiida-code-registry not working anymore. One possible way is we also release `aiida-core-registry` with a tag for instance `1.x` and make the config database support select data by the tag. Do you have any better way? @csadorf @yakutovicha

Can you elaborate on what exactly is not working? Maybe it is possible to do an automated migration? I'd prefer that over maintaining multiple versions for different "target AiiDA versions".

unkcpz:
> Can you elaborate on what exactly is not working?

From aiida 1.x to 2.x the transport and scheduler are changed from an endpoint, for example, `ssh` to `core.ssh` and `slurm` to `core.slurm`. They are hard coded in the config file.

yakutovicha:
> This is for sure a good option but will increase the complexity of the `aiida-code-registry` which is already not very convenient to add a new entity for HPC resources.

Ok, let me tell you my perspective on this.

1. If the registry is not convenient, we should make it convenient. What are the issues with adding a new code/computer there? I am happy to add a code/computer template or anything to make it more user-friendly.
Can you put some suggestions to the registry in a form of an issue? If things are not working as expected, we (as developers) should not be silent about them.
2. I believe that @csadorf is talking about an automated migration within AWB. The problem is - at some point the registry will have to be also migrated, which makes it unavoidable to keep old and new configs there. Tbh, I lean a bit towards the suggestion of @unkcpz. We might automatically generate the computer/code config files that are suited for the given version of AiiDA. It won't mean maintaining two registries, it would rather mean having one registry that automatically generates config files for AiiDA-1.x and AiiDA-2.x. If we agree, I can take care of this approach.

unkcpz:
> If the registry is not convenient, we should make it convenient. What are the issues with adding a new code/computer there? I am happy to add a code/computer template or anything to make it more user-friendly.

No issue with adding new code/computer. I misunderstood that @csadorf ask to have for same configs files have two versions of aiida supported. 
But I have some ideas in mind on how to improve the `aiida-code-registry`, I will make an AEP for it.

> I believe that @csadorf is talking about an automated migration within AWB. 

In this case, we need to think about which version to keep in `main` branch and doing transfer in ABW from 1.x -> 2.x or vice versa. So I'd vote to auto-generate config files AiiDA-1.x and AiiDA-2.x. 

@csadorf Sorry that I didn't read Sasha's comment carefully and did not make it very clear in the materials cloud meeting.

csadorf:
As I explained in the MC meeting today, we need to interpret the code registry as a database that is providing information for the automated configuration of AiiDA codes. The specific semantics (`ssh` vs `core.ssh`) are in that sense completely irrelevant to the database. As long as the information needed to configure codes in aiida-core 1.x and 2.x remains the same we can always do an automated translation between the different semantics, there is no need to maintain two different versions of the same database.

That translation can happen at various points (at the database level itself), when we read from the database, when we actually configure the codes. Is there any kind of API regarding the registry? Are there any other consumers despite the AWB library? If not, then I would suggest to simply migrate the whole database, otherwise, just do a translation here in the library code. In either case there is no need and we should not maintain multiple database versions.

danielhollas:
Just a note, it would be great if you could release one more AWB version before this is merged. :-)

danielhollas:
> I believe you prefer to have https://github.com/aiidalab/aiidalab-widgets-base/pull/348 merged before the release, right?

Yeah, that would be nice. More importantly for me, I am also working on a follow-up on #339, to support multiple structure upload in StructureUploadWidget. I'll submit a PR today and then you can judge whether it can be merged quickly or will need more thought and should not block release. Thanks!

unkcpz:
Hi @yakutovicha, since this is not block the aiidalab-qe, I think it is ready to have another round of review.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -702,10 +702,19 @@ def __init__(self, **kwargs):
             style=STYLE,
         )
 
+        # Memory per node.
+        self.default_memory_per_machine = ipw.IntText(
+            value=16000000,
+            step=1000000,
+            description="#Memory(RAM) per node (kB):",
```
csadorf, line 709: I don't think it's a good idea to ask the user to provide the value in kB since the number of machines providing RAM on the order of kB should at this point be restricted to edge and embedded device that are not typically known to run AiiDA or Jupyter. You can always internally convert the value if that is what is expected by the code plugin.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -888,11 +906,13 @@ def _reset(self):
         self.work_dir.value = ""
         self.mpirun_command.value = "mpirun -n {tot_num_mpiprocs}"
         self.mpiprocs_per_machine.value = 1
-        self.transport.value = "ssh"
+        self.default_memory_per_machine.value = 16000000
```
csadorf, line 909: Where is this default value coming from?

aiidalab_widgets_base/databases.py
```diff
@@ -475,7 +285,7 @@ def clean_up_database(self, database, plugin):
     def update(self, _=None):
         with self.hold_trait_notifications():
             database = requests.get(
-                "https://aiidateam.github.io/aiida-code-registry/database.json"
+                "https://aiidateam.github.io/aiida-code-registry/database_v2.json"
```
csadorf, line 288: I am not sure that the `hold_trait_notification()` context manager is really appropriate here. Are we changing multiple traits at once within the same thread? The hold should not be confused with a lock. See the [docs](https://traitlets.readthedocs.io/en/stable/using_traitlets.html#holding-trait-cross-validation-and-notifications) for reference.

Also, why not make the request outside of the hold?

_Since my comment is not directly related to this change set, please address it in a separate issue or PR if any action is deemed necessary._

aiidalab_widgets_base/process.py
```diff
@@ -96,26 +93,11 @@ def __init__(  # pylint: disable=too-many-arguments
...
-                (
-                    "The `input_dictionary_function` argument is deprecated and "
-                    "will be removed in the release 1.1 of the aiidalab-widgets-base package. "
-                    "Please use the `inputs_generator` argument instead."
-                ),
-                DeprecationWarning,
-            )
 
-        # Checking if the inputs generator is
+        # Checking if the inputs generator is callable
```
csadorf, line 100: I'd recommend the following sequence:
```python
if callable(inputs_generator):
    self.inputs_generator = inputs_generator
else:
    raise ValueError("The `inputs_generator` argument must be provided and a function that ...")
```

.readthedocs.yml
```diff
@@ -10,7 +10,7 @@ version: 2
 build:
     os: ubuntu-20.04
     tools:
-        python: '3.7'
+        python: '3.9'
```
csadorf, line 13: Why not 3.10?

csadorf (dismissed):

aiidalab_widgets_base/computational_resources.py
unkcpz: It's should be set for AiiDA 2.x, I think @yakutovicha add it to the computer setup and made it mandatory.
It is hard to say if we should provide the default value for this field, maybe we shouldn't provide the default one. 
But this comes with the problem that the code-registry-database doesn't have this field set yet, the "quick setup" of widget will fail since no value is provided.

aiidalab_widgets_base/process.py
unkcpz: I don't understand what you mean, the code you paste here is all the same.

csadorf: You have an additional `is None` check which is unnecessary I think (`None` is not `callable`).

unkcpz: Thanks! I see. I remove the `None` as the default value of `inputs_generator` argument.

aiidalab_widgets_base/computational_resources.py
csadorf: I don't really think it matters where it comes from, but I think there should be some kind of justification for it and then it should be documented here in the form of a short comment.

yakutovicha (commented):
Thanks, @unkcpz. Not too many comments from my side.

unkcpz:
@yakutovicha thanks!

> There is a way to have no default value for the memory when running verdi computer setup. It is !, which corresponds to the None as an internal value.

The `self.default_memory_per_machine` is an `ipw.IntText` where the value should be an int. I think there is no simple to use the `None` for the value, it must be an int initial number and allow the user to increase or decrease the number. 
One possible solution is using Text type and accept None as input.

unkcpz:
I use `-1` for not setting the default value and works fine in my test.

unkcpz:
Hi @yakutovicha, I still don't quite understand what you mean of we can set the `default_memory_per_machine` as `None`. Yes, when it pass to ComputerBuilder it can be `None` then the value just not set. 
But here the problem is more about we have the widget now I call it `default_memory_per_machine_gb` of type `ipw.IntText`, and it must has an int value. If I didn't set the value for the widget, the value will be `0` as default. 
Please have a look at my current implementation, if the value is `0` I regard it as not set.

unkcpz:
I think there is a more subtle issue with the aiida 2.x which will break the computer setup widget here. 
When a new computer setup from the `ComputerBuilder` API, a new user will be created that bind to the computer even if the email is the same, and this user cannot be shown by `verdi user list`. 
If the computer setup from the command line interface it will use the `default_user` for the computer being setup.

Pining @sphuber @chrisjsewell for input. Can you explain a bit about how the user creates and bind to the computer?

sphuber:
A `Computer` doesn't have a user. It is the `AuthInfo` class that has a `User` and is joined on a `Computer`. So when you run `verdi computer setup` a `Computer` instance is created that has no associated user. It is when you run `verdi computer configure` (which does `Computer.configure()` ) that an `AuthInfo` is created for the computer for a particular `User`.

unkcpz:
Thanks @sphuber! It works.

@yakutovicha, I fix the default user issue and rebase the PR, now the newly added code can all be show in the drowdown. Please make sure to rebase and add more commits to it.

danielhollas:
Hi @unkcpz,

I've been testing my app with the new `aiidalab/full-stack` images and this AWB branch, and I've ran into the same issue as you describe in https://github.com/aiidalab/aiidalab-qe/issues/279 with `'not persistent within this Session'.

As you noted elsewhere, it seems the problem is caused by multithreading. Have you tested various widgets here that use multithreading?

I am reusing some code from aiidalab-qe, and the error came from the `CalcJobOutputFollower`, https://github.com/aiidalab/aiidalab-qe/blob/379d20977c0e6fb85c8494fdc8f78cdcecb6ebc5/aiidalab_qe/widgets.py#L243.
I can see similar code in AWB widget `ProcessMonitor` in `aiidalab_widgets_base/process.py` so was wondering if widgets in that file work.

unkcpz:
Hi @danielhollas, sorry I just see the message somehow I didn't get GitHub notifications today. 
I think you are correct, the problem happened when two thread access aiida database. If the tread locks are implemented properly, it shouldn't be a problem. I'll take a close look at all these cases.

danielhollas:
Hi,

Yes, CalcJobOutputFollower is a straight copy form QeApp. I don't know how the current QeApp looks, but you should be able to trigger it if you click on a CalcJob node in a NodeTreeWidget. Let me know if you need more detailed info and I'll get to you when I am at my computer.

unkcpz:
Hi @danielhollas @yakutovicha, I think I have a clear view of how/why AiiDA 2.x is incompatible with the thread widgets. From my point of view, the key to solving it is to prevent using traitslet object being the aiida object gets from other threads. I think it has to be using Node UUID for those traitlets, I don't think there is another solution for it.

Can we have a meeting to settle it and I will work on changing those widgets first.

unkcpz:
I create a minimal example for it, using a list as the mutable object to mimic traitlets. 

```python
from aiida import orm
import threading

load_profile()

def _db_operation(mut: list):
    code = orm.load_code("doubler@localhost")
    mut.append(code)

def main():
    """Query the list of available codes."""

    mut = []
    thread = threading.Thread(target=_db_operation, args=(mut,))
    thread.start()
    thread.join()
    
    code = mut[0]
    print(code)

    
if __name__ == "__main__":
    main()
```

End up with `Instance '<DbNode at 0x7f714adcfdc0>' is not persistent within this Session`.

danielhollas:
Hi @unkcpz,

I am assuming that the refactor to use UUIDs for traitlets would be only needed for widgets that use threading? 

Thank you for the example, that's helpful. It would be great to see an example of a refactored widget to better understand the implications.

unkcpz:
> I am assuming that the refactor to use UUIDs for traitlets would be only needed for widgets that use threading?

Yes and no. The widget may not use thread but its methods can be triggered by other thread and update aiida object traitlets from that thread and cause the issue. It is more subtle to find and prevent.

But I think refactoring the widget using treading is the first step. And since we must use threading in those cases, changing the traitlets by using UUID is the only solution I think. 
If you and @yakutovicha have no objection, I'll rebase this PR and add a refactoring to those widgets.

danielhollas:
@unkcpz given the already long discussion here and many other changes, I would recommend landing this PR first, maybe to a temporary branch 'aiida2' for example, and then open a new PR with the threading changes against it.

yakutovicha (commented):
I have a little comment. Otherwise, good to go.

We should probably merge this and start releasing an alpha version of 2.0.0 to keep going.

The remaining tasks from a to-do list should probably go to separate issues.

danielhollas (approved):

.github/workflows/ci.yml
```diff
@@ -18,30 +18,36 @@ jobs:
...
         needs: [pre-commit]
 
-        runs-on: ubuntu-latest
-        timeout-minutes: 10
-
         strategy:
             matrix:
-                tag: [stable]
-                browser: [chrome, firefox]
+                tag: [edge]
```
danielhollas, line 31: The tag is not used in this file anymore.

.github/workflows/ci.yml
```diff
@@ -18,30 +18,36 @@ jobs:
...
               uses: actions/checkout@v2
 
-            - name: Test app
-              uses: aiidalab/aiidalab-test-app-action@v2
+            - name: Setup Python
+              uses: actions/setup-python@v2
               with:
-                  image: aiidalab/aiidalab-docker-stack:${{ matrix.tag }}
-                  browser: ${{ matrix.browser }}
+                  python-version: '3.9'
```
danielhollas, line 46: Maybe the Python version should be defined globally since it is at two places now.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -733,6 +767,11 @@ def __init__(self, **kwargs):
             style=STYLE,
         )
 
+        # Use double quotes to espace.
+        self.use_double_quotes = ipw.Checkbox(
+            value=False, description="Use double quotes"
```
danielhollas, line 772: Maybe add a tooltip with a more detailed explanation of what this is used for? Also the comment in the code should explain in more detail why the escaping is needed.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -904,8 +954,13 @@ def _observe_computer_setup(self, _=None):
             return
         if "setup" in self.computer_setup:
             for key, value in self.computer_setup["setup"].items():
-                if hasattr(self, key):
+                if key == "default_memory_per_machine":
+                    self.default_memory_per_machine_widget.value = int(
+                        value / 1024**2
```
danielhollas, line 959: I don't understand why the 1024 value is squared?

tests/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+            RMQHOST: messaging
+            TZ: Europe/Zurich
+            DOCKER_STACKS_JUPYTER_CMD: notebook
+            SETUP_DEFAULT_AIIDA_PROFILE: 'true'
+            AIIDALAB_DEFAULT_APPS: ''
+            JUPYTER_TOKEN: ${JUPYTER_TOKEN}
+        volumes:
+            - ..:/home/jovyan/apps/aiidalab-widgets-base
+        ports:
+            - 8998:8888
```
danielhollas, line 18: Any reason not to use `8888:8888`?

danielhollas (changes requested):
@unkcpz thanks so much for the tests, that's awesome that you've been able to make it work. :tada: 

Just a few minor comments / questions from me.

Since this is basically ready and has the tests, I would suggest to merge this one first, and build the uuid PR on top of it. I propose to have a temporary branch (`aiida2`) where we merge this and then we can thoroughly test on the uuid PR before merging to master. What do you think @yakutovicha

tests/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+---
+version: '3.4'
+
+services:
+
+    aiidalab:
+        image: aiidalab/full-stack:edge
```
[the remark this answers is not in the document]
unkcpz: You are correct, the latest is the "stable" one and should be used for other apps. `edge` is point to the latest commit.

.github/workflows/ci.yml
unkcpz: I keep it since going to use it as an ENV_VAR for using in the docker-compose. I'll change the docker-compose file.

tests/docker-compose.yml
unkcpz: Ha nice catch. Only personal reason since the `8888` of my workstation is occupied by some wired raid service. Will change it back to `8888`.

yakutovicha:
> Since this is basically ready and has the tests, I would suggest to merge this one first, and build the uuid PR on top of it. I propose to have a temporary branch (`aiida2`) where we merge this and then we can thoroughly test on the uuid PR before merging to master. What do you think @yakutovicha

I would merge those things directly to `master` and make only bug fixes in the [support/aiida-1.x](https://github.com/aiidalab/aiidalab-widgets-base/tree/support/aiida-1.x) branch. I would release 1.4.x versions from the support/aiida-1.x branch.

unkcpz:
> I would merge those things directly to master and make only bug fixes in the [support/aiida-1.x](https://github.com/aiidalab/aiidalab-widgets-base/tree/support/aiida-1.x) branch. I would release 1.4.x versions from the support/aiida-1.x branch.

I agree! 
The issue I mentioned for Node viewer is not dependent on the traitlets UUID change PR. I will fix the things mentioned in @danielhollas review and then should be good to go.

aiidalab_widgets_base/computational_resources.py
yakutovicha: The value in the database is in kB, but we agreed previously to keep GB for the widget. Just to answer why 1024 is squared.

I also noticed that this line is wrong, though. Thanks, @danielhollas, for pointing at it. The values shouldn't be an integer but a string. 

Possible solutions are:
```python
self.default_memory_per_machine_widget.value =
         str(decimal.Decimal(f"{value/1024**2:0.3f}").normalize()) + ' GB'
```
This requires importing `decimal`.

Another solution is simply to say that the imported value is in kB
```python
f"{value KB}
```

Any approach is good for me.

I am going to disappear, for some time, so I let @unkcpz and/or @danielhollas pick up an appropriate solution.

tests/docker-compose.yml
unkcpz: Hi @danielhollas, I am thinking maybe it is better to keep the host port other than `8888`, since `8888` is the default port for jupyter and high possibility that already been used by user. This docker-compose.yml will only be used for test purposes and is triggered by `pytest`, it more convenient if it does not conflict with `8888`.

aiidalab_widgets_base/computational_resources.py
unkcpz: I prefer the second one without the `decimal` dependency.

danielhollas (dismissed):
Thanks @unkcpz.

I am still not a fan of the current `default_memory_per_machine` widget. To me the fact that it mixes both number and text is super confusing, not to mention that we're introducing an extra dependency just to parse this string. When this field is not prefilled from the loaded computer, it is not clear what the units are, or even that one can specify the units.

Instead I would suggest to use two separate widgets, `ipw.IntText` (or `ipw.BoundedIntText`) for the numerical value itself, and `ipw.Dropdown` for specifying the units, perhaps backed by an Enum like this.

```python
class MemoryUnits(IntEnum):
    KB = 1024
    MB = 1024**2
    GB = 1024**3

self.memory_units = ipw.Dropdown(
    options=[(unit.name, unit) for unit in MemoryUnits],
    value=MemoryUnits.MB,
    description=""
)   
```

Nevertheless, I am approving so this can be done and iterated upon in a separate PR.

yakutovicha:
> I am still not a fan of the current `default_memory_per_machine` widget. To me the fact that it mixes both number and text is super confusing, not to mention that we're introducing an extra dependency just to parse this string. When this field is not prefilled from the loaded computer, it is not clear what the units are, or even that one can specify the units.

Since I am responsible for this one, I try to response.

From my perspective, the fact that the widget is human readable is rather a plus: one can essentially put any value here and make sure that things will work. If one writes something that can’t be parsed, the cross will appear making sure the user is doing something wrong.

if the field is not filled - it is not used. I added a placeholder that tries to explain that. Feel free to suggest a better one.

yakutovicha (dismissed):
Thanks @unkcpz 👍

.github/workflows/ci.yml
```diff
@@ -18,30 +18,44 @@ jobs:
...
+              uses: actions/setup-python@v2
+              with:
+                  python-version: ${{ matrix.python-version }}
+
+            - name: Install dependencies for test
+              run: |
+                  pip install -U -e .[tests]
+
+            - name: Run pytest
+              run: TAG=$(${{ matrix.tag }}) JUPYTER_TOKEN=$(openssl rand -hex 32) pytest --driver ${{ matrix.browser }}
```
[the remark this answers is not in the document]
unkcpz: True, changed. Can you approve it again?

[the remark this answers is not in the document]
unkcpz: Nevermind, since this is a tiny fix if the tests passed. I can force merge it anyway.

danielhollas (approved):

PR #354, opened 2022-09-07
Use UUID rather than the aiida node instance as traitlet

unkcpz:
From AiiDA 2.x every orm entity instance is now tied to a specific storage instance (which is connected to the database).
So once the storage instance is closed (and so close the connection), then those orm instances are now "dead" and can not be used for other reloaded sessions. 

Since this requires changing the API of the `ComputationalResourceWidget`, I create this dedicated PR for it. 
Do we also consider the back-ward compatibility or do we just leave this all to the 2.x support by put all these changes to #344 ?

### DIFF
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -4,13 +4,13 @@ import subprocess
 import threading
 from copy import copy
 from pathlib import Path
+from uuid import UUID
 
 import ipywidgets as ipw
 import pexpect
 import shortuuid
 import traitlets
 from aiida import common, orm, plugins
-from aiida.common.exceptions import NotExistent
 from aiida.orm.utils.builders.code import CodeBuilder
 from aiida.orm.utils.builders.computer import ComputerBuilder
 from aiida.transports.plugins.ssh import parse_sshconfig
@@ -207,10 +207,10 @@ class ComputationalResourcesWidget(ipw.VBox):
             return None
 
         try:
-            _ = orm.load_code(code_uuid)
-        except NotExistent:
-            self.output.value = f"""The supplied code UUID '<span style="color:red">{code_uuid}</span>'
-                was not found in the AiiDA database."""
+            _ = UUID(code_uuid, version=4)
+        except ValueError:
+            self.output.value = f"""'<span style="color:red">{code_uuid}</span>'
+            is not a valid UUID."""
         else:
             return code_uuid
 
@@ -1148,9 +1148,9 @@ class ComputerDropdownWidget(ipw.VBox):
             return None
 
         try:
-            _ = orm.load_computer(computer_uuid)
-        except NotExistent:
-            self.output.value = f"""The computer UUID '<span style="color:red">{computer_uuid}</span>'
-                supplied was not found in the AiiDA database."""
+            _ = UUID(computer_uuid, version=4)
+        except ValueError:
+            self.output.value = f"""'<span style="color:red">{computer_uuid}</span>'
+            is not a valid UUID."""
         else:
             return computer_uuid
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -27,13 +27,13 @@ class ComputationalResourcesWidget(ipw.VBox):
     """Code selection widget.
     Attributes:
 
-    value(Unicode or uuid of the code node): Trait that points to
-    the selected uuid of the code instance.
-    It can be set either to an AiiDA code uuid or to a code label.
+    value(Unicode or UUID of the code node): Trait that points to
+    the selected UUID of the code instance.
+    It can be set either to an AiiDA code UUID or to a code label.
     It is linked to the 'value' trait of
     the `self.code_select_dropdown` widget.
 
-    codes(Dict): Trait that contains a dictionary (label => Code uuid) for all
+    codes(Dict): Trait that contains a dictionary (label => Code UUID) for all
     codes found in the AiiDA database for the selected plugin. It is linked
     to the 'options' trait of the `self.code_select_dropdown` widget.
 
@@ -210,7 +210,7 @@ class ComputationalResourcesWidget(ipw.VBox):
             try:
                 _ = orm.load_code(code_uuid)
             except NotExistent:
-                self.output.value = f"""The code uuid '<span style="color:red">{code_uuid}</span>'
+                self.output.value = f"""The code UUID '<span style="color:red">{code_uuid}</span>'
                     supplied was not found in the AiiDA database."""
             else:
                 return code_uuid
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/computational_resources.py
```diff
@@ -199,24 +199,21 @@ def refresh(self, _=None):
...
-            if label in self.codes:
-                return code
-            self.output.value = f"""The code instance '<span style="color:red">{code}</span>'
-            supplied was not found in the AiiDA database."""
+        if isinstance(code_uuid, str):  # Check code by label.
+            try:
+                _ = orm.load_code(code_uuid)
+            except NotExistent:
+                self.output.value = f"""The code uuid '<span style="color:red">{code_uuid}</span>'
+                    supplied was not found in the AiiDA database."""
```
csadorf, line 214:
```suggestion
                self.output.value = f"""The supplied code UUID '<span style="color:red">{code_uuid}</span>'
                    was not found in the AiiDA database."""
```

aiidalab_widgets_base/computational_resources.py
```diff
@@ -199,24 +199,21 @@ def refresh(self, _=None):
     def _validate_value(self, change):
```
csadorf, line 199: I think we are misusing traitlet's validation mechanism here. The validation should not extend beyond basic checks IMO. In this case it would be a check whether the string is actually a valid UUID.

Then you can have a second trait, that actually contains the code which is None or set to a magic value in case that the requested code does not exist.

csadorf (commented):
I've started reviewing this and have a few suggestions and a few questions. I am a bit confused about how we use the validation mechanism here. Maybe we can clarify this before I move forward with further review.

> Do we also consider the back-ward compatibility or do we just leave this all to the 2.x support by put all these changes to https://github.com/aiidalab/aiidalab-widgets-base/pull/344 ?

I think we established that AWB that supports AiiDA 2.x will not support 1.x. In so far, this change set does not need to be backwards-compatible.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -26,12 +27,13 @@ class ComputationalResourcesWidget(ipw.VBox):
     """Code selection widget.
     Attributes:
 
-    value(Unicode or Code): Trait that points to the selected Code instance.
-    It can be set either to an AiiDA Code instance or to a code label (will automatically
-    be replaced by the corresponding Code instance). It is linked to the 'value' trait of
+    value(Unicode or uuid of the code node): Trait that points to
```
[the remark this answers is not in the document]
unkcpz: Changed.

aiidalab_widgets_base/computational_resources.py
unkcpz: Good point, I didn't check the logic carefully. I think in the validate function, we only need to check the UUID is valid value pointer to the node that exists in DB.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -197,29 +197,22 @@ def refresh(self, _=None):
...
-            label = self._full_code_label(code)
-            if label in self.codes:
-                return code
-            self.output.value = f"""The code instance '<span style="color:red">{code}</span>'
-            supplied was not found in the AiiDA database."""
-
-        # This place will never be reached, because the trait's type is checked.
-        return None
+        try:
+            _ = orm.load_code(code_uuid)
```
csadorf, line 210: To my knowledge, the widget will be blocked while the validation is running. Therefore, I don't think we should run code as part of the validation that might require a significant amount of time (like a database query). I think the validation should not extend beyond very basic checks in the sense of whether the widget will enter a valid state with the provided value.

Anything beyond that should be handled as part of the business logic of the application. That means, we should inform consumers that the selected value does not correspond to a valid code and then allow those consumers to react to that in an appropriate manner.

unkcpz: I see, thanks! The UUID is got from an instance queried from the database, to be honest, I see less point to check its existence. 
Changed and please check the last commit.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -207,10 +207,10 @@ def _validate_value(self, change):
...
 
         try:
-            _ = orm.load_code(code_uuid)
-        except NotExistent:
-            self.output.value = f"""The supplied code UUID '<span style="color:red">{code_uuid}</span>'
-                was not found in the AiiDA database."""
+            _ = UUID(code_uuid, version=4)
+        except ValueError:
+            self.output.value = f"""'<span style="color:red">{code_uuid}</span>'
+            is not a valid UUID."""
```
csadorf, line 213: I think the standard behavior is to raise a `TraitError` in this case, see [here](https://traitlets.readthedocs.io/en/stable/using_traitlets.html#validation-and-coercion).

danielhollas:
> I agree. But this means we need to care about the release. I suggest cherry pick and release a version without this feature for 1.x supported only and after that having this new API a new version released with PR https://github.com/aiidalab/aiidalab-widgets-base/pull/344.

Yes please :blush: I asked @yakutovicha to cut a new 1.x release now to release my recent PRs so you might want to hold on merging this till then maybe (or have a separate 2.x branch).

aiidalab_widgets_base/computational_resources.py
unkcpz: No, this is the exception raised when the `code_uuid` is not a valid UUID. `uuid.UUID` module will raise a `ValueError`
For example:
```python
from uuid import UUID

UUID("IAMNOTUUID", version=4)
```

yakutovicha:
> I agree. But this means we need to care about the release. I suggest cherry pick and release a version without this feature for 1.x supported only and after that having this new API a new version released with PR #344.

As a side note: I just created [support/aiida-1.x](https://github.com/aiidalab/aiidalab-widgets-base/tree/support/aiida-1.x) branch. All the bug fixes for the aiida-1.x should go there. You can now feel free to brake things in the master branch 😄 .

aiidalab_widgets_base/computational_resources.py
csadorf: Yes, but **we** should raise a `TraitError` in response.

yakutovicha (commented):
Just to mention that I am a bit puzzled by the fact that an AiiDA object cannot be a trait. There are many widgets in this repository relying on AiiDA objects being traits. 

I am still missing why the error pops up for the `ComputationalResourcesWidget` and doesn't manifest itself for the `StructureManagerWidget` (as we tested with @unkcpz).

Before approving this I would really like to understand:

1. Should we not use AiiDA objects as traits?
2. If we shouldn't: why is that?
3. If we still can, what are the new restrictions?

Can @chrisjsewell and/or @unkcpz comment on this?

unkcpz:
Hi @yakutovicha, is this possible related to `self.hold_trait_notifications()`? I don't know how this contextmanager work under the hood, but from its functionality, it will run the code (where inside `refresh` function it will query for the aiida data for the code/computer node) and hold the notification. I guess the issue comes from where calls the update of this widget is a new session and the error pop up. 

I make the above assumption since when testing the `ComputationalResourcesWidget` itself from the notebook all works fine, the issue occurs only when embedding the widget into qeapp.

yakutovicha:
> Let me know how to test this further. If `hold_trait_notifications` is the catch, I'd say we should not use AiiDA objects as traits in all cases.

I don't know how to test it at the moment. I need to look closer. But I would very much prefer to keep using AiiDA objects as traits, cause otherwise, we will need a huge rewrite of the AWB library.

unkcpz:
As discussed offline, the issue reported in https://github.com/aiidalab/aiidalab-qe/issues/279 is rooted in using unsafe multiple threads calls to access the AiiDA database which is not well handled by AiiDA at the moment. 
I'll close this PR and fix the thread issue in QeApp.

unkcpz:
Sure! Deleted.

PR #359, opened 2022-10-07
fix bug when removing atoms 

cpignedoli:
When deleting  atoms the selection should be set to an empty list otherwise the visualizer may crash


yakutovicha (approved):
Good catch, thanks a lot @cpignedoli!

Issue #367, opened 2022-10-14
Cell Editor: enable surface cleavage.


yakutovicha:

superstar54:
Sure, I will implement this.

yakutovicha:
As far as I can tell this is not yet fixed.

superstar54:
> As far as I can tell this is not yet fixed.

I agree. In a demo for aiidalab-qe structure editor plugin,  I did a test to cut the surface using [ase.build.surface](https://wiki.fysik.dtu.dk/ase/ase/build/surface.html#ase.build.surface). It works well, and is in the demo [repo](https://github.com/aiidalab/aiidalab-qe-plugin-demos/blob/main/aiidalab_qe_plugin_demos/surface/editor.py). I may make a PR to AWB when it's polished.

PR #372, opened 2022-10-19
Enhance cell editor

### LINKED ISSUE #365: support building supercell.
Bring up by @superstar54 from https://github.com/aiidalab/aiidalab-widgets-base/pull/327#issuecomment-1275101340

> Cell Transformation is a more general concept. If we use this term, it is better to include the general transformation matrix. > Please have a look at make_supercell method here. 
> https://wiki.fysik.dtu.dk/ase/ase/build/tools.html#ase.build.make_supercell

superstar54:
fixes #365
- editing cell vectors directly.
- cell transformation using a 3x4 transformation matrix.

Screenshot from 2022-10-19 08-03-10


unkcpz:
@superstar54 Thanks! 
That "Cell vectors" and "Cell transformation" duplicate each other? Why not just keep the "Cell transformation"?
I would suggest adding instructions on how the matrix apply to the cell.

superstar54:
> That "Cell vectors" and "Cell transformation" duplicate each other?

No. The "Cell vectors" are the cell itself (3x3 array). This enables the user to edit the cell directly, e.g. cell[2][2] += 10, which will add a vacuum in the z-direction, making a surface slab.
The "Cell transformation" transforms the cell by multiplying a (3x4) matrix.

> I would suggest adding instructions on how the matrix apply to the cell.

Agree. I will add a tooltip for every button.

unkcpz:
> This enables the user to edit the cell directly, e.g. cell[2][2] += 10, which will add a vacuum in the z-direction, making a surface slab.

I see, thanks for the explanation, this is interesting. But don't you think it is mind-binding to set the off-diagonal for changing the cell in this way? 

> Agree. I will add a tooltip for every button.

I think that is not enough, my suggestion is for each matrix you keep the matrix on the left and on the right have a paragraph for the instruction, just like image

If the special symbols are hard to add, we can leave a text paragraph for future polish.

superstar54:
> keep the matrix on the left and on the right have a paragraph for the instruction, just like

Will this make too much text on the GUI? I think a detailed explanation would be better put in the documentation.

yakutovicha:
@cpignedoli, could you please also have a look/test this PR. Just to make sure that it is implemented according to the user needs.

cpignedoli:
Great idea @superstar54 , I had requests to implement a cell editor to create surfaces but still had no time to go in this direction.
Happy to see that things started.

I have a few comments from my experience with the simulation of surfaces:
typically I start from a problem defined, for example, as "simulate a molecule  on a stepped ( e.g. Au(11 11 12) ) gold surface

Or simulate the S1(110) surface.

What I do is
1)I start from a bulk and scale the bulk to the desired lattice parameter (will I use the exp lattice parameter, or the one compatible with my DFT settings, or, am I studying a stressed crystal,..)

2)I enter the cleavage plane in the editor (e.g. 1, 1 ,1  or 11, 11, 12) and select a thickness for the slab (this is definitely not intuitive with high indexes, you need the graphical support..), and the amount of vacuum I want to add (can also be 0 if I want to simulate a bulk in "a surface cell" (e.g. to compute projected bulk band structures) and I click the "cleavage button"

3) I adjust the thickness to the desired one

4)I "orient to standard" so that  a1 coincides with x, a2 is in the xy plane (or I select a different standard if needed)

5) At this point in many cases (e.g. (111) of a cubic crystal that would produce a hexagonal cell) I redefine the lattice as a linear combination of the primitive 2D vectors to obtain for example a rectangular cell (e.g. I typically work not with the hexagonal Au(111) but with (a1 +a2 , a2-a1) that is rectangular

6)I reorient to standard

7) I replicate as desired

All of this without deriving myself the transformation matrices because for sure I would do hundreds of mistakes.

I hope my comments can be useful. We can also have a zoom call to see more examples




image (1)

AndresOrtegaGuerrero:
@yakutovicha  Should we merge this PR , so the QE app can have the supercell and cell parameters editor?

unkcpz:
Can anyone remind me what blocks this PR? Please add a comment.

unkcpz:
Hi @yakutovicha @cpignedoli, any update on this PR? Can we get only the supercell functionality merged and if there are more you want to add to this editor, we leave it to the future PRs?

yakutovicha:
> Hi @yakutovicha @cpignedoli, any update on this PR? Can we get only the supercell functionality merged and if there are more you want to add to this editor, we leave it to the future PRs?

We briefly discussed this with PR @cpignedoli. We will give it a look this afternoon.

cpignedoli (commented):
I think the editor is great but difficult to use for a user not expert. It should be merged but we should afterwards add few features:

1. axes.. this is a viewer problem but becomes crucial here
2. HTML help text with e.g. an example on how you get a 111 surface from a gold conventional cell (or, how do I get Au(788) )
3. button to reset the matrix to  1 1 1
4. the matrix should be probably reset when we import a new structure (but teh button above is enough)

There is one issue that in my opinion should be implemented before merging:
when changing the cell vectors I need the **checkbox "scale coordiantes"**

- checkbox marked: I use this e.g. to change lattice parameter of my bulk (I uploaded a structure that has exp lattice parameter I want to scale it to PBE lattice parameter)
- checkbox not marked: I want to set to 40 the z axis to add vacuum, the atomic positions should not be affected

unkcpz:
thanks @cpignedoli. 

@superstar54 please rebase this PR with the master branch and see if you can address the points mentioned by @cpignedoli. Thanks. As we discussed, this feature is better included in the next release.

superstar54:
Thanks @cpignedoli, for the suggestions. I have implemented:


- [x] button to reset the matrix to 1 1 1
- [x] the matrix should be probably reset when we import a new structure
- [x] checkbox to scale atoms positions
- [x]  unit test


For the other two comments:

> axes.. this is a viewer problem but becomes crucial here

I agree. We should address this in another PR.

> HTML help text with e.g. an example on how you get a 111 surface from a gold conventional cell (or, how do I get Au(788) )

Note, that this PR is only used to edit the cell parameters, or make supercell, not to cut the surface.

cpignedoli (approved):

superstar54:
HI@yakutovicha would you like to review this, or should I proceed with merging it?

yakutovicha:
> Hi @yakutovicha would you like to review this, or should I proceed with merging it?

Let me give this PR a quick look today, just to be sure. It will be quick.

yakutovicha:
> I agree. We should address this in another PR.
> 
> >

I made an issue for this #532

yakutovicha:
In that 

> Note, that this PR is only used to edit the cell parameters, or make supercell, not to cut the surface.

I guess in this case we should modify the comment at the top, which says

> Add the following features (https://github.com/aiidalab/aiidalab-widgets-base/issues/367, https://github.com/aiidalab/aiidalab-widgets-base/issues/365)

to prevent an accidental issue close

aiidalab_widgets_base/structures.py
```diff
@@ -943,6 +1018,75 @@ def _to_standard_cell(
...
+            cell_parameters = change["new"].cell.cellpar()
+            for i in range(6):
+                self.cell_parameters.children[i].children[1].value = round(
+                    cell_parameters[i], 4
+                )
+        else:
+            for i in range(6):
+                self.cell_parameters.children[i].children[1].value = 0
+        # reset transformation matrix
+        self.reset_cell_transformation_matrix()
```
yakutovicha, line 1034:
```suggestion
```
I suggest not resetting the matrix automatically. The user will lose the information they entered.

I think it is safer to do that only when pressing the button.

aiidalab_widgets_base/structures.py
```diff
@@ -943,6 +1018,75 @@ def _to_standard_cell(
...
+
+    @_register_structure
+    def apply_cell_parameters(self, _=None, atoms=None):
```
yakutovicha, line 1037: Please make this method private:
```suggestion
    def _apply_cell_parameters(self, _=None, atoms=None):
```

aiidalab_widgets_base/structures.py
```diff
@@ -943,6 +1018,75 @@ def _to_standard_cell(
...
+        ]
+        if atoms is not None:
+            atoms.set_cell(
+                Cell.fromcellpar(cell_parameters),
+                scale_atoms=self.scale_atoms_position.value,
+            )
+            self.structure = atoms
+
+    @_register_structure
+    def apply_cell_transformation(self, _=None, atoms=None):
```
yakutovicha, line 1053: Private
```suggestion
    def _apply_cell_transformation(self, _=None, atoms=None):
```

aiidalab_widgets_base/structures.py
```diff
@@ -943,6 +1018,75 @@ def _to_standard_cell(
...
+            """.format(
+                    e
+                )
+                return
+            # translate
+            atoms.translate(-atoms.cell.array.dot(translate))
+            self.structure = atoms
+
+    @_register_structure
+    def reset_cell_transformation_matrix(self, _=None, atoms=None):
```
yakutovicha, line 1083: Private
```suggestion
    def _reset_cell_transformation_matrix(self, _=None, atoms=None):
```

aiidalab_widgets_base/structures.py
```diff
@@ -943,6 +1018,75 @@ def _to_standard_cell(
...
+        else:
+            for i in range(6):
+                self.cell_parameters.children[i].children[1].value = 0
+        # reset transformation matrix
+        self.reset_cell_transformation_matrix()
+
+    @_register_structure
+    def apply_cell_parameters(self, _=None, atoms=None):
+        """Apply the cell parameters to the structure."""
+        from ase.cell import Cell
```
yakutovicha, line 1039: not needed (see below)
```suggestion
```

aiidalab_widgets_base/structures.py
```diff
@@ -943,6 +1018,75 @@ def _to_standard_cell(
...
+
+        # only update structure when atoms is not None.
+        cell_parameters = [
+            self.cell_parameters.children[i].children[1].value for i in range(6)
+        ]
+        if atoms is not None:
+            atoms.set_cell(
+                Cell.fromcellpar(cell_parameters),
```
yakutovicha, line 1047:
```suggestion
                ase.cell.Cell.fromcellpar(cell_parameters),
```

yakutovicha (changes requested):
thanks @superstar54.

I requested a bunch of trivial changes, once done - good to go from my side 👍

aiidalab_widgets_base/structures.py
```diff
@@ -895,7 +895,53 @@ def __init__(self, title="Cell transform"):
...
+                        ),
+                        ipw.FloatText(
+                            value=cell_parameters[i], layout={"width": "100px"}
+                        ),
+                    ]
+                )
+                for i in range(6)
+            ]
+        )
+        #
```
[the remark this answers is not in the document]
superstar54: I have added a comment.

aiidalab_widgets_base/structures.py
superstar54: I changed as you suggested.

superstar54:
Hi @yakutovicha , thanks for your review. I've incorporated the changes based on your suggestions.

yakutovicha (approved):
LGTM, thanks @superstar54!

Issue #342, opened 2022-08-18
Better design of `_StructureDataBaseViewer` and `StructureDataViewer`


yakutovicha:
As a developer, one would like to have a clear separation of concern between `_StructureDataBaseViewer` and `StructureDataViewer` allowing for easy further improvements and modifications.

The base class is meant to provide a common functionality of structure viewer and (yet to come) trajectory viewer. Unfortunately, due to some design flows it cannot be easily done with the current implementation.

Herein, after a meeting with @unkcpz and @danielhollas, we agreed to have a closer look at those two classes and come up with a better design.

## Known issues

1. Both `structure` and `displayed_structure` traits are defined in the `StructureDataViewer` class but used in `_StructureDataBaseViewer`.
2. Point 1. also makes it impossible to use  `_StructureDataBaseViewer` as the base class for the trajectory viewer.
3. An addition of new displayed properties requires new traits to `_StructureDataBaseViewer`, which is not optimal as most of the features are stored in the `structure` trait.

yakutovicha:
Overview of the `_StructureDataBaseViewer`:

```python
class _StructureDataBaseViewer(ipw.VBox):
    """Base viewer class for AiiDA structure or trajectory objects.

    :param configure_view: If True, add configuration tabs (deprecated)
    :type configure_view: bool
    :param configuration_tabs: List of configuration tabs (default: ["Selection", "Appearance", "Cell", "Download"])
    :type configure_view: list
    :param default_camera: default camera (orthographic|perspective), can be changed in the Appearance tab
    :type default_camera: string

    """

    selection = List(Int)
    selection_adv = Unicode()
    supercell = List(Int)
    cell = Instance(Cell, allow_none=True)
    DEFAULT_SELECTION_OPACITY = 0.2
    DEFAULT_SELECTION_RADIUS = 6
    DEFAULT_SELECTION_COLOR = "green"

    def __init__(self, ...):

    def _selection_tab(self):  # Defines a tab with the selection options.

    def _appearance_tab(self):  # Defines a tab which manages appearances.

    def _cell_tab(self):  # Defines a tab which displays cell info

    @observe("cell")
    def _observe_cell(self, _=None):  # Observes cell modifications and updates the displayed cell parameters.

    def _download_tab(self):  # Defines a tab with download options.

    def _render_structure(self, change=None):  # Function to render structure with povray.

    def _on_atom_click(self, _=None):  # Update selection when clicking on atom

[...]

PR #373, opened 2022-10-19
Implement viewer representations

### LINKED ISSUE #366: support different model styles for the structure, e.g. Sphere, Ball-and-Stick
I am working on it migrating what we did for the empa_viewer, I will need the help of @yakutovicha  to finalize it

cpignedoli:
fixes #366 

We have to take an important design decision before we continue on this:
I tried to have some concepts inherited from other viewers, especially:
if I have a structure and I define for it a couple of representations (e.g. a slab with VDW radii and a molecule on the slab as ball+stick) it is crucial that when I modify the structure (e.g. removing some atoms) my representations remain active and properly working.

This requires that in the editors we keep track of the defined representations, 

https://github.com/cpignedoli/aiidalab-widgets-base/blob/3d642d16cb291c308d0614d1f302d7885b4cf723/aiidalab_widgets_base/structures.py#L1516

not a big deal but an important addition one should not forget. Do we agree on this? is there a better design for this concept?
(the code above avoid, e.g., taht if I rmeove atoms form a structure (see image) teh other atoms will lose theyr correct representation)
image (14)

Also, I would like to avoid redrawing a representation unnecessarily but this seems to be difficult...

At the moment update of the viewer is triggered:
-if a structure is imported
-if a structure is modified
-if a user changes the representations and clicks on "apply representation"


superstar54:
@cpignedoli Thanks for implementing this. The concept is OK to me.

danielhollas:
Hi @cpignedoli,

Thanks for the PR and work. I think all of this also ties into our previous discussing with @yakutovicha and @unkcpz about the need to refactor the StructureViewer classes, please take a look at the issue #342. I am afraid that we should first solve that issue, because here you're adding a lot of new code to an already overloaded class, which would make it even harder to refactor later.

I would also like to raise the question about the scope of this widget. Do we really need to support different representations for different part of the system? That seems to me like a big complication. Are we trying to essentially implement a full-fledged molecular editor here, or should we limit ourselves to make things simple. It would help if you could write a bit more about your use case.

By the way, would you mind providing some screenshots of how the interface looks like right now?

In general, it definitely seems weird to have editors care about the representation.

cpignedoli:
Hi @danielhollas 
I fully see your point and I had discussions with @yakutovicha who shares your concerns.

From the perspective of a user, I can imagine a viewer without an editor (for example VMD, an excellent viewer that is not really meant for editing) but I cannot imagine an editor without a viewer, you should see what you edit and in many cases, you do not want to edit parts of a system that you decided to hide.

This creates necessarily a two-way communication between the editor and the viewer.

The most simple implementation I am able to imagine for this (not the current implementation) is that the viewer and the editors work on e.g. an ase geometry that has an additional channel (e.g. an integer but only 'tag' would be available) indicating to which representation an atom belongs. The viewer knows what to do out of that integer.

As soon as a geometry is imported a default value is assigned for the representation, if representations are created the 'tags' are modified

Coming back to why as a user I am strongly in favor of having the possibility of multiple representations;
I think I never managed to work without:
in the snapshot below you can see that already the simple task of recognizing the stacking mode in a surface reconstruction (VMD example for reconstructed Au(111) requires hiding some layers and highlighting others in different ways.

If it comes to modifying a geometry before running a simulation, I provide an example of a molecule on gold:
The standard AiiDAlab viewer offers a perspective that is almost useless I even cannot identify easily an adatom that I may want to move below a phenyl ring.

Having two representations enhances enormously the possibility to edit promptly the structure.

Not to speak about simulaitons where you wan tto partition a system in a QM and an MM part, it oculd well be you need to hide or simplyfy teh representation of the atom sbelonging to the MM part.

Having two representations enhances enormously the possibility to edit promptly the structure.

In terms of complexity, let me recall that, at the viewer level (with nglview), dealing with one or several representations means
going back and forth between two dictionaries

`def _gen_translation_indexes(self):`
`     """Transfromation of indexes in case of multiple representations`
`     dictionaries for  back and forth transformations.`
`      suppose we have 3 representations:`
`      component = 0,1,2`
`      and a structure with 10 atoms.`
`      If we assign the first 3 atoms to the first representation`
`      atom 4 to the second and the rest to the third`
`      the two dictionaries will look like:`
`      {0:(0,0),1:(0,1),2:(0,2),3:(1,0),4:(2,0),5:(2,1),6:(2,2),7:(2,3),8:(2,4),9:(2,5)}`
`      and`
`      {(0,0):0,(0,1):1,(0,2):2,(1,0):3,(2,0):4,(2,1):5,(2,2):6,(2,3):7,(2,4):8,(2,5):9}`
`"""`

I am available for a zoom call with all of you to discuss further a possible route to create a meaningful viewer.

P.S. the version I drafted (at least before the latest bugs I introduced) handles supercells (from the appearance tab) and if an atom from a replica is selected, the corresponding atom in the "principal unit" is highlighted. Distances are computed within the unit not within replicas. This of course could be changed in favor of MIC convention


examples

yakutovicha:
We discussed this with @cpignedoli and possibly came up with a better solution. It should satisfy both: the ability to represent atoms differently (as expressed by @cpignedoli) and our strive for simplicity (as highlighted by @danielhollas).

It is true, that passing representation to the editors is a direct violation of the separation of concerns concept. So I am fully on @danielhollas's side here.

However, there is one trick that we can do to avoid explicitly passing this information to the editors: store it internally in the `ase.Atoms` object as a `representation` array. Here is how this can be achieved:

```python
from ase.build import molecule
import numpy as np
m = molecule('H2O')
m.cell = [5,5,5]
m.set_array('representations', np.array([0, 1, 0]))  # <-- each integer corresponds to a different representation.
m.get_array('representations')
```

The advantage of this approach is that the `representation` array is correctly handled along all the transformations:

Screenshot 2022-10-20 at 15 07 33

The only disadvantage of the current approach is that it doesn't handle same atom in different representations. But this could be resolved by creating a separate array for every representation, something like `representation_0`, `representation_1`, etc...

We agreed that @cpignedoli will further develop this concept.

aiidalab_widgets_base/viewers.py
```diff
@@ -156,6 +163,61 @@ def __init__(self, parameter, downloadable=True, **kwargs):
         super().__init__([self.widget], **kwargs)
 
 
+class Representation(ipw.VBox):
```
yakutovicha, line 166: Make a more concrete name. It is too generic to call it `Representation`.

aiidalab_widgets_base/viewers.py
```diff
@@ -335,10 +396,156 @@ def change_camera(change):
         center_button = ipw.Button(description="Center molecule")
         center_button.on_click(lambda c: self._viewer.center())
 
+        # 5. representations buttons
+        self.atoms_not_represented = ipw.Output()
```
yakutovicha, line 400: Make it `ipw.HTML`.

aiidalab_widgets_base/viewers.py
```diff
@@ -335,10 +396,156 @@ def change_camera(change):
...
+
+        self.all_representations = (
+            self.all_representations[:index] + self.all_representations[index + 1 :]
+        )
+        del representation
+        self.n_all_representations -= 1
+        self.apply_representations()
+
+    @observe("all_representations")
+    def _observe_representations(self, change):
```
yakutovicha, line 441:
```suggestion
    def _observe_all_representations(self, change):
```

aiidalab_widgets_base/viewers.py
```diff
@@ -625,41 +832,124 @@ def _render_structure(self, change=None):
...
+                    -com[2],
+                    1,
+                ]
+                self._viewer._set_camera_orientation(top_z_orientation)
+                self._viewer.center()
+                return False
+            else:
+                return True
+        except AttributeError:
+            return True
```
yakutovicha, line 952: Move to another PR.

aiidalab_widgets_base/structures.py
```diff
@@ -97,7 +111,7 @@ def __init__(
             self.viewer = viewer
         else:
             self.viewer = StructureDataViewer(**kwargs)
-        dlink((self, "structure_node"), (self.viewer, "structure"))
+        link((self, "structure"), (self.viewer, "structure"))
```
yakutovicha, line 114: Please move back to `dlink` and test that things work.

yakutovicha (commented):
Thanks, @cpignedoli. Some preliminary suggestions.

cpignedoli:
Please @yakutovicha @danielhollas @superstar54 give it a try.
I removed multiple ngl components and work only with representations on component_0
This simply things but ngl is not able to do ball+stick if a component has too many atoms.
Also, components can be switched between view/hide for representations this does not work and I did a workaround.

On top of this, we can implement https://github.com/aiidalab/aiidalab-widgets-base/issues/342

Next will be to include more representations (or to open to the user more details
). Also, @yakutovicha suggested computing bonds independently from ngl and then passing them to ngl as additional shapes. This can be done.

aiidalab_widgets_base/structures.py
cpignedoli: done

aiidalab_widgets_base/viewers.py
cpignedoli: done

aiidalab_widgets_base/viewers.py
cpignedoli: done

aiidalab_widgets_base/viewers.py
cpignedoli: done

aiidalab_widgets_base/viewers.py
cpignedoli: done

cpignedoli (commented):
Many things got lost:
-highlighting is almost invisible
-a bug when deleting atoms has been reintroduced
-after copying /adding atoms the representations are not updated correctly and atoms are invisible

In the original version when copying /adding atoms the new atoms were assigned automatically to the representation from where they originated and the selection was updated to the new atoms.
Automatic update of selection was inhibited in case of deleting atoms otherwise the viewer could crash

yakutovicha:
@cpignedoli's comment on the last commit:

Now copying atoms works: the new atoms are visualized and assigned to the correct representation but adding a new group does not work:
the atoms are not assigned to a representation and only by clicking on "apply rep" you get the added atoms listed as non represented

CRASH:
-if a non-existing index is assigned to a representation
-I had a crash with undo but I was not able to reproduce it

superstar54:
@cpignedoli @yakutovicha Thanks for the work. I tested the widget. Here is some feedback.
I used SiO2 for the test.
- when I set supercell 5x5x5 (750 atoms), the bond does not show anymore. Is there a limit to the number of atoms? 
- Better to align the text as marked in the image.
- Better to find a way to separate the `presentation` part from `Camera type` and `Center molecule` buttons.

yakutovicha:
Thanks for the comment, @superstar54 

> * when I set supercell 5x5x5 (750 atoms), the bond does not show anymore. Is there a limit to the number of atoms?

That is expected, there is a follow-up PR on bonds #377 

> * Better to align the text as marked in the image.
> * Better to find a way to separate the `representation` part from `Camera type` and `Center molecule` buttons.

Could you maybe sketch how you think things should be placed? I then implement that.

superstar54:
I suggest moving the `Center molecule` above and changing the representation section to an accordion (if possible).

yakutovicha:
This PR is ready for review. The tests with Python 3.10 are failing for an unknown reason, but it has nothing to do with this PR.

Please have a look @cpignedoli, @superstar54, @danielhollas.

danielhollas:
I plan to have a look, but it would be great if @cpignedoli and @superstar54 went through it first, since I do not really plan to use this feature.

@yakutovicha please merge with master which has now cleaned up all the warning from unit tests. We are also now turning all warnings to errors to prevent them croping up again (and I'll note that a bunch of them were actually minor bugs).

unkcpz:
After rebasing with `master`, this PR ad #511 both has a weird exception raised but not failing the unit tests. I had not idea where this comes. Does this only appear after #509? 

```
Exception ignored in: <_io.FileIO name=14 mode='rb' closefd=True>
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/_pytest/unraisableexception.py", line 59, in __exit__
    del self.unraisable
ResourceWarning: unclosed file <_io.BufferedReader name=14>
Exception ignored in: <_io.FileIO name=16 mode='rb' closefd=True>
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/_pytest/unraisableexception.py", line 59, in __exit__
    del self.unraisable
ResourceWarning: unclosed file <_io.BufferedReader name=16>
```

danielhollas:
@unkcpz yeah, I should have mentioned that. For some reason these exceptions pop up when I turned the warnings into errors. I could not figure out where it is coming from. Would definitely be good to solve eventually since it might indicate some resource leak, but it also can be just something connected to the tests and the fixtures that we use.

(Maybe the DB connection is not closed properly?)

unkcpz:
Okay, then I am not worried. I guess the same might be caused from the `conftest.py`

superstar54:
Thanks @cpignedoli @yakutovicha . I did a test using a surface slab, it works great. here is a screenshot.
Screenshot from 2023-10-18 17-01-15


Few comments:
- The meaning of `Size` is not clear.
- Better to add tests for different types (e.g. 'spacefill' and 'ball-stick')

yakutovicha:
> * Better to add tests for different types (e.g. 'spacefill' and 'ball-stick')

done in 13f9254

cpignedoli:
works.. maybe I tested an older verison

aiidalab_widgets_base/computational_resources.py
```diff
@@ -196,7 +196,7 @@ def refresh(self, _=None):
         with self.hold_trait_notifications():
             self.code_select_dropdown.options = self._get_codes()
             if not self.code_select_dropdown.options:
-                self.output.value = f"No codes found for default calcjob plugin '{self.default_calc_job_plugin}'."
+                self.output.value = f"No codes found for default calcjob plugin {self.default_calc_job_plugin!r}."
```
unkcpz, line 199: What is this `!r` for?

tests/test_structures.py
```diff
@@ -17,8 +17,8 @@ def test_structure_manager_widget(structure_data_object):
         editors=[
             awb.BasicStructureEditor(title="Basic Editor"),
         ],
+        input_structure=structure_data_object,
```
danielhollas, line 20: @yakutovicha the failing test indicates (thanks to pytest warning turned into error) that this change is not correct, probably `input_structure` cannot be passed in the constructor?

I am planning to go through the code today (visual testing has been done by others already)

aiidalab_widgets_base/computational_resources.py
yakutovicha: Chat GPT gave a very complete answer 😄 :

In the provided f-string in Python, !r is used to format the value of self.default_calc_job_plugin using the repr() function.

The !r conversion flag is used within f-strings to represent a value in a way that it can be used as valid Python code to recreate the object. It is often used for debugging or when you want to display the exact representation of an object, including any quotes, escape characters, or special characters it might contain.

For example, if self.default_calc_job_plugin contains a string "example", using !r would format it as 'example', including the single quotes. This can be helpful for making sure you see the exact content of the variable, especially if it includes characters that might be special in Python, like newline characters or escape sequences.

In your specific f-string, it's being used to format the value of self.default_calc_job_plugin with single quotes around it, so the resulting string will look like this:

"No codes found for default calcjob plugin 'value_of_self.default_calc_job_plugin'."

This can be useful for debugging or providing informative error messages.

tests/test_structures.py
yakutovicha: Thanks, @danielhollas, I will have a closer look!

danielhollas (changes requested):
@yakutovicha I made the first pass on the review (lot of code!). Generally looks great!
In this round I mostly have questions about the code and asks for better naming of variables and more comments / docstrings / types.

Thank you so much for extending the tests, although I have not reviewed them yet.
It's a pity we don't have Codecov setup on this repo yet, right now it's impossible to know whether the tests cover everything. This should ideally be done soon.

aiidalab_widgets_base/structures.py
```diff
@@ -14,7 +13,7 @@
 from aiida import engine, orm, plugins
 
 # Local imports
-from .data import LigandSelectorWidget
+from .data import FunctionalGroupSelectorWidget
```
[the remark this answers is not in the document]
yakutovicha: The old name is deprecated, so I would consider it relevant.

aiidalab_widgets_base/structures.py
```diff
@@ -85,7 +84,7 @@ def __init__(
             self.viewer = viewer
         else:
             self.viewer = StructureDataViewer(**kwargs)
-        tl.dlink((self, "structure_node"), (self.viewer, "structure"))
+        tl.dlink((self, "structure"), (self.viewer, "structure"))
```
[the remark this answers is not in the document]
yakutovicha: It was a bug to link `structure_node`. The `structure` trait is supposed to contain the current state of things, while `structure_node` is an unstored AiiDA object.

In the previous (buggy) version the AiiDA node would then be converted back to an ASE object losing loads of information.

unkcpz:
> It's a pity we don't have Codecov setup on this repo yet

Hi @danielhollas, we have this in this repo it will report after all tests passed. Or am I missing your point?

yakutovicha:
Thanks for the review, @danielhollas, I am going through your comments. There is one thing that I wanted to let you know first:

> It's a pity we don't have Codecov setup on this repo yet, right now it's impossible to know whether the tests cover everything.

We do have it. The comment from codecov is hidden among other comments. You can see the report [here](https://app.codecov.io/gh/aiidalab/aiidalab-widgets-base/pull/373).

aiidalab_widgets_base/structures.py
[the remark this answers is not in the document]
yakutovicha: Maybe it is a good idea to make the `structure_node` trait immutable because it is not meant to be modified externally. Its job is literally to be the StructureNode object corresponding to the `structure` trait (ASE).

> Hmm, but I think there were some design decisions involved in that.

And, again, it was a mistake (probably mine) to link the `structure_node` trait instead of the `structure`.

tests/test_structures.py
```diff
@@ -41,20 +45,39 @@ def test_structure_manager_widget(structure_data_object):
...
         ],
+        viewer=awb.viewers.StructureDataViewer(),
+        node_class="StructureData",
     )
 
     assert structure_manager_widget.structure is None
+    structure_manager_widget.input_structure = stored_structure
+
+    # Set action vector perpendicular to screen.
+    base_editor.def_perpendicular_to_screen()
```
unkcpz, line 80: I assume this changes the axis, would be better to add an assert for the effect.

unkcpz (commented):
Hi @yakutovicha , I just go through https://github.com/aiidalab/aiidalab-widgets-base/pull/373, from code (and data flow :laughing: ) point of view I have to admit there is nothing to nitpick, it is quite clear what the PR wants to achieve with nice unit tests. I nitpick on one unit test to add an assertion. 
I didn't do the actual test with complex structure, so I leave this as a comment.

tests/test_structures.py
yakutovicha: this has been resolved in 374128a517af85c52e310807f8a69710cf5e4d67

aiidalab_widgets_base/viewers.py
```diff
@@ -21,6 +22,15 @@
 from .misc import CopyToClipboardButton, ReversePolishNotation
 from .utils import ase2spglib, list_to_string_range, string_range_to_list
 
+
+def list_to_nglview(the_list):
```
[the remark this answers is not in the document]
yakutovicha: I made it to a one-liner in 9592cb3384431babe52718fac3d4672f52a9c584. I think there is no need to keep a function for this. It does a trivial transformation from a Python list to the nglview selection format.

aiidalab_widgets_base/viewers.py
```diff
@@ -123,6 +133,133 @@ def __init__(self, parameter, downloadable=True, **kwargs):
         super().__init__([self.widget], **kwargs)
 
 
+class NglViewerRepresentation(ipw.HBox):
+    """Representation for StructureData in nglviewer"""
+
+    master_class = None
```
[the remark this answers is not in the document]
yakutovicha: Done in 3b4843bed91d5d68c869bb6df0f96f9ac6eecdd6. As for the name, it is clear to me, but I am cannot say for the others.

aiidalab_widgets_base/viewers.py
```diff
@@ -123,6 +133,133 @@ def __init__(self, parameter, downloadable=True, **kwargs):
...
+        """Update representation from the structure object."""
+        if structure:
+            if self.uuid in structure.arrays:
+                self.selection.value = list_to_string_range(
+                    np.where(self.atoms_in_representaion(structure))[0], shift=1
+                )
+                return True
+        return False
+
+    def add_myself_to_atoms_object(self, structure):
```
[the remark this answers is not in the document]
yakutovicha: done in 781684f5a0837b34f27d548ff2090b2f5f03504a

aiidalab_widgets_base/viewers.py
```diff
@@ -123,6 +133,133 @@ def __init__(self, parameter, downloadable=True, **kwargs):
...
+    def atoms_in_representaion(self, structure=None):
+        """Return an array of booleans indicating which atoms are present in the representation."""
+        if structure:
+            if self.uuid in structure.arrays:
+                return structure.arrays[self.uuid] >= self.array_threshold
+        return False
+
+    def nglview_parameters(self, indices):
+        """Return the parameters dictionary of a representation."""
+        params = {
```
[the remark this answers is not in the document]
yakutovicha: No, it is not recursive. Fixed in 58f6a37df7b11cc111425cf3b7a13dcecaa5ed12.

aiidalab_widgets_base/viewers.py
```diff
@@ -135,6 +272,8 @@ class _StructureDataBaseViewer(ipw.VBox):
 
     """
```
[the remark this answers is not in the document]
yakutovicha: done in b6c84e1dcab9811f4c546b2b4deeb4218bc9aed7

aiidalab_widgets_base/viewers.py
```diff
@@ -665,20 +977,20 @@ def _observe_input_selection(self, value):
...
 
         # Exclude everything that is beyond the total number of atoms.
-        selection_list = [x for x in value["new"] if x < self.natom]
+        selection_list = [x for x in value["new"] if x < self.natoms]
 
         # In the case of a super cell, we need to multiply the selection as well
         multiplier = sum(self.supercell) - 2
         selection_list = [
-            x + self.natom * i for x in selection_list for i in range(multiplier)
+            x + self.natoms * i for x in selection_list for i in range(multiplier)
```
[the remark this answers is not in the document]
yakutovicha: In the `ase.Atoms` object cell is always present (it is just equal to `[0, 0, 0]` for molecules.

tests/test_structures.py
yakutovicha: I am not sure how to check that properly. It depends on the starting orientation of the viewer.

aiidalab_widgets_base/viewers.py
```diff
@@ -123,6 +133,133 @@ def __init__(self, parameter, downloadable=True, **kwargs):
...
+        """Return the parameters dictionary of a representation."""
+        params = {
+            "type": self.type.value,
+            "params": {
+                "sele": list_to_nglview(indices),
+                "opacity": 1,
+                "color": self.color.value,
+            },
+        }
+        if self.type.value == "ball+stick":
```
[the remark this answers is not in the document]
yakutovicha: I am not sure this is possible. The dropdown widget accepts lists, tuples and dictionaries as options.

aiidalab_widgets_base/viewers.py
```diff
@@ -300,10 +445,148 @@ def change_camera(change):
...
+        )
+
+    def _add_representation(self, _=None, uuid=None, indices=None):
+        """Add a representation to the list of representations."""
+        self._all_representations = self._all_representations + [
+            NglViewerRepresentation(uuid=uuid, indices=indices)
+        ]
+        self._apply_representations()
+
+    def delete_representation(self, representation):
```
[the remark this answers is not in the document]
yakutovicha: done in 255564efeafc99b07fc98c754045772ec0efe553

yakutovicha:
@danielhollas, @unkcpz Thanks for the review. I addressed all the changes or replied to your comments about why I couldn't do certain things.

Please have a look. Also, if things can be postponed to the follow-up PR (which is #377), I would appreciate it. This one is already quite big and quite thoroughly tested by @cpignedoli.

For the documentation build failing - don't worry about it. I will fix it before I merge the PR.

aiidalab_widgets_base/viewers.py
```diff
@@ -766,68 +1078,112 @@ class StructureDataViewer(_StructureDataBaseViewer):
...
-        self.natom = len(self.structure) if self.structure is not None else 0
-        # Remove the current structure(s) from the viewer.
-        if change["new"] is not None:
-            self.set_trait("displayed_structure", change["new"].repeat(self.supercell))
-            self.set_trait("cell", change["new"].cell)
+        structure = change["new"]
+
+        self._viewer.clear_representations(component=0)
+
+        if structure:
```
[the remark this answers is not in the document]
yakutovicha: good idea, thanks!

done in 694cf47fdfba475b39a9c45d69df4a689c1a9312

yakutovicha:
I have to say that I dislike those random sphinx failures so much... From day to night the documentation build suddenly starts to fail, and then one needs to spend time understanding what has happened.

aiidalab_widgets_base/viewers.py
```diff
@@ -123,6 +133,133 @@ def __init__(self, parameter, downloadable=True, **kwargs):
...
+        self.master_class.delete_representation(self)
+
+    def sync_myself_to_array_from_atoms_object(self, structure):
+        """Update representation from the structure object."""
+        if structure:
+            if self.uuid in structure.arrays:
+                self.selection.value = list_to_string_range(
+                    np.where(self.atoms_in_representaion(structure))[0], shift=1
+                )
+                return True
```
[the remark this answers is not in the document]
danielhollas: Thanks, the same comment applies to `add_myself_to_atoms_object` function as well. I checked and the only place where it is called does not care about returned value..

aiidalab_widgets_base/viewers.py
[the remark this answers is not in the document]
danielhollas: It is. You can provide the options as a list of 2-tuples, where first element is the underlying value (which can be anything, e.g. Enum, and the second is a string that will be displayed to the user). See how I use it here:

https://github.com/ispg-group/aiidalab-ispg/blob/d4ab6ce6972b1be45a6669007cb343c99ecec90c/aiidalab_ispg/app/spectrum.py#L214

imo this makes the code more robust, since you're using a proper type and not just a string, (any misspeling will be caught by the linter for example)

aiidalab_widgets_base/viewers.py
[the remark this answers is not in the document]
yakutovicha: Good point, I did it in https://github.com/aiidalab/aiidalab-widgets-base/pull/373/commits/23f9a2f0912f8cbf0ca36fc6b84dd563fe3d18b8.

yakutovicha:
@cpignedoli, @danielhollas, @superstar54 is there anything else you want me to implement in this PR?

danielhollas:
@yakutovicha sorry for the delay, I got bog down with other. Still wanted to get a second pass on the review if that is okay. Will work on it now I hope, if the connection in the train is decent.

tests/test_structures.py
danielhollas: @yakutovicha I don't think your solution in that commit is correct (it looks like a breaking change possibly). `input_structure` cannot be passed via the constructor. I think you just need to revert the change in this test.
Unless I am missing something.

tests/test_structures.py
```diff
@@ -41,20 +45,39 @@ def test_structure_manager_widget(structure_data_object):
...
     # Undo the structure modification.
     structure_manager_widget.undo()
     assert np.any(
         structure_manager_widget.structure[0].position != new_structure[0].position
     )
+    structure_manager_widget.undo()  # Undo the structure creation.
+    assert structure_manager_widget.structure is None
 
-    # test the widget can be instantiated with empty inputs
+    # Test the widget with multiple importers, editors. Specify the viewer and the node class
```
danielhollas, line 61: This test is already very long, it would be good to split this up. Feel free to do in a follow up.

tests/test_viewers.py
```diff
@@ -61,44 +60,175 @@ def test_structure_data_viewer(structure_data_object):
...
-    ]
+    # Display 2*2*2 supercell
+    v.supercell = [2, 2, 2]
+    assert len(v.structure) == 2
+    assert len(v.displayed_structure) == 16
 
-    for fmt, out in format_cases:
-        v.file_format.label = fmt
-        assert v._prepare_payload() == out
+    # Test intersection of the selection with the supercell.
```
danielhollas, line 115: Same here, we should try to have shorter tests for maintainibility.

aiidalab_widgets_base/viewers.py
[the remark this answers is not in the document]
danielhollas: > For this, you need to create an Enum that is further converted to a tuple, so why not just create a tuple at first?

Because by using Enum you have a single source of truth, and you can then use it elsewhere when you have code that depends on the options (instead of comparing strings).  In other words using a type is nicer than using strings: https://www.lesinskis.com/stringly-typed-functions.html

But agree, let's not derail this here. :-D

aiidalab_widgets_base/viewers.py
```diff
@@ -766,68 +1083,111 @@ class StructureDataViewer(_StructureDataBaseViewer):
...
-            return None  # if no structure provided, the rest of the code can be skipped
-
         if isinstance(structure, ase.Atoms):
             self.pk = None
-            return structure
-        if isinstance(structure, orm.Node):
+        elif isinstance(structure, (orm.StructureData, orm.CifData)):
             self.pk = structure.pk
-            return structure.get_ase()
-        raise TypeError(
```
danielhollas, line 798: Why was this removed? Because traitlets take care of the validation already?

aiidalab_widgets_base/viewers.py
```diff
@@ -766,68 +1083,111 @@ class StructureDataViewer(_StructureDataBaseViewer):
...
+        if not structure:
             self.set_trait("displayed_structure", None)
             self.set_trait("cell", None)
+            return
+
+        # Make sure that the representation arrays from structure are present in the viewer.
+        structure_uuids = [
+            uuid
+            for uuid in structure.arrays
+            if uuid.startswith("_aiidalab_viewer_representation_")
```
danielhollas, line 1143: Should this reference the new constants defined at the top?

danielhollas (changes requested):
@yakutovicha here's what I managed to get through for now. Besides some minor or optional things, there is one important change about the failing test that I think needs to be reverted,

aiidalab_widgets_base/viewers.py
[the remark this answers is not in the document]
danielhollas: @yakutovicha thanks for the comment. As for the name, `viewer_class` seems more appropriate to me?

Side note: Github really makes it hard for such large changes and multiple reviews. :-(

aiidalab_widgets_base/viewers.py
yakutovicha: Yes, exactly for that. The code never reaches this point.

aiidalab_widgets_base/viewers.py
yakutovicha: ah, right, not sure how I missed it. Thanks 👍 
done in https://github.com/aiidalab/aiidalab-widgets-base/pull/373/commits/66ccfbdf2e689b91e927ec2c8b25af3d602f9262

aiidalab_widgets_base/structures.py
```diff
@@ -39,9 +38,7 @@ class StructureManagerWidget(ipw.VBox):
     input_structure = tl.Union(
         [tl.Instance(ase.Atoms), tl.Instance(orm.Data)], allow_none=True
     )
-    structure = tl.Union(
-        [tl.Instance(ase.Atoms), tl.Instance(orm.Data)], allow_none=True
-    )
+    structure = tl.Instance(ase.Atoms, allow_none=True)
```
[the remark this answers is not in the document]
yakutovicha: The `structure` train has always been an `ase.Atoms` object. All editors were assuming that they were acting on the `ase.Atoms`, not on the aiida object.

Also, if you look into the `_observe_input_structure` you will notice that the input structure object is immediately converted to ASE.

So I would consider this as a bug fix.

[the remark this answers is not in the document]
danielhollas: Yep, sounds good.

tests/test_structures.py
yakutovicha: done in https://github.com/aiidalab/aiidalab-widgets-base/pull/373/commits/cb6e6bb83df0f4ebe7086e8e9b69a1e4254691fc

tests/test_viewers.py
yakutovicha: done in https://github.com/aiidalab/aiidalab-widgets-base/pull/373/commits/cb6e6bb83df0f4ebe7086e8e9b69a1e4254691fc.

The `test_structure_data_viewer_selection` remains a bit long, but I think it is fine since it focuses on different ways of atom selection.

aiidalab_widgets_base/structures.py
```diff
@@ -1271,7 +1275,7 @@ def translate_dxdydz(self, _=None, atoms=None, selection=None):
 
         # The action.
         atoms.positions[self.selection] += np.array(self.str2vec(self.dxyz.value))
-
+        self.input_selection = None  # Clear selection.
```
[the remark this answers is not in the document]
yakutovicha: I made an issue to address this: https://github.com/aiidalab/aiidalab-widgets-base/issues/531

aiidalab_widgets_base/structures.py
```diff
@@ -1439,6 +1450,8 @@ def remove(self, _=None, atoms=None, selection=None):
         """Remove selected atoms."""
         del [atoms[selection]]
 
+        # The order of the traitlets below is important -
+        # we must be sure trait atoms is set before trait selection
         self.structure = atoms
         self.input_selection = None
```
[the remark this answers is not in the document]
yakutovicha: You are right, this was not needed. I think we just blindly followed the schema that was used for all the other structure modifiers:
```python
self.input_selectin = None
self.input_selection = <new_selection>
```

But agree, it doesn't make sense here.

What concerns the comment, I think we will remove it as soon as #531 is fixed.

aiidalab_widgets_base/viewers.py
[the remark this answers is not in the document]
yakutovicha: > @yakutovicha thanks for the comment. As for the name, `viewer_class` seems more appropriate to me?

That sounds good to me! Done in https://github.com/aiidalab/aiidalab-widgets-base/pull/373/commits/5abbf5d601879bd50f033f8900caf7d221d9a080



> Side note: Github really makes it hard for such large changes and multiple reviews. :-(

I know what you mean. This usually happens when you make individual comments as part of your review. Then they appear in both places: in your review and as a reply to the thread.

Usually, I reply to the threads first and then do the review.

tests/test_structures.py
yakutovicha: Let me explain why I did it this way.

1. The `structure` trait is the one that should be the only trait that is linked between the manager and the viewer. It represents the current molecule (possibly with modifications) that user is supposed to see.
2. The `input_structure` trait is the one receiving the structure from importers, it doesn't make sense to pass it to the viewer.
3. Usually, I pass `**kwargs` to the parent class, because they might contain some styling properties. In that case, styling of the manager has very little to do with the styling of the viewer - they are just different.

Based on this I decided not to pass `**kwargs` at all. If one wants to update styling, one can directly provide this info here:

```python
self.viewer = StructureDataViewer(style=..., layout=...)
```

yakutovicha:
@danielhollas, I believe I addressed all your suggestions in one way or another. Please have a look.

And thank you very much for the careful review. I appreciate your time and effort.

yakutovicha:
During tomorrow's documentation day, I will draw a data-flow schema and will add it to the documentation. This will clarify a lot of things.

tests/test_structures.py
danielhollas: Right, that all makes sense, thanks! However, it still seems to me that this change in the test needs to be reverted, no? Looking at the StructureManagerWidget code, passing `input_structure` via constructur will have no effect, since it would be an keyword argument. Am I still missing something?

yakutovicha: Oh, I think I am getting now what is the problem.

When dealing with traits it is ok to pass a trait through the constructor. They get initialised here:
https://github.com/aiidalab/aiidalab-widgets-base/blob/5abbf5d601879bd50f033f8900caf7d221d9a080/aiidalab_widgets_base/structures.py#L134

Here is the smallest example you can play with:
```python
from traitlets import HasTraits, Int

class Person(HasTraits):
    age = Int(allow_none=True)

# Creating an instance of the Person class
person = Person(age=30)

# Accessing and modifying attributes
print(person.age)  # Output: 30

person.age = 35

print(person.age)  # Output: 35

```

aiidalab_widgets_base/structures.py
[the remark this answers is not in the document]
danielhollas: > Maybe it is a good idea to make the structure_node trait immutable because it is not meant to be modified externally.

`structure_node` trait is already read-only so I think we're good here. Thanks for the explanation. I'll leave this unresolved just to keep this discussion visible.

yakutovicha:
@danielhollas, do you want to have another look at this PR? It would be great to close this story to move on with other things.

danielhollas (approved):
LGTM, thanks for implementing the changes @yakutovicha. I wanted to do a quick test in my app, but realized that I hide the Appearence tab to simplify the UI. :smile: 

Excited to get this big feature merged, congrats! :tada: :rocket: 

> During tomorrow's documentation day, I will draw a data-flow schema and will add it to the documentation. This will clarify a lot of things.

This is an awesome idea! Did you end up doing that? If so, could you please post it here as well? In general I think we should do that for all widgets that are a bit more complicated (all the widgets around the structure / viewer for example).

> And thank you very much for the careful review. I appreciate your time and effort.

:heart: 

> ...I have to admit there is nothing to nitpick, it is quite clear what the PR wants to achieve with nice unit tests. I nitpick on one unit test to add an assertion.

I guess I was the "reviewer number 2" for this PR :blush: :joy:

yakutovicha:
> This is an awesome idea! Did you end up doing that? If so, could you please post it here as well? In general I think we should do that for all widgets that are a bit more complicated (all the widgets around the structure / viewer for example).

I didn't manage to finish that task, but I added an issue to keep in mind #534.

unkcpz:
Congrats! @yakutovicha :clinking_glasses:

tests/test_structures.py
danielhollas: CC @edan-bainglass

unkcpz: > since sometimes the traitlets are in fact explicitly defined in the constructor.

This also I noticed and I think we should avoid doing it.

PR #375, opened 2022-10-21
UUID as traitlets for threading related widgets

unkcpz:
The widget changed are:

- [x] `ComputationalResourcesWidget` because of QeApp code setup trigger `refresh` it in thread.
- [x] `ComputerDropdownWidget`
- [x] `ProcessListWidget` because thread call in update function for process.
- [x] `ProcessMonitor` thread call in `join`
- [x] `ProcessNodesTreeWidget`


danielhollas:
@unkcpz sorry for a delay on this. I'll try to take a closer look and test this next week. Thanks!

unkcpz:
> @unkcpz sorry for a delay on this. I'll try to take a closer look and test this next week. Thanks!

Thanks! Since this PR is separated from the migration to 2.x and it is not manifest the thread issue in 1.6.x, I think you can have a test to make sure everything is working with the new changes in this PR.

danielhollas:
I made some progress today, was able to convert my app to support this branch and did some basic testing. Will try to do more testing tomorrow.

unkcpz:
Hi @danielhollas, do you have any update on using and testing this? I want to try to finalize this and hopefully have QeApp support AiiDA 2.x supported by next week.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -1228,21 +1215,16 @@ def refresh(self, _=None):
...
             return None
 
-        if isinstance(computer, orm.Computer):
-            if computer.label in self.computers:
-                return computer
-            self.output.value = f"""Computer instance '<span style="color:red">{computer.label}</span>'
-            is not configured in your AiiDA profile."""
-        return None
+        try:
+            _ = UUID(computer_uuid, version=4)
```
danielhollas, line 1225: Does this work also if `computer_uuid is None`? There was an explicit check for this previously.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -1149,21 +1141,16 @@ class ComputerDropdownWidget(ipw.VBox):
     """Widget to select a configured computer.
 
     Attributes:
-        selected_computer(Unicode or Computer): Trait that points to the selected Computer instance.
-            It can be set either to an AiiDA Computer instance or to a computer label (will
-            automatically be replaced by the corresponding Computer instance). It is linked to the
+        selected_computer(Unicode of computer UUID): Trait that points to the selected Computer instance.
```
danielhollas, line 1144: Should we keep support for using a computer label instead of UUID?

aiidalab_widgets_base/computational_resources.py
```diff
@@ -197,29 +197,21 @@ def refresh(self, _=None):
...
-        elif isinstance(code, orm.Code):  # Check code by value.
-            label = self._full_code_label(code)
-            if label in self.codes:
-                return code
-            self.output.value = f"""The code instance '<span style="color:red">{code}</span>'
-            supplied was not found in the AiiDA database."""
-
-        # This place will never be reached, because the trait's type is checked.
-        return None
+        try:
```
danielhollas, line 208: I think for code we still want to support selecting code by their labels, no?
In which case this validation should not be here, because code labels will not be valid UUIDs? Or are they?

aiidalab_widgets_base/process.py
```diff
@@ -545,10 +542,10 @@ class ProcessListWidget(ipw.VBox):
 
     past_days (int): Sumulations that were submitted in the last `past_days`.
 
-    incoming_node (int, str, Node): Trait that takes node id or uuid and returns the node that must
+    incoming_node (tr): Trait that takes node uuid that must
```
danielhollas, line 545: Any reason why not keep support of passing `id` instead of `uuid`?

aiidalab_widgets_base/process.py
```diff
@@ -701,10 +708,10 @@ def __init__(self, callbacks=None, on_sealed=None, timeout=None, **kwargs):
 
         super().__init__(**kwargs)
 
-    @traitlets.observe("process")
+    @traitlets.observe("process_uuid")
     def _observe_process(self, change):
-        process = change["new"]
-        if process is None or process.id != getattr(change["old"], "id", None):
+        process_uuid = change["new"]
+        if process_uuid is None or process_uuid != change["old"]:
```
danielhollas, line 714: Can we clean this up a bit to reduce the indentation levels? The first if statement is super confusing to me.

If I understood the logic correctly, this should work?
```suggestion
        if process_uuid == change["old"]:
            return
        # stop thread (if running)
        if self._monitor_thread is not None:
            with self._monitor_thread_lock:
                self._monitor_thread_stop.set()
                self._monitor_thread.join()
        
        if process_uuid is None:
            return

        with self.hold_trait_notifications():
            with self._monitor_thread_lock:
                self._monitor_thread_stop.clear()
                self._monitor_thread = Thread(
                    target=self._monitor_process, args=(process_uuid,)
                )
                self._monitor_thread.start()
        
```

danielhollas (changes requested):
@unkcpz here's my initial review. I agree with the target of merging this next week.

I did some testing within my own application, but need to do a bit more and want to test individual widgets as well.

I have also ran into an interesting edge case that possibly not covered here. I had a widget that passed an AiiDA node to another widget based on the UI interaction (e.g. user clicks a button and AiiDA node is passed to a widget). Presumably because the UI interaction was handled by a different thread than the thread that originally passed the Node,  it ran into the same issue as we're solving here, even though the widget itself did not explicitly use threading. So something to be mindful of.

aiidalab_widgets_base/computational_resources.py
unkcpz: > I think for code we still want to support selecting code by their labels, no?

No, I don't plan to support the label as value of the widget. What do you think? I'll update the docstring if you agree with it. 

Also pinning @yakutovicha for comment, since you author this widget.

aiidalab_widgets_base/computational_resources.py
unkcpz: As the comment above, my initial plan is to only use UUID, since I think it is only used internally for the widget. The label will be shown from the key of `computer(Dict)` trait.
IMO, the value should be as static/identical as possible.

aiidalab_widgets_base/computational_resources.py
unkcpz: `computer_uuid is None` is included by `if not computer_uuid:` (two lines ahead), correct?

aiidalab_widgets_base/process.py
unkcpz, line 545:
```suggestion
    incoming_node (str): Trait that takes node uuid that must
```

unkcpz: You mean `pk`? Do you have any use case for where you what to use primary key, since I think you can use UUID to include all the cases for `pk`?

aiidalab_widgets_base/process.py
unkcpz: Yes, fully agree. I try to simplify it more. Please have a test of it. I'll do the test further more on qeApp.

unkcpz (commented):
@danielhollas thanks a lot for the test. I think the main open issue is whether we need to using only the UUID or also support the other entities (label/pk etc.) as the value of those widgets. 

I update the ProcessMornitor to try to make it more clear.

aiidalab_widgets_base/process.py
```diff
@@ -701,29 +708,31 @@ def __init__(self, callbacks=None, on_sealed=None, timeout=None, **kwargs):
...
-        assert process_id is not None
-        process = load_node(process_id)
+        """When the process_uuid is changed, stop the previous
+        monitor if exist. Start a new one in thread."""
+        process_uuid = change["new"]
+
+        if process_uuid is None:
+            return
+
+        # stop thread (if running)
```
danielhollas, line 720: @unkcpz I think the thread needs to be stopped when the process_uuid is set to None

aiidalab_widgets_base/computational_resources.py
yakutovicha: Originally, when developing this, I aimed to mimic the behaviour of `verdi code show` (or similar) that would accept any way of selecting the code (by `name`, `id`, `uuid`). It is nice to have, but if using only `uuid` lets us significantly simplify the code, maybe let's stick to `uuid` only.

I do not have strong preferences w.r.t both options.

unkcpz:
> Next week I will for sure need to do some production calculations so will test this more, perhaps this can be merged but still not released as a full 2.0 version before more testing is done?

Yes, I think we can set some milestones in the next meeting for `2.0.0`. I think the major version is for backward compatibility breakup. This means if we change other traitslet, it breaks compatibility. I think we need to make the release after that?

aiidalab_widgets_base/computational_resources.py
yakutovicha: I guess we are all only back next week 😄 . I am off till Tuesday. @unkcpz mentioned that he is packed till the end of this week also.

unkcpz:
Hi @yakutovicha, are you reviewing this? I think we can have another alpha release with this merged. Then I can keep on working on QeApp. 
@danielhollas, also if you want to have a final check/test.

aiidalab_widgets_base/computational_resources.py
yakutovicha, line 1144:
```suggestion
        value(computer UUID): Trait that points to the selected Computer instance.
```

aiidalab_widgets_base/computational_resources.py
yakutovicha: > think for codes and computers it makes a lot of sense to be able to use the label because that's what is independent of the concrete database.

I believe that is not exactly true. Suppose a computer was exported to a database which already contains a computer with the same label. The only way they could be distinguished is by UUID. So I am not sure what is the advantage of keeping the label here.

aiidalab_widgets_base/computational_resources.py
yakutovicha: I prefer to keep UUID only, tbh. See my reasoning above.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -27,12 +28,13 @@ class ComputationalResourcesWidget(ipw.VBox):
     """Code selection widget.
     Attributes:
 
-    value(Unicode or Code): Trait that points to the selected Code instance.
-    It can be set either to an AiiDA Code instance or to a code label (will automatically
-    be replaced by the corresponding Code instance). It is linked to the 'value' trait of
+    value(Unicode or UUID of the code node): Trait that points to
```
yakutovicha, line 31:
```suggestion
    value(code UUID): Trait that points to
```

aiidalab_widgets_base/process.py
```diff
@@ -688,7 +695,7 @@ def start_autoupdate(self, update_interval=10):
 class ProcessMonitor(traitlets.HasTraits):
     """Monitor a process and execute callback functions at specified intervals."""
 
-    process = traitlets.Instance(ProcessNode, allow_none=True)
+    process_uuid = Unicode(allow_none=True)
```
yakutovicha, line 698: What about naming it `value`?

yakutovicha (commented):
Overall looks good to me.

Looking at the discussion, I see one unclear point, whether to allow to use of node labels together with UUIDs. IMO, it would be ok to keep UUIDs only, as I wrote in the review comments.

@danielhollas please let me know if you agree with that.

danielhollas:
Thanks @unkcpz. I was testing it today / yesterday in my app and did not find any issues, but was not testing all the functionality here. Unfortunately, I had  some other urgent stuff to do for tomorrow's meeting. I'll try to do one more round of testing today late evening if I get to it. Otherwise, this is good to merge from my side and release a new alpha version.

@yakutovicha I think you're right about the UUID value, I am find with only supporting UUID. One question, how does the Code selection UI looks if there are two computers with the same label? e.g. if one imports nodes from another DB? Perhaps we should display PKs in the Dropdown in such cases, since otherwise the user will only see two identical labels?

yakutovicha:
> One question, how does the Code selection UI looks if there are two computers with the same label? e.g. if one imports nodes from another DB? Perhaps we should display PKs in the Dropdown in such cases, since otherwise the user will only see two identical labels?

I think they are indistinguishable from the current UI. I would agree with you that adding a PK number together with the title would be a good option 👍

PR #377, opened 2022-10-26
Feature/bonds

### LINKED ISSUE #48: nglview: figure out why bonds are not always shown by deffault
Probably, this is not something we can do ourselves, and the issue should go to the nglview repository. 

For some structures, nglview shows only atoms, while for the others, it shows atoms and bonds.

- [x] Figure out whether the problem is on our side or on nglview side. (The problem is on the nglview side. After certain system size bonds are not shown anymore)

- [x] Depending on the previous point either fix it or create an issue for nglview

cpignedoli:
fixes #48 

In nglview if a system is too big bonds will not be computed.
I bypass the "ball+stick" representation defining it as "spacefill" plus shapes that are computed via ase connectivity matrix.

I would not draw also the atom spheres through  shapes otherwise it would be a problem to pick atoms with mouse selection

Despite I tested it also with ~1000 atoms it is meant to work on smaller system sizes.

The image below is outdated, just the "ball+stick" representation is needed to produce the image

image (15)

To Do:

- [ ] Consider using Enum for the dropdown widgets.


aiidalab_widgets_base/viewers.py
```diff
@@ -335,9 +392,344 @@ def change_camera(change):
...
+        print("compute bonds")
+        if len(structure) > 1:
+            matrix = nl.get_connectivity_matrix()
+            for k in matrix.keys():
+                i = structure[k[0]]
+                j = structure[k[1]]
+
+                v1 = np.array([i.x, i.y, i.z])
+                v2 = np.array([j.x, j.y, j.z])
+                mic_vector = structure.get_distance(k[0], k[1], mic=True, vector=True)
```
superstar54, line 529: Here has two performance issues:
- you calculate the distance between atoms twice. 1) `nl.get_connectivity_matrix`; 2) `structure.get_distance`.
- you use a `for` loop, which is slow.

It takes around 1.3 s for 2000 atoms.

Please optimize this code. Feel free to have a discussion.

aiidalab_widgets_base/viewers.py
```diff
@@ -16,6 +16,7 @@
 from aiida.orm import Node
 from ase import Atoms, neighborlist
```
superstar54, line 17: Please move this import into the function `compute_bonds`.

superstar54 (changes requested):
@cpignedoli Thanks for adding this. very helpful.

aiidalab_widgets_base/viewers.py
cpignedoli: not really sure how to improve this:
connectivity_matrix is the fastest way I know via ASE to get bonds in a structure after that
the loop runs only on bonded atoms (so is not N*N) and the get_distance is needed to have the MIC vector connecting two atoms needed to draw proper bonds also in the case atoms are at the edge of the cell. Do you know how to speed up this?

aiidalab_widgets_base/viewers.py
cpignedoli: good catch, I will :)

aiidalab_widgets_base/viewers.py
cpignedoli: wonderful @superstar54 I will try this way, thanks a lot!

cpignedoli: done :)

aiidalab_widgets_base/viewers.py
cpignedoli: done :)

PR #379, opened 2022-11-01
Fix: Do not assume Selection tab in StructureDataViewer

danielhollas:
Passing an empty `configuration_tabs` to StructureDataViewer would throw a ValueError, because the code would assume that the Selection tab is always present.

### DIFF
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -212,8 +212,12 @@ class _StructureDataBaseViewer(ipw.VBox):
         if configuration_tabs is None:
             configuration_tabs = ["Selection", "Appearance", "Cell", "Download"]
 
+        self.selection_tab_idx = None
         if len(configuration_tabs) != 0:
-            self.selection_tab_idx = configuration_tabs.index("Selection")
+            try:
+                self.selection_tab_idx = configuration_tabs.index("Selection")
+            except ValueError:
+                pass
             self.configuration_box = ipw.Tab(
                 layout=ipw.Layout(flex="1 1 auto", width="auto")
             )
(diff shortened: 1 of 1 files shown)


unkcpz:
I see your point now. If passing a list without `Selection` would cause the issue.

aiidalab_widgets_base/viewers.py
```diff
@@ -212,8 +212,12 @@ def __init__(
         if configuration_tabs is None:
             configuration_tabs = ["Selection", "Appearance", "Cell", "Download"]
 
+        self.selection_tab_idx = None
         if len(configuration_tabs) != 0:
-            self.selection_tab_idx = configuration_tabs.index("Selection")
+            try:
+                self.selection_tab_idx = configuration_tabs.index("Selection")
+            except ValueError:
+                pass
```
unkcpz, line 220:
```suggestion
                self.selection_tab_idx = None
```
I'd suggest keeping it more compact.

unkcpz (changes requested):

aiidalab_widgets_base/viewers.py
danielhollas: It is set to None by default before the if statement. It must be set there, otherwise if an empty configuration list is passed it would be undefined. But maybe I misunderstood and you want to set it again here?

unkcpz: You are not misunderstood, I was suggesting removing the code above. 
I see your point, you are correct 😁

unkcpz (approved):
LGTM! 🚀

danielhollas:
Thanks @unkcpz! As this is currently blocking a feature I am working on, would you mind publishing a new 1.4.x version? There have been a couple of other fixes since last version as well.

unkcpz:
No problem, I’ll do it tomorrow when I back to office.

unkcpz:
Hi @danielhollas, I invite you to our aiidalab app slack channel, feel free to join.

unkcpz:
@danielhollas `1.4.2` is released.

PR #383, opened 2022-11-12
Fix/traitlet structure fixes #381 blocked by #377

### LINKED ISSUE #381: Structure importers: cannot import the same structure twice.
Suppose I have two importers: "from file" and "SMILES":

If I do the following:
1. Create a molecule from SMILES `C=C`, I will obtain the correct `C2H4` molecule
2. After that, I import the `file.xyz`, and the molecule will be correctly replaced by the uploaded structure.
3. Then, switching back to SMILES and pressing "Generate molecule" will do nothing because the SMILES string is unchanged.
4. Modifying the SMILES string to another value (`C` for instance) correctly provides a new molecule.
5. Additionally, if I go back to "Import from computer" and re-import the same `file. xyz`, it will do nothing again. If I import a different file, it will work.

It seems to be related to the traits notification problem. Since the trait somehow keeps the same value it does not notifies the structure update.

## Additional observation.

For an unknown reason the information about the change of the `structure` trait of an importer doesn't go to the `input_structure` of the manager, even though it should:
https://github.com/aiidalab/aiidalab-widgets-base/blob/035047f38f4e92a3b13ecc99879cb4a61fd6cde4/aiidalab_widgets_base/structures.py#L170

cpignedoli:
Fixes #381

This fixes the setting of `self.structure = None` when an action button is pressed for an importer ("SMILES" and "import from computer")

Thanks, @danielhollas, I tried `self.structure = None` just in observe("input_structure") but this does not work still, I keep the modification also in that observe.


danielhollas:
@cpignedoli why is this blocked by #377? Seems like this should be a separate fix?

cpignedoli:
it is true but I implemented it on top of another blocked PR

yakutovicha:
@cpignedoli let's try to move this PR somewhere. I see @danielhollas cherry-picked https://github.com/aiidalab/aiidalab-widgets-base/pull/383/commits/bc179c4dea0ea97156200d72d0fab6e013a4f4da into a separate PR (#549). Is it the only relevant change that remains here?

cpignedoli:
yes this is so old that for sure it has been fixed somewhere else somehow

cpignedoli:
we checked and the bug is no more there

cpignedoli:
I close

Issue #392, opened 2022-11-21
Prune dependencies


danielhollas:
It would be good to do an inventory of AWB dependencies, especially those that are not currently included in the `aiidalab/full-stack` image. This will help us when we integrate AWB to the image to mitigate the size increase.

- [ ] #591
- [x] #590
- [x] Make ipython an explicit dependency, since we're using its APIs directly in AWB. We should only set a minimum version, we can take a version from the aiidalab/full-stack:aiida-2.1 image.
- [x] Make optimade optional
- [x] Make aiidalab-eln optional (#609)

danielhollas:
The only remaining point here is removing pandas, which is tracked in #591 so going to close this.

Issue #384, opened 2022-11-14
nbconvert version is too strict and other dependencies


unkcpz:
https://github.com/aiidalab/aiidalab-widgets-base/blob/035047f38f4e92a3b13ecc99879cb4a61fd6cde4/setup.cfg#L33

Hi @yakutovicha, do you remember why this has to be pinned with `nbconvert<6`? I think we can go through the dependencies and polish them.

yakutovicha:
Honestly, I don't. I agree - we should revise the dependencies. It should probably go along with enabling tests.

unkcpz:
This is closed by #395

PR #395, opened 2022-11-26
clean up dependencies requirements

unkcpz:
The principle of my cleaning process is the packages in the list do not conflict with each other and do not duplicate. For example, if the package is installed by aiida-core dep, it should not be specified again in the list.
As for the jupyter backend packages, there is no need to install then, since 

`aiida-core` > 2 will lead to install `numpy~=1.19` therefore `numpy~=1.23` is installed.
- numpy~=1.17

The following packages are even in conflict and will override the deps which causes the Jupyter engine to fail, we need to pining version when we find future confliction.
- `jupyter-client<7`
- `nbconvert<6`

- `aiidalab>=21.11.2` is already pinned in the stack

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -19,25 +19,18 @@ classifiers =
 [options]
 packages = find:
 install_requires =
-    aiida-core>=2.0,<3
-    aiidalab>=21.11.2
+    aiida-core[atomic-tools]>=2.0,<3
     aiidalab-eln>=0.1.2,~=0.1
     ansi2html~=1.6
-    ase<3.20
     bokeh~=2.0
     humanfriendly~=10.0
     ipytree~=0.2
-    ipywidgets~=7.6
-    jupyter-client<7
     more-itertools~=8.0
-    nbconvert<6
     nglview~=3.0
-    numpy~=1.17
     optimade-client==2022.9.19
-    pandas~=1.0
     scikit-learn~=0.24
-    spglib~=1.16
     vapory~=0.1.2
+    ipywidgets~=7.7
 python_requires = >=3.8
 include_package_data = True
 zip_safe = False
--- a/setup.cfg
+++ b/setup.cfg
@@ -19,21 +19,21 @@ classifiers =
 [options]
 packages = find:
 install_requires =
+    PyCifRW~=4.4
     aiida-core>=2.0,<3
     aiidalab-eln>=0.1.2,~=0.1
     ansi2html~=1.6
+    ase~=3.18
     bokeh~=2.0
     humanfriendly~=10.0
     ipytree~=0.2
+    ipywidgets~=7.7
     more-itertools~=8.0
     nglview~=3.0
     optimade-client==2022.9.19
     scikit-learn~=0.24
-    vapory~=0.1.2
-    ipywidgets~=7.7
-    PyCifRW~=4.4
-    ase~=3.18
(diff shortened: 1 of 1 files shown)


unkcpz:
The test failed as expected, it rely on https://github.com/aiidalab/aiidalab-docker-stack/pull/333

danielhollas:
I wonder was there a specific reason for this constraint? `nbconvert <6` Do you know @yakutovicha? Are we risking breakage when `nbconvert` is updated?

yakutovicha:
> nbconvert

I remember nothing about this package. One needs to test...

yakutovicha:
> > I wonder was there a specific reason for this constraint? `nbconvert <6` Do you know @yakutovicha? Are we risking breakage when `nbconvert` is updated?
> 
> I remember nothing about this package. One needs to test...

btw, there is also https://github.com/aiidalab/aiidalab-widgets-base/issues/384.

danielhollas:
Thanks @yakutovicha, I was not aware of that issue. Couple days ago I also opened #392 which is related.

danielhollas (changes requested):
I think we should take this step by step.

> The following packages are even in conflict and will override the deps which causes the Jupyter engine to fail, we need to pining version when we find future confliction.
    jupyter-client<7
    nbconvert<6
    aiidalab>=21.11.2 is already pinned in the stack

@unkcpz could you open a new PR which would only fix the dependencies for the current Docker stack, listed above? `aiida-core[atomic_tools]` is a separate discussion.

unkcpz:
> @unkcpz could you open a new PR which would only fix the dependencies for the current Docker stack, listed above? aiida-core[atomic_tools] is a separate discussion.

Sure, updated. I add `PyCifRW` explicitly since it is required by this package.

danielhollas (approved):
Thanks @unkcpz. I did a quick test locally and it looks good to me. Approving but we should wait until we release a new docker-stack version so that the tests are passing here.

unkcpz:
The firefox tests are expected to fail and PR #397 is for it, there are two chrome tests that also fail but I think if there are two passes then the implementation is correct.. I'll merge this and if anything happened I'll fix it afterward.

yakutovicha:
Sorry, I am late to the party. 

> * `aiidalab>=21.11.2` is already pinned in the stack

I am not 100% sure this is a good thing to do. AiiDAlab could be setup outside of the stack, and AWB does require aiidalab package to function.

unkcpz:
You are right, and I find the doc build always lack the necessary packages if `aiidalab` package is not explicitly installed. I add `aiidalab>=21.11.2` back in https://github.com/aiidalab/aiidalab-widgets-base/pull/397 to fix the doc build test.

setup.cfg
```diff
@@ -19,24 +19,20 @@ classifiers =
...
     ipytree~=0.2
-    ipywidgets~=7.6
-    jupyter-client<7
+    ipywidgets~=7.7
     more-itertools~=8.0
-    nbconvert<6
     nglview~=3.0
-    numpy~=1.17
     optimade-client==2022.9.19
-    pandas~=1.0
```
danielhollas, line 37: @unkcpz I just noticed that you removed pandas here. Do you know where else it is coming from?

danielhollas: Ah, it seems it comes from `optimade-client`, never mind.

PR #397, opened 2022-11-28
Install firefox/geckodriver in GH workflow for test

unkcpz:


unkcpz:
> I am confused, the tests seem to be duplicated? e.g. running on both pull_reqest and push.

That's because I make the PR from this repo directly. Other repositories are the same has actions on pull request and push.

danielhollas:
Fair enough. Can you explain a bit the changes here? What are they supposed to solve?
Some tests seems to be still failing.

unkcpz:
The test failed since somehow the github server doesn't install firefox by default but only the chrome engine. I think the change happened last week both QeApp and AWB encounter the same issue. However I cannot find the change log from web. 
Install the geckodriver and firefox, and then specify the path to the firefox engine in selenium test solve the issue

danielhollas:
> The test through selenium is still not very robust, it depends on whether the page is successfully loaded, where I set sleep time to wait to load but not always reach the end. Probably there are a pattern we can use to sure the page is loaded. Such like.

That's an interesting pattern, thanks! Are you planning to implement it as part of this PR.

I think it would be good to also pin the setuptools to get rid of the readthedocs failure. I guess it will take us some time to migrate to flit+pyproject.toml so we should fix it in the meantime. https://github.com/aiidateam/aiida-core/pull/5782/files

unkcpz:
> That's an interesting pattern, thanks! Are you planning to implement it as part of this PR? Would be good to have the tests passing again.

I simply increase the load time to workaround 😁, open issue https://github.com/aiidalab/aiidalab-widgets-base/issues/398 to be addressed soon.

> I think it would be good to also pin the setuptools to get rid of the readthedocs failure. I guess it will take us some time to migrate to flit+pyproject.toml so we should fix it in the meantime

Yes, I agree. The `pyproject.toml` is added and pinning the `setuptools` version to `65.5.0`.

.github/workflows/ci.yml
```diff
@@ -57,5 +58,20 @@ jobs:
               run: |
                   pip install -U -e .[tests]
 
+            - name: Set jupyter token env
+              run: echo "JUPYTER_TOKEN=$(openssl rand -hex 32)" >> $GITHUB_ENV
+
+            - name: Install Firefox geckodriver
```
danielhollas, line 64: I don't understand, here you use this action, but also install it manually below? If it is neccessary to do like this, can you please add a comment with an explanation?

.github/workflows/ci.yml
```diff
@@ -57,5 +58,20 @@ jobs:
...
+              uses: browser-actions/setup-firefox@latest
+              with:
+                  firefox-version: ${{ matrix.firefox }}
+
+            - name: Install geckodriver
```
danielhollas, line 69: This installation (and the above) should only happen when matrix.browser==Firefox.

danielhollas (changes requested):

unkcpz:
Fair, thanks a lot! I add the check for the matrix and add a comment.

.github/workflows/ci.yml
```diff
@@ -57,5 +58,25 @@ jobs:
...
+              run: |
+                  wget -c https://github.com/mozilla/geckodriver/releases/download/v0.30.0/geckodriver-v0.30.0-linux64.tar.gz
+                  tar xf geckodriver-v0.30.0-linux64.tar.gz -C /usr/local/bin
+              if: matrix.browser == 'Firefox'
+
             - name: Run pytest
-              run: TAG=${{ matrix.tag }} JUPYTER_TOKEN=$(openssl rand -hex 32) pytest --driver ${{ matrix.browser }}
+              run: |
+                  pytest --driver ${{ matrix.browser }}
```
danielhollas, line 80:
```suggestion
              run: pytest --driver ${{ matrix.browser }}
```

danielhollas (approved):
All tests passing! :tada: Thanks so much @unkcpz

PR #399, opened 2022-11-29
Update `flake8` config for `flake8>=6`

yakutovicha:
This should fix the failed PRs, similar to https://github.com/aiidalab/aiidalab-widgets-base/pull/396

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,10 +58,10 @@ tests =
 
 [flake8]
 ignore =
-    E501  # Line length handled by black.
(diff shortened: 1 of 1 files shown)


unkcpz (approved):
Thanks! I think aiidalab-qe has same issue. LGTM!
Can you make another pre-release after this?

unkcpz:
> Or we are not sure this will be supported by aiidalab?

I am sure it is *NOT* supported 😆, it will take longer to support I think.

PR #400, opened 2022-11-29
Chore: maintain pre-commit checks

yakutovicha:


unkcpz (approved):
Thanks for taking of this! I think if the pre-commit test pass then it should all be good.

PR #401, opened 2022-11-30
Work on pre-commit hooks, small readme update.

yakutovicha:
* Remove unused .prospector.yaml
* Readme: add --tag-num option to update tag number
* Flake8: add more plugins to better analyse the code:
  * `flake8-bugbear` for detecting bugs and potential design problems
  * `flake8-builtins` for verifying that builtins are not used as variable names
  * `flake8-comprehensions` for writing better and consistent comprehensions
  * `flake8-debugger` for checking that there are no forgotten breakpoints
  * `flake8-logging-format` for consistent logging,
  * pep8-naming `for verifying that PEP8 naming conventions are followed`
  * `pyflakes` for checking source files for various errors
  * [tryceratops](https://github.com/guilatrova/tryceratops) to avoid exception anti-patterns


yakutovicha:
Hey reviewers 👋 

As you can see, I enabled a few flake8 plugins to check the code for various issues, antipatterns, etc. However, there are two types of errors I am not happy with:
* `TC101` - "Too many try blocks in your function". In principle, I can split them into subfunctions to have one try-except block per function. But I don't know why this is needed. What advantage does it bring?
* `TC003` - "Avoid specifying long messages outside the exception class". I don't get what it wants from me 😞 . Shortening the string doesn't help with the issue.

My first intention is to disable those checks unless you have a different opinion.

aiidalab_widgets_base/bug_report.py
```diff
@@ -111,11 +111,11 @@ def _convert_ansi_codes_to_html(msg):
 def _format_truncated_traceback(traceback, max_num_chars=2000):
     """Truncate the traceback to the given character length."""
     n = 0
-    for i, line in enumerate(reversed(traceback)):
+    for _i, line in enumerate(reversed(traceback)):
```
danielhollas, line 114: I don't understand this change, it looks very confusing to me. Which rule is violated here?

aiidalab_widgets_base/utils/__init__.py
```diff
@@ -36,14 +36,14 @@ def predefine_settings(obj, **kwargs):
             raise AttributeError(f"'{obj}' object has no attribute '{key}'")
 
 
-def get_ase_from_file(fname, format=None):  # pylint: disable=redefined-builtin
+def get_ase_from_file(fname, file_format=None):  # pylint: disable=redefined-builtin
```
danielhollas, line 39: This is a breaking change, are we fine with that? (I am using this in my app). Did you modify all the instance which call this function? I don't see it, but there should only be a few.

It is not a public method so perhaps it's fine.

aiidalab_widgets_base/process.py
```diff
@@ -94,7 +94,7 @@ def __init__(  # pylint: disable=too-many-arguments
         if callable(inputs_generator):
             self.inputs_generator = inputs_generator
         else:
-            raise ValueError(
+            raise TypeError(
                 "The `inputs_generator` argument must be a function that "
                 f"returns input dictionary, got {inputs_generator}"
```
danielhollas, line 99: To make this consistent with other such exceptions that are in this PR.

```suggestion
                f"returns input dictionary, got {type(inputs_generator)}"
```

danielhollas (changes requested):

aiidalab_widgets_base/utils/__init__.py
yakutovicha: We are in the alpha-releases stage. So when if not now 😄

aiidalab_widgets_base/bug_report.py
yakutovicha: The format checker said something like: "`i` was not used inside the loop, so add an underscore in front."

danielhollas: Fair, but still looks very confusing to me, never seen a variable (not method) prefixed with underscore. Maybe call it `lineno` or `line_number`?

yakutovicha:
> They both make sense to me, but I don't know where you see these violation so cannot comment if it makes sense to fix or ignore in our context.

Just have a look at the pre-commit ci. They are there.

aiidalab_widgets_base/bug_report.py
yakutovicha: > Fair, but still looks very confusing to me, never seen a variable (not method) prefixed with underscore. Maybe call it `lineno` or `line_number`?

You meant `_lineno`/`line_number`? I don’t think the name without an underscore will satisfy the checks.

Here are a few words about the use of underscore for the variable names: https://medium.com/python-features/naming-conventions-with-underscores-in-python-791251ac7097

danielhollas: Ah, I misunderstood then. So be it, but I'd still use a more descriptive variable name to make things clearer. Thanks!

In the article they explain the use the use of a leading underscore to indicate private vars / methods, which I was familiar with, but I don't think that is the case here so it still looks odd to me tbh.

danielhollas:
As the TC101 indicates, the `on_setup_computer()` function does look quite busy with three `try` statements. I don't know how to get rid of the first try, but I'd suggest to combine the second and the third in a way that TC101 suggests:

```python
        try:
            computer = computer_builder.new()
            computer.store()
        except (
            ComputerBuilder.ComputerValidationError,
            common.exceptions.ValidationError,
        ) as err:
            self.message = f"{type(err).__name__}: {err}"
            return False
        except common.exceptions.ValidationError as err:
            self.message = f"Unable to store the computer: {err}."
            return False
```

The `try` statements in `SmilesWidget.__init__()` cannot be combined so I'd suppress the error locally.

danielhollas:
@yakutovicha I would also suggest to add the [nbstripout](https://github.com/kynan/nbstripout) pre-commit, which automatically strips output cells from Jupyter notebooks. I am using it in my App and find it very useful, and @unkcpz recently added it to QeApp as well. 

https://github.com/ispg-group/aiidalab-ispg/blob/00f054b675902f2f5100d36e8858dae76d465257/.pre-commit-config.yaml#L14)

aiidalab_widgets_base/bug_report.py
yakutovicha: You are right. In the end, this is only to satisfy the check. Probably, this part of the code could be written better.

yakutovicha:
> @yakutovicha I would also suggest to add the [nbstripout](https://github.com/kynan/nbstripout) pre-commit, which automatically strips output cells from Jupyter notebooks. I am using it in my App and find it very useful, and @unkcpz recently added it to QeApp as well.
> 
> https://github.com/ispg-group/aiidalab-ispg/blob/00f054b675902f2f5100d36e8858dae76d465257/.pre-commit-config.yaml#L14)

Thanks, I've just added this.

yakutovicha:
> As for TC003, I don't know how to get rid of them since we're using the built-in exception classes. 🤷

I think they propose avoiding custom strings in exceptions completely. At least, this is what I understood looking at the [link](https://github.com/guilatrova/tryceratops/blob/main/docs/violations/TC003.md) you've shared above.

At the same time, I don't get why the following code receives complaints:
https://github.com/aiidalab/aiidalab-widgets-base/blob/87f5f4cc2b85bae36cff4a36e6feac1fd33d5c0d/aiidalab_widgets_base/wizard.py#L96-L98

Why the code below doesn't:
https://github.com/aiidalab/aiidalab-widgets-base/blob/87f5f4cc2b85bae36cff4a36e6feac1fd33d5c0d/aiidalab_widgets_base/viewers.py#L791-L794

aiidalab_widgets_base/utils/exceptions.py
```diff
@@ -0,0 +1,16 @@
+class AtLeastTwoStepsError(ValueError):
+    """Raised when the number of steps is less than two."""
```
danielhollas, line 2: Would be good to have a more thorough description here, it's not clear that this exception is for the WizardAppWidget. Which makes me think, since this exception is rather specific, it would make sense to put it in the same file as the widget itself imo.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -852,19 +852,26 @@ def _configure_computer(self, computer: orm.Computer):
 
         return True
 
+    def _run_things_on_existing_computer(self, label):
```
danielhollas, line 855: Could the name be a more descriptive. It is not clear at all what are the "things". 
What are the examples of functions that run on `_on_computer_success`?

aiidalab_widgets_base/computational_resources.py
```diff
@@ -852,19 +852,26 @@ def _configure_computer(self, computer: orm.Computer):
...
+        try:
+            orm.Computer.objects.get(label=label)
+            for function in self._on_setup_computer_success:
+                function()
+        except common.NotExistent:
+            return False
+        else:
+            return True
+
     def on_setup_computer(self, _=None):
```
danielhollas, line 866: I am a bit confused here. The function name is `on_setup_computer`, but the doc string says `Create a new computer`. What is the main purpose of this function?

aiidalab_widgets_base/computational_resources.py
danielhollas: Maybe `_run_callbacks_on_existing_computer()`?

danielhollas:
One of the tests seems to be broken, looks like the `structures.ipynb` notebook is not working.

unkcpz:
@yakutovicha I make a commit to fix the fail notebook test, hope you don't mind.

aiidalab_widgets_base/computational_resources.py
unkcpz: The main purpose of this function is when the "setup computer" button is clicked, setup (create) a new computer.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -852,19 +852,26 @@ def _configure_computer(self, computer: orm.Computer):
...
         """Create a new computer."""
         if self.label.value == "":  # check hostname
             self.message = "Please specify the computer name (for AiiDA)"
             return False
-        try:
-            computer = orm.Computer.objects.get(label=self.label.value)
-            self.message = f"A computer called {self.label.value} already exists."
```
danielhollas, line 862: This line is missing in the new function.

aiidalab_widgets_base/computational_resources.py
danielhollas: Hmm, but then should these registered function even run if the computer already exist?
Supposedly refresh would not be needed in this case...

I guess the name should be `_run_callbacks_if_computer_exists`, otherwise it implies we're are running something on the remote computers...

aiidalab_widgets_base/utils/exceptions.py
yakutovicha: I also thought about this, good that you've mentioned it. Agreed, moved to `wizard..py`.

aiidalab_widgets_base/computational_resources.py
yakutovicha: I've just put it back to a more appropriate place 👍

aiidalab_widgets_base/computational_resources.py
yakutovicha: > The main purpose of this function is when the "setup computer" button is clicked, setup (create) a new computer.

yep, that is correct.

danielhollas: @yakutovicha thanks for the context. Okay, in that case it would be great to expand the doc string on this function to explain when this function is called. This widget is doing so many things that it is beneficial for things to be documented well.

danielhollas (changes requested):
@unkcpz @yakutovicha I've looked through the code, and the only callbacks that seem to be registered in `_on_setup_computer_success` are calls that refresh the widgets' state. But in that case, I don't think these should be run if the computer already exists in the first place. Also, the name `on_setup_computer_success` implies to me that they should only be run when a new computer is created.

The same comment probably applies to `on_setup_code_success` callbacks.

Since this PR is about linters, I do not want to block it on this issue, but it should be addressed in a separate PR since we've already dug into it anyway.

unkcpz:
> I've looked through the code, and the only callbacks that seem to be registered in _on_setup_computer_success are calls that refresh the widgets' state. But in that case, I don't think these should be run if the computer already exists in the first place

Why you think this should not be run when the computer already exists? The function `_run_callbacks_if_computer_exists` is called in two condition, if the computer exists, it is run to update the computer dropdown of the code setup widget. If the computer not exists, it is run after the computer is create/setup and then update the computer dropdown widget of the code setup widget. Do you aware that the `refresh` is not refresh the computer setup widget but the code setup widget, maybe this is why you are confused?

aiidalab_widgets_base/computational_resources.py
```diff
@@ -892,24 +900,18 @@ def on_setup_computer(self, _=None):
...
-
-        if self._configure_computer(computer):
-            for function in self._on_setup_computer_success:
-                function()
-        self.message = f"Computer<{computer.pk}> {computer.label} created"
-        return True
+        if self._run_callbacks_if_computer_exists(self.label.value):
+            self.message = f"Computer<{computer.pk}> {computer.label} created"
+            return True
+        return False
```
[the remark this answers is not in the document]
unkcpz: I think this is a correct suggestion, I add it. Although I think this part will never touch, if the computer setup fail it will caught by the try.except statement.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -892,24 +900,18 @@ def on_setup_computer(self, _=None):
...
             self.message = f"{type(err).__name__}: {err}"
             return False
 
-        try:
-            computer.store()
-        except common.exceptions.ValidationError as err:
-            self.message = f"Unable to store the computer: {err}."
-            return False
-
-        if self._configure_computer(computer):
```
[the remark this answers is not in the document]
danielhollas: It would be good if you could verify that your new tests fail when this line is deleted. :-)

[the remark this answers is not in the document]
unkcpz: Great! that exactly manifest the bug. Add the configure function back to solve the issue. You can find the change from last commit https://github.com/aiidalab/aiidalab-widgets-base/pull/401/commits/1c568d38db453712c4f2c11ea862bb0621b036de

I also change a bit of the setup-configure flow, I combine the computer "setup" and "configure" in the block of a try, which only store the computer node when the setup and the configure are both successful. I did this because we only provide the widget called "Computer Setup" but no "computer configure". When the computer is setup and stored but the configure phase is not finished. The computer in the database has the UUID but leaves it in a very strange state and can not be used for further code setup.

danielhollas:
> Why you think this should not be run when the computer already exists? The function _run_callbacks_if_computer_exists is called in two condition, if the computer exists, it is run to update the computer dropdown of the code setup widget.

Well, my thinking was that if the computer already exists, it would have been in the Dropdown already so no need to refresh. But I guess the computer could be create somehow outside of the widget, making the UI not up to date. Anyway, feel free to ignore this.

danielhollas (approved):
@unkcpz thanks for finishing this!

unkcpz:
@danielhollas thanks for the review. @yakutovicha I'll let you have a final check and merge the PR.

yakutovicha:
All good, @unkcpz @danielhollas! Thanks a lot for your help!

aiidalab_widgets_base/computational_resources.py
```diff
@@ -835,14 +835,13 @@ def _configure_computer(self, computer: orm.Computer):
             "use_login_shell": self.use_login_shell.value,
             "safe_interval": self.safe_interval.value,
         }
-        if "user" in sshcfg:
```
danielhollas, line 838: @unkcpz I suspect this change broke the setup for local computer, reported in #417.
Could you please revert it and and a test for localhost computers?
(not a huge priority for me)

unkcpz: Ha! You are correct, that makes sense, I'll reconsider how this part of the code should work. It currently leads to two new small issues.

PR #409, opened 2022-12-07
Fix/406/default calc job plugin

### LINKED ISSUE #406: 'AiidaCodeSetup' object has no attribute 'default_calc_job_plugin'
When code setup button clicked, I got the exception.

```
AttributeError                            Traceback (most recent call last)
~/.local/lib/python3.9/site-packages/aiidalab_widgets_base/computational_resources.py in on_setup_code(self, _)
   1066             ]
   1067 
-> 1068             kwargs = {key: getattr(self, key).value for key in items_to_configure}
   1069 
   1070             # Checking if the code with this name already exists

~/.local/lib/python3.9/site-packages/aiidalab_widgets_base/computational_resources.py in <dictcomp>(.0)
   1066             ]
   1067 
-> 1068             kwargs = {key: getattr(self, key).value for key in items_to_configure}
   1069 
   1070             # Checking if the code with this name already exists

AttributeError: 'AiidaCodeSetup' object has no attribute 'default_calc_job_plugin'
```

unkcpz:
fixes #406

### DIFF
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -1069,6 +1069,9 @@ class AiidaCodeSetup(ipw.VBox):
 
             kwargs = {key: getattr(self, key).value for key in items_to_configure}
 
+            # convert computer to AiiDA node from its UUID
+            kwargs["computer"] = orm.load_computer(kwargs["computer"])
+
             # Checking if the code with this name already exists
             qb = orm.QueryBuilder()
             qb.append(
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -46,7 +46,7 @@ class ComputationalResourcesWidget(ipw.VBox):
     codes = traitlets.Dict(allow_none=True)
     allow_hidden_codes = traitlets.Bool(False)
     allow_disabled_computers = traitlets.Bool(False)
-    input_plugin = traitlets.Unicode(allow_none=True)
+    default_calc_job_plugin = traitlets.Unicode(allow_none=True)
 
     def __init__(self, description="Select code:", path_to_root="../", **kwargs):
         """Dropdown for Codes for one input plugin.
@@ -82,7 +82,7 @@ class ComputationalResourcesWidget(ipw.VBox):
 
         # Setting up codes and computers.
         self.comp_resources_database = ComputationalResourcesDatabaseWidget(
-            input_plugin=self.input_plugin
+            default_calc_job_plugin=self.default_calc_job_plugin
         )
 
         self.ssh_computer_setup = SshComputerSetup()
@@ -166,7 +166,7 @@ class ComputationalResourcesWidget(ipw.VBox):
             for c in orm.QueryBuilder()
             .append(
                 orm.Code,
-                filters={"attributes.input_plugin": self.input_plugin},
+                filters={"attributes.input_plugin": self.default_calc_job_plugin},
             )
             .all()
             if c[0].computer.is_user_configured(user)
@@ -188,9 +188,7 @@ class ComputationalResourcesWidget(ipw.VBox):
         with self.hold_trait_notifications():
             self.code_select_dropdown.options = self._get_codes()
             if not self.code_select_dropdown.options:
-                self.output.value = (
-                    f"No codes found for input plugin '{self.input_plugin}'."
-                )
+                self.output.value = f"No codes found for default calcjob plugin '{self.default_calc_job_plugin}'."
                 self.code_select_dropdown.disabled = True
             else:
                 self.code_select_dropdown.disabled = False
@@ -983,7 +981,7 @@ class AiidaCodeSetup(ipw.VBox):
         )
 
         # Code plugin.
-        self.input_plugin = ipw.Dropdown(
+        self.default_calc_job_plugin = ipw.Dropdown(
             options=sorted(
                 (ep.name, ep.name)
                 for ep in plugins.entry_point.get_entry_points("aiida.calculations")
@@ -1001,7 +999,7 @@ class AiidaCodeSetup(ipw.VBox):
             style=STYLE,
         )
 
-        self.remote_abs_path = ipw.Text(
+        self.filepath_executable = ipw.Text(
             placeholder="/path/to/executable",
             description="Absolute path to executable:",
             layout=LAYOUT,
@@ -1033,9 +1031,9 @@ class AiidaCodeSetup(ipw.VBox):
         children = [
             self.label,
             self.computer,
-            self.input_plugin,
+            self.default_calc_job_plugin,
             self.description,
-            self.remote_abs_path,
+            self.filepath_executable,
             self.use_double_quotes,
             self.prepend_text,
             self.append_text,
@@ -1044,10 +1042,10 @@ class AiidaCodeSetup(ipw.VBox):
         ]
         super().__init__(children, **kwargs)
 
-    @traitlets.validate("input_plugin")
-    def _validate_input_plugin(self, proposal):
+    @traitlets.validate("default_calc_job_plugin")
+    def _validate_default_calc_job_plugin(self, proposal):
         plugin = proposal["value"]
-        return plugin if plugin in self.input_plugin.options else None
+        return plugin if plugin in self.default_calc_job_plugin.options else None
 
     def on_setup_code(self, _=None):
         """Setup an AiiDA code."""
@@ -1114,7 +1112,7 @@ class AiidaCodeSetup(ipw.VBox):
         self.label.value = ""
         self.computer.value = ""
         self.description.value = ""
-        self.remote_abs_path.value = ""
+        self.filepath_executable.value = ""
         self.use_double_quotes.value = False
         self.prepend_text.value = ""
         self.append_text.value = ""
@@ -1130,7 +1128,7 @@ class AiidaCodeSetup(ipw.VBox):
             self._reset()
         for key, value in self.code_setup.items():
             if hasattr(self, key):
-                if key == "input_plugin":
+                if key == "default_calc_job_plugin":
                     try:
                         getattr(self, key).label = value
                     except traitlets.TraitError:
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -975,7 +975,8 @@ class AiidaCodeSetup(ipw.VBox):
             style=STYLE,
         )
 
-        # Computer on which the code is installed. Two dlinks are needed to make sure we get a Computer instance.
+        # Computer on which the code is installed. The value of this widget is
+        # the UUID of the selected computer.
         self.computer = ComputerDropdownWidget(
             path_to_root=path_to_root,
         )
@@ -1058,7 +1059,6 @@ class AiidaCodeSetup(ipw.VBox):
 
             items_to_configure = [
                 "label",
-                "computer",
                 "description",
                 "default_calc_job_plugin",
                 "filepath_executable",
@@ -1069,14 +1069,12 @@ class AiidaCodeSetup(ipw.VBox):
 
             kwargs = {key: getattr(self, key).value for key in items_to_configure}
 
-            # convert computer to AiiDA node from its UUID
-            kwargs["computer"] = orm.load_computer(kwargs["computer"])
+            # set computer from its widget value the UUID of the computer.
+            computer = orm.load_computer(self.computer.value)
 
             # Checking if the code with this name already exists
             qb = orm.QueryBuilder()
-            qb.append(
-                orm.Computer, filters={"uuid": kwargs["computer"].uuid}, tag="computer"
-            )
+            qb.append(orm.Computer, filters={"uuid": computer.uuid}, tag="computer")
             qb.append(
                 orm.InstalledCode,
                 with_computer="computer",
@@ -1084,12 +1082,12 @@ class AiidaCodeSetup(ipw.VBox):
             )
             if qb.count() > 0:
                 self.message = (
-                    f"Code {kwargs['label']}@{kwargs['computer'].label} already exists."
+                    f"Code {kwargs['label']}@{computer.label} already exists."
                 )
                 return False
 
             try:
-                code = orm.InstalledCode(**kwargs)
+                code = orm.InstalledCode(computer=computer, **kwargs)
             except (common.exceptions.InputValidationError, KeyError) as exception:
                 self.message = f"Invalid inputs: {exception}"
                 return False
--- a/notebooks/computational_resources.ipynb
+++ b/notebooks/computational_resources.ipynb
@@ -2,25 +2,10 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "id": "4013ce60",
    "metadata": {},
-   "outputs": [
-    {
-     "data": {
-      "application/javascript": [
-       "IPython.OutputArea.prototype._should_scroll = function(lines) {\n",
-       "    return false;\n",
-       "}\n"
-      ],
-      "text/plain": [
-       "<IPython.core.display.Javascript object>"
-      ]
-     },
-     "metadata": {},
-     "output_type": "display_data"
-    }
-   ],
+   "outputs": [],
    "source": [
     "%%javascript\n",
     "IPython.OutputArea.prototype._should_scroll = function(lines) {\n",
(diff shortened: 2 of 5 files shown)


danielhollas:
@unkcpz it would be great if you could somehow, if possible, extend the test so that it would have failed before this fix.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -45,7 +45,7 @@ class ComputationalResourcesWidget(ipw.VBox):
     codes = traitlets.Dict(allow_none=True)
     allow_hidden_codes = traitlets.Bool(False)
     allow_disabled_computers = traitlets.Bool(False)
-    input_plugin = traitlets.Unicode(allow_none=True)
+    default_calc_job_plugin = traitlets.Unicode(allow_none=True)
```
danielhollas, line 48: Why are we changing this traitlet? This imo unnecessarily breaks backward compatibility and makes an extra work for AiiDAlab apps' migration.

unkcpz: Okay, I think I see your point. Only mapping the interface when loaded from the database.

unkcpz: Hi @danielhollas, I gave it a try and think it is better to all change to `default_calc_job_plugin `. Since `orm.InstalledCode` only receive `default_calc_job_plugin` as parameter, I think it is better to keep the widget API match with this.

unkcpz:
> @unkcpz it would be great if you could somehow, if possible, extend the test so that it would have failed before this fix.

Agree, not sure how hard to add a unittest-like test to this widget. The tricky part is we use threading in the widget which to me is extremely mind-bending.

aiidalab_widgets_base/computational_resources.py
yakutovicha: I think those traits are supposed to mimic the aiida attributes of a `Code` object. So I would prefer that we change them to be compatible with the latest updates of AiiDA core.

unkcpz:
> so this PR depends on https://github.com/aiidateam/aiida-code-registry/pull/77, right?

Yes, and I have change the database URL. Once that merged, I'll change it to the aiidateam one.

unkcpz:
Hi @yakutovicha, I update this PR and you can give it a test. The known issue that we'll discuss in the meeting is that when quick setup button is clicked, the setup process will not pause in SSH setup but will skip to the computer setup and then code setup and get everything "done" but for sure the SSH is not properly setup. I test it with the old docker-stack and awb version 1.4.2 it has the same issue.

yakutovicha:
> Hi @yakutovicha, I update this PR and you can give it a test. The known issue that we'll discuss in the meeting is that when quick setup button is clicked, the setup process will not pause in SSH setup but will skip to the computer setup and then code setup and get everything "done" but for sure the SSH is not properly setup. I test it with the old docker-stack and awb version 1.4.2 it has the same issue.

For the record - we quickly discussed this topic during a zoom call. In fact, this is not a bug. It is okay to set up a computer without having ssh access to it. The ssh access can always be setup later.

yakutovicha (dismissed):
Tested locally, and seems to work fine 👍

tests/test_notebooks.py
```diff
@@ -43,6 +45,22 @@ def test_eln_import(selenium_driver):
...
+    command = f"exec --workdir /home/jovyan/apps/aiidalab-widgets-base -T aiidalab {install_command}"
+
+    docker_compose.execute(command)
+
+    url_with_token = urljoin(
+        url,
+        f"apps/apps/aiidalab-widgets-base/notebooks/computational_resources.ipynb?token={token}",
+    )
+    selenium.get(f"{url_with_token}")
+    selenium.implicitly_wait(10)  # must wait until the app loaded
```
danielhollas, line 63: This should be set in the `notebook_service` fixture in `conftest`, that way it will be applied to all tests. Same way as it is done in the QeApp:

https://github.com/aiidalab/aiidalab-qe/blob/50406cc3cf346d9dcfd123ddb5e7f0b345a0c313/tests/conftest.py#L55

tests/test_notebooks.py
```diff
@@ -43,6 +45,22 @@ def test_eln_import(selenium_driver):
...
 
-def test_computational_resources(selenium_driver):
-    driver = selenium_driver("notebooks/computational_resources.ipynb")
+def test_computational_resources(notebook_service, selenium):
+    """Test the quicksetup of the code"""
+    url, token, docker_compose = notebook_service
+
+    # install the aiida-quantumespresso, we need it for test of some widgets
+    install_command = "bash -c 'pip install -U aiida-quantumespresso==4.1.0'"
+    command = f"exec --workdir /home/jovyan/apps/aiidalab-widgets-base -T aiidalab {install_command}"
```
danielhollas, line 54: I would use the `aiidalab_exec` fixture from aiidalab-docker-stack 

https://github.com/aiidalab/aiidalab-docker-stack/blob/e89338181ada44420051e775645fdea00f91cff3/tests/conftest.py#L46

tests/test_notebooks.py
```diff
@@ -43,6 +45,22 @@ def test_eln_import(selenium_driver):
     driver.find_element(By.ID, "tooltip")
 
 
-def test_computational_resources(selenium_driver):
-    driver = selenium_driver("notebooks/computational_resources.ipynb")
+def test_computational_resources(notebook_service, selenium):
+    """Test the quicksetup of the code"""
+    url, token, docker_compose = notebook_service
+
+    # install the aiida-quantumespresso, we need it for test of some widgets
```
danielhollas, line 52: Is this needed? Couldn't we just use one of the sample CalcJobs from `aiida-core`?

tests/conftest.py
```diff
@@ -38,13 +38,13 @@ def notebook_service(docker_ip, docker_services):
...
     )
-    return url, token
+    return url, token, docker_compose
 
 
 @pytest.fixture(scope="function")
 def selenium_driver(selenium, notebook_service):
     def _selenium_driver(nb_path):
-        url, token = notebook_service
+        url, token, _ = notebook_service
```
danielhollas, line 47: I would use the aiidalab_exec fixture from aiidalab-docker-stack

https://github.com/aiidalab/aiidalab-docker-stack/blob/e89338181ada44420051e775645fdea00f91cff3/tests/conftest.py#L46

aiidalab_widgets_base/computational_resources.py
```diff
@@ -1071,6 +1069,9 @@ def on_setup_code(self, _=None):
 
             kwargs = {key: getattr(self, key).value for key in items_to_configure}
 
+            # convert computer to AiiDA node from its UUID
```
danielhollas, line 1072: I don't see the conversion to UUID here?

danielhollas (changes requested):

tests/test_notebooks.py
unkcpz: This is not true, the scope of `selenium` fixture is the "function", it will create for every test. Check here: https://github.com/pytest-dev/pytest-selenium/blob/47f4ef38f4c479c7fab0b9644043cafa9564edeb/src/pytest_selenium/pytest_selenium.py#L223-L225

tests/test_notebooks.py
unkcpz: Yes, I think that makes things easier. Do you know the different between using `bash -c` and not use it? I am confused about it and think we don't need it in all the case here. But why it is needed for `"bash -c 'echo \"${PATH}\"'"` in https://github.com/aiidalab/aiidalab-docker-stack/blob/e89338181ada44420051e775645fdea00f91cff3/tests/test_aiidalab.py#L139?

tests/test_notebooks.py
unkcpz: Since this computational resource widget will load the plugin from aiida-code-registry, there is no "dummy" calcjobs from aiida-core. You can search for `default_calc_job_plugin` in https://aiidateam.github.io/aiida-code-registry/database_v2_1.json

tests/test_notebooks.py
danielhollas: I think it is needed for escaping. Otherwise the `echo $PATH` does not work because the $PATH is not expanded.

aiidalab_widgets_base/computational_resources.py
unkcpz: Well, the `kwargs["computer"]` is a UUID of a computer. I admit this is confuse, maybe it is clear to name the key as "computer_uuid" then convert and pass the computer instance to the `InstalledCode`. 

All confusing from `orm.InstalledCode(**kwargs)`, I think it is better to make this clear by convert and passing computer as parameter explicitly.

tests/test_notebooks.py
unkcpz: Makes sense! I'll try here removing all `bash -c`, it should work if you're correct about this.

tests/test_notebooks.py
danielhollas: I am not sure I understand your point, but yes I meant to modify the `selenium_drive` fixture, sorry.

tests/test_notebooks.py
danielhollas: I don't think this is true. You should definitely be able to setup a code that is not in the registry. This is what I am doing with `aiida-orca`, which is not in the registry but the widget should still allow me to create it.

unkcpz: Interesting, I make a test and it truly works without the entrypoint is installed. 
I think we can not install this, that's great!

aiidalab_widgets_base/computational_resources.py
unkcpz: addressed.

tests/conftest.py
```diff
@@ -41,6 +54,19 @@ def notebook_service(docker_ip, docker_services):
...
+@pytest.fixture(scope="session", autouse=True)
+def install_package(docker_compose):
+    # assurance for host user UID other that 1000
+    command = build_command(
+        "chown -R jovyan:users /home/jovyan/apps/aiidalab-widgets-base", user="root"
+    )
+    docker_compose.execute(command)
+
+    # install the package
+    command = build_command("pip install -U .")
```
danielhollas, line 66: Do not use the `-U` upgrade option, that's not what `aiidalab install` does.

tests/test_notebooks.py
```diff
@@ -43,6 +43,35 @@ def test_eln_import(selenium_driver):
     driver.find_element(By.ID, "tooltip")
 
 
-def test_computational_resources(selenium_driver):
+def test_computational_resources(selenium_driver, aiidalab_exec):
```
danielhollas, line 46:
```suggestion
def test_computational_resources_code_setup(selenium_driver, aiidalab_exec):
```

tests/test_notebooks.py
```diff
@@ -43,6 +43,35 @@ def test_eln_import(selenium_driver):
...
+    driver.find_element(By.XPATH, '//option[text()="pw-7.0-multicore"]').click()
+
+    # fill the SSH username
+    driver.find_element(
+        By.XPATH, "//label[text()='SSH username:']/following-sibling::input"
+    ).send_keys("dummyuser")
+
+    # click the quick setup
+    driver.find_element(By.XPATH, '//button[text()="Quick Setup"]').click()
+    driver.implicitly_wait(1.0)
```
danielhollas, line 73: I don't think `implicitly_wait` does what you need here.
https://selenium-python.readthedocs.io/waits.html#implicit-waits

```suggestion
    time.sleep(1.0)
```

tests/test_notebooks.py
```diff
@@ -43,6 +43,35 @@ def test_eln_import(selenium_driver):
...
+
+    # check the code pw-7.0 is not in code list
```
danielhollas, line 75:
```suggestion
    # check the new code pw-7.0 is in code list
```

tests/conftest.py
```diff
@@ -41,6 +54,19 @@ def notebook_service(docker_ip, docker_services):
     return url, token
 
 
+@pytest.fixture(scope="session", autouse=True)
+def install_package(docker_compose):
+    # assurance for host user UID other that 1000
+    command = build_command(
```
danielhollas, line 60: I don't understand why are you not using the `aiidalab_exec` fixture here?

tests/conftest.py
```diff
@@ -17,20 +17,33 @@ def is_responsive(url):
...
-    """Ensure that HTTP service is up and responsive."""
+def docker_compose(docker_services):
+    return docker_services._docker_compose
 
-    docker_compose = docker_services._docker_compose
 
-    # assurance for host user UID other that 1000
-    chown_command = "exec -T -u root aiidalab bash -c 'chown -R jovyan:users /home/jovyan/apps/aiidalab-widgets-base'"
-    docker_compose.execute(chown_command)
+def build_command(command, user=None):
```
danielhollas, line 24: Does this need to be a separate function? Make more sense to me inside `aiidalab_exec` fixture.

danielhollas (changes requested):
This is great, thanks so much @unkcpz for adding the test!

One more ask, could you please also add a test for setting up a new computer?
I don't think you'll be able to use `core.transport.ssh`, but you should be able to create a new computer on localhost with `core.transport.local`.

unkcpz:
> One more ask, could you please also add a test for setting up a new computer?

I think this is included in the test of quick test. From start, there is no computer, and the quick setup will go through SSH setup -> computer setup -> code setup. Am I missing something?

tests/conftest.py
unkcpz: I think here we are not going to mimic what the `aiidalab install` is doing. For example when we bump the version of a dependency we want to see the effect of that.

tests/test_notebooks.py
unkcpz: However, I still do not quite understand the difference between their documentation. But I think you are right. 
We need anyway move to a more efficient "expected condition" wait.

tests/conftest.py
unkcpz: I tried but the "function" scope fixture is not allowed to be used for the "session" scope fixture.

unkcpz: But since we are execute the command for one container that we launched from start, I think maybe the best practice is to also set the `aiidalab_exec` as a session scope fixture? Then every thing is simple. I'll give it a try.

danielhollas:
@unkcpz I guess you are right. But we should still have tests for manual code and computer creation outside of quicksetup imo.

danielhollas:
Yes, I think that is what I did in QeApp PR.

Dne pá 9. 12. 2022 9:47 dop. uživatel Jusong Yu ***@***.***>
napsal:

danielhollas:
I think the main thing as I understand it is that Implicitly_wait function
does not wait on its own, it only influences selenium functions executed
after.

In this case you're not calling any selenium function so I don't think it
has any effect here.

Dne pá 9. 12. 2022 9:39 dop. uživatel Jusong Yu ***@***.***>
napsal:

unkcpz:
> @unkcpz I guess you are right. But we should still have tests for manual code and computer creation outside of quicksetup imo.

I understand, but the test will be all covered by this one since the way "quick setup" run is to fill up the computer setup widget. I personally don't see the point of a duplicate test. 

> Yes, I think that is what I did in QeApp PR. 

(This is reply for the aiidalab_exec I guess? Github folder the reply from the email). I change it to session scope and works fine. Cheers!

> I think the main thing as I understand it is that Implicitly_wait function
does not wait on its own, it only influences selenium functions executed
after.

If so, since the test passed in the first run, it should be safe not to wait here.

unkcpz:
> I don't think you'll be able to use core.transport.ssh, but you should be able to create a new computer on localhost with core.transport.local.

Actually, the current test is to use `core.transport.ssh` to create a computer, creating the computer does not need the SSH connection to be valid. The SSH setup is a tricky part and I think it require a well test but I don't know how at the moment. 

> I like your optimism. 😅

I regard it as a compliment 😆. I add it back. One second is a fair wait for this test.

tests/conftest.py
danielhollas: @unkcpz I think this function can now be moved to `aiidalab_exec` rather than being separete.

tests/conftest.py
```diff
@@ -41,6 +54,24 @@ def notebook_service(docker_ip, docker_services):
...
 
+@pytest.fixture(scope="session", autouse=True)
+def install_package(aiidalab_exec):
+    # assurance for host user UID other that 1000
+    # command = build_command(
+    #     "chown -R jovyan:users /home/jovyan/apps/aiidalab-widgets-base", user="root"
+    # )
+    # docker_compose.execute(command)
+
+    # # install the package
```
danielhollas, line 65: This should be deleted.

danielhollas (changes requested):

notebooks/computational_resources.ipynb
```diff
@@ -2,10 +2,25 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "id": "4013ce60",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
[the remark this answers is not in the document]
unkcpz: True! Forget to do the cleanup.

unkcpz:
@danielhollas thanks a lot for the check. I am careless and forget about clean the commit. Should be fine now.

tests/conftest.py
unkcpz: Removed.

tests/conftest.py
```diff
@@ -41,6 +54,24 @@ def notebook_service(docker_ip, docker_services):
     return url, token
 
 
+@pytest.fixture(scope="session", autouse=True)
+def install_package(aiidalab_exec):
```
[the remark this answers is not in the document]
unkcpz: Yes, I separate it for test purpose. Not necessary for this small test case. Changed.

danielhollas (commented):
@unkcpz thanks a lot for working on this, the new tests are important. :+1: 

I am withholding approve for now because I would like to merge the screenshot PR first and see how the screenshots look in this PR.

unkcpz:
Okay, no problem. Let me know when you finish that one I'll take a look.

unkcpz:
@danielhollas looks pretty good! 😄
image

danielhollas (approved):

PR #410, opened 2022-12-08
CI improvement and clean up

unkcpz:
This one is inspired by https://github.com/aiidalab/aiidalab-qe/pull/331 from @danielhollas 
The test on multiple python versions is useless since the nightly test is inside the container which as one specific python version.


danielhollas (approved):
Thanks @unkcpz !

PR #411, opened 2022-12-09
Take screenshots in notebook tests and upload them as artifacts

danielhollas:
@unkcpz here I took an alternative approach that you suggested at the QeApp PR, where I am using chmod instead of chown to fix AWB directory permissions inside the test container.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -20,7 +20,7 @@ def is_responsive(url):
 def screenshot_dir():
     # sdir = Path.joinpath(Path.home(), "screenshots")
     # TODO: Make this better
-    sdir = '/home/runner/work/aiidalab-widgets-base/aiidalab-widgets-base/screenshots'
+    sdir = "/home/runner/work/aiidalab-widgets-base/aiidalab-widgets-base/screenshots"
     os.mkdir(sdir)
     return sdir
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,9 +33,10 @@ def notebook_service(docker_ip, docker_services):
 
     docker_compose = docker_services._docker_compose
 
-    # assurance for host user UID other that 1000
-    chown_command = "exec -T -u root aiidalab bash -c 'chmod -R a+rw /home/jovyan/apps/aiidalab-widgets-base'"
-    docker_compose.execute(chown_command)
+    # Directory ~/apps/aiidalab-widgets-base/ is mounted by docker,
+    # make it writeable for jovyan user, needed for `pip install`
+    chmod_command = "exec -T -u root aiidalab bash -c 'chmod -R a+rw /home/jovyan/apps/aiidalab-widgets-base'"
+    docker_compose.execute(chmod_command)
 
     install_command = "bash -c 'pip install .'"
     command = f"exec --workdir /home/jovyan/apps/aiidalab-widgets-base -T aiidalab {install_command}"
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -16,17 +17,27 @@ def is_responsive(url):
...
+
 @pytest.fixture(scope="session")
 def notebook_service(docker_ip, docker_services):
     """Ensure that HTTP service is up and responsive."""
 
     docker_compose = docker_services._docker_compose
 
     # assurance for host user UID other that 1000
-    chown_command = "exec -T -u root aiidalab bash -c 'chown -R jovyan:users /home/jovyan/apps/aiidalab-widgets-base'"
+    chown_command = "exec -T -u root aiidalab bash -c 'chmod -R a+rw /home/jovyan/apps/aiidalab-widgets-base'"
```
unkcpz, line 37:
```suggestion
    chmod_command = "exec -T -u root aiidalab bash -c 'chmod -R a+rw /home/jovyan/apps/aiidalab-widgets-base'"
```

unkcpz: nitpick. We have `aiidalab_exec` soon by https://github.com/aiidalab/aiidalab-widgets-base/pull/409

tests/conftest.py
```diff
@@ -16,17 +17,27 @@ def is_responsive(url):
...
     docker_compose.execute(chown_command)
 
-    install_command = "bash -c 'pip install -U .'"
+    install_command = "bash -c 'pip install .'"
```
unkcpz, line 40: As I comment on https://github.com/aiidalab/aiidalab-widgets-base/pull/409, I think here we want to keep all dependencies updated, no?

Hmmm.... I think I understand your point, you are worry that the `-U` will conflict and override other dependencies? 
I think we need other tests to test the conflicting dependencies. Here we want to test the dependencies we set here is working.

unkcpz (changes requested):
@danielhollas thanks! I think this is the correct way to go, if there is no security issue of doing this. Just some minor requests.

unkcpz:
Can you also show here the pytest warnings before and after this fix?
I am still not sure why we need the write access for pytest.

danielhollas:
The write access is mainly needed for pip install. 

Pytest tries to write some kind of cache into .cache, you should see the warnings in any CI run outside of this PR. But it doesn't seem to be crucial.

tests/conftest.py
danielhollas: Yep, I am aware, I would like to merge this first so that we have screenshots in that PR. You can then create multiple screenshots alo the various stages of the new test.

tests/conftest.py
```diff
@@ -16,17 +17,27 @@ def is_responsive(url):
...
+    return sdir
+
+
 @pytest.fixture(scope="session")
 def notebook_service(docker_ip, docker_services):
     """Ensure that HTTP service is up and responsive."""
 
     docker_compose = docker_services._docker_compose
 
     # assurance for host user UID other that 1000
```
unkcpz, line 36: > The write access is mainly needed for pip install.

Okay. Can you update the comment here, it is outdated then.

tests/conftest.py
danielhollas: The dependencies should be satisfied even without the -U command. Since we're typically installing into a new Python environment, the -U shouldn't really have any effect. I am not quite sure how it interacts with pip cache, but we bust the cache anyway on any change to setup.cfg so it shouldn't matter.

danielhollas:
@unkcpz thanks for the review. :pray: I've incorporated your suggestions and further tweaked the screenshots. Some tests need to be improved but that's for another PR. This is good to go from my side. :rocket:

unkcpz (approved):
@danielhollas thanks! Looks very good, much appreciated that you also taking care of this. cheers!

PR #412, opened 2022-12-09
Delayed import of scikit-learn and pandas + other dependency tweaks

danielhollas:
Various tweaks to dependencies:

1. Delayed import of scikit-learn and bokeh. 
Both of these are very large packages that may [negatively impact startup time](https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html#top-level-python-code), so it makes sense to import them only when necessary. It would be great to get rid of them altogether. We use pandas only to generate HTML tables from python dictionaries. Perhaps there is a more lightweight solution for that. Sklearn is only used to do PCA in SmilesWidget, I'll try to remove it in another PR.

2. Declare rdkit and sklearn as optional dependencies for SmilesWidget. 

3. Add basic test for SmilesWidget

4. Updated test dependencies and moved them from `setup.cfg` to `requirements_test.txt`, same way as in QeApp. The pip installation on GHA went from 1 minute to 5 seconds so this is worthwhile.

5.  Removed stale `requirements.json` file. I am assuming it is not needed given the `setup.cfg` but correct me if I am wrong.

I added this to the milestone, since we're removing sklearn as a direct dependency, which might potentially break apps that relied on it. Note that SmilesWidget was already broken in principle, since rdkit is not installed in the new docker stack. At least now apps can reliably install needed dependencies by specifying `aiidalab-widgets-base[smiles]`.


danielhollas:
Tests are failing, not sure why. Will enable screenshots first in #411 to debug this.

unkcpz:
To reproduce the failing test, you can launch a docker container and run the failed notebook with the branch.

unkcpz:
Hi @danielhollas, just request review again when you think this is ready.

danielhollas:
@unkcpz I need to fix one of the XPath selector to fix the new test, but otherwise this is good to review.

setup.cfg
```diff
@@ -32,21 +32,20 @@ install_requires =
...
     optimade-client==2022.9.19
-    scikit-learn~=0.24
     spglib~=1.14
     vapory~=0.1.2
 python_requires = >=3.8
 include_package_data = True
 zip_safe = False
 
 [options.extras_require]
+smiles =
```
unkcpz, line 42: I am not sure about this, if you want "smiles" related packages installed by `aiidalab install`, it not supports install from extras dependencies yet.

setup.cfg
```diff
@@ -32,21 +32,20 @@ install_requires =
...
+    rdkit>=2021.09.2
+    scikit-learn~=0.24
 dev =
     bumpver==2021.1114
     pre-commit==2.10.1
-    pytest~=6.0
-    pytest-docker~=1.0
-    pytest-selenium~=4.0
-    webdriver-manager~=3.8
+test = file: requirements_test.txt
```
unkcpz, line 48: I guess this is leftover when copy paste from qeapp. Nice catch.

setup.cfg
danielhollas: That's okay I think, because I can declare the dependency in my setup.cfg with these extras dependencies, so it should pull the package from pip.

unkcpz: Ah, this is clever, but only for packages like `aiidalab-widgets-base` and when we not treat it as "app" anymore. Anyway, good for the change here.

unkcpz (approved):

unkcpz:
@danielhollas I think you have permission to merge, correct? I'll leave it to you in case you still want to add things. It is all good to me. 
This one actually bring the improvement of the performance for some widget, it reminds me that we want to add the benchmark test to the widgets, but we can do in a separate PR.

setup.cfg
danielhollas: Yes, that is a good point. But @yakutovicha said many times he wants to turn AWB into library anyways, so it should be okay.

danielhollas:
I don't have permission to merge actually. This is good to go from my side.

> This one actually bring the improvement of the performance for some widget, it reminds me that we want to add the benchmark test to the widgets, but we can do in a separate PR.

Yes, that would be great. I suspect this PR alone does not bring performance improvements much. Pandas is still imported by `optimade-client`. But I don't know how to measure this well.

danielhollas:
@yakutovicha can you confirm it's okay to delete `requirements.txt`?

.github/workflows/ci.yml
```diff
@@ -58,7 +58,7 @@ jobs:
                       **/requirements*.txt
 
             - name: Install dependencies for test
-              run: pip install -U -e .[dev]
+              run: pip install -r requirements_test.txt
```
yakutovicha, line 61: I don't understand the need for such a change. All `dev` requirements have to be installed here anyways. Why create a separate file for that?

.pre-commit-config.yaml
```diff
@@ -32,11 +32,6 @@ repos:
           - id: isort
             args: [--profile, black, --filter-files]
 
-    - repo: https://github.com/asottile/setup-cfg-fmt
-      rev: v2.2.0
-      hooks:
-          - id: setup-cfg-fmt
-
```
yakutovicha, line 39: Why do we remove the `setup.cfg` formatting?

setup.cfg
yakutovicha, line 48: I don't get why we need to separate `test` from `dev`. They always go together.

yakutovicha (commented):
I am a bit concerned by some changes in this PR. Specifically, I don't see a need of separating `dev` into `dev` and `testing`. All packages will be installed during the CI run anyways. Some while ago we actually did the opposite thing: we merged them together into `dev`.

.github/workflows/ci.yml
danielhollas: The issue is that pip cannot install *only* the dev requirements, it installs all the main requirements *and* Dev. But the tests themselves run in the container so we don't need to install AWB here. This change saves 1 minute of CI time per job, which is significant given the overall time is around 4 minutes.

I have no issue of having `dev` together with `test`. Although that would also mean we install test requirements in the pre-commit CI job.

.pre-commit-config.yaml
danielhollas: It currently doesn't support the 'file: ' syntax.

https://github.com/asottile/setup-cfg-fmt/issues/161

Since we want to move to project.toml anyway I think it's okay to delete it. I looked at what it does and it doesn't seem so useful when the setup.cfg is mostly stable. (and in general I hate when linters become obstacles instead of being helpful). But I am happy to concede this point.

.github/workflows/ci.yml
yakutovicha: I think we **do** want to test the installation. What would it be for a test otherwise?

I would rather work on python dependencies. As you've pointed out `pandas` takes quite a lot of time.

unkcpz: > @unkcpz I assume this was your motivation in QeApp as well?

This is one aspect for QeApp. It has another problem if install with `pip install .[test]` in QeApp that it will check the dependencies and the QeAppWorkChain may lead to the conflicting. 

> I think we do want to test the installation. What would it be for a test otherwise?

The test is done inside the container by `aiidalab_exec` fixture.

yakutovicha: > The test is done inside the container by `aiidalab_exec` fixture.

This is done for the basic testing of the main features of AWB (I guess it is classified as _smoke testing_).

However, we plan to add unit tests to check individual components of the package that @danielhollas was mentioning during the AiiDAlab meeting. That won't run inside the container, right? The individual functions will be executed with a specific input and checked against their output. The package dependencies are still needed.

yakutovicha: The other option would be to run **everything** inside the container. I think this is doable as well. There is no need to run pytest outside of the docker container. I guess that everything (chrome, firefox, pytest, awb, notebook) could be kept inside of the container and checked from there.

.pre-commit-config.yaml
danielhollas: I put the pre-commit hook back.

yakutovicha (approved):

PR #415, opened 2022-12-12
Manifest the issue of setting multiple codes

### LINKED ISSUE #416: The computer widget of code setup is not updated and will throw UUID not found error
When setting multiple codes in one remote computer using the quick setup, once the first code is setup, selecting the second code from the domain dropdown will cause the issue to complain that the computer dropdown widget has no value. 
This is rooted in `getattr(self, key).value = computer.uuid` of `_observe_code_setup` method of `AiidaCodeSetup` widget, that when domain is select the `code_setup` traitlet will trigger this function to set the value of computer setup widget. 
However, computer setup widget is not refreshed so the `computer.uuid` is not a valid value for the widget. 

The solution will be refresh the computer widget inside this `_observe_code_setup`.

unkcpz:
fixes #416

### DIFF
--- a/tests/test_notebooks.py
+++ b/tests/test_notebooks.py
@@ -94,7 +94,7 @@ def test_computational_resources_code_setup(
     output = aiidalab_exec("verdi code list").decode().strip()
     assert "pw-7.0@daint-mc" in output
 
-    # Set the second code of the same comptuer
+    # Set the second code of the same computer
     # issue https://github.com/aiidalab/aiidalab-widgets-base/issues/416
     # click the "Setup new code" button
     driver.find_element(By.XPATH, '(//button[text()="Setup new code"])[2]').click()
--- a/tests/test_notebooks.py
+++ b/tests/test_notebooks.py
@@ -70,29 +70,58 @@ def test_computational_resources_code_setup(
     driver = selenium_driver("notebooks/computational_resources.ipynb")
 
     # click the "Setup new code" button
-    driver.find_element(By.XPATH, '//button[text()="Setup new code"]').click()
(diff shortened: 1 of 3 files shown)


tests/test_notebooks.py
```diff
@@ -70,29 +70,58 @@ def test_computational_resources_code_setup(
...
     # click the quick setup
-    driver.find_element(By.XPATH, '//button[text()="Quick Setup"]').click()
+    driver.find_element(By.XPATH, '(//button[text()="Quick Setup"])[1]').click()
     time.sleep(1.0)
 
     # check the new code pw-7.0@daint-mc is in code list
     output = aiidalab_exec("verdi code list").decode().strip()
     assert "pw-7.0@daint-mc" in output
 
+    # Set the second code of the same comptuer
```
danielhollas, line 97:
```suggestion
    # Set the second code of the same computer
```

danielhollas (dismissed):
Was this bug introduced during the migration, or was it there before?

Thanks for the new tests, I assume they would fail before the fix.

unkcpz:
> Was this bug introduced during the migration, or was it there before?

I think it will not raised previously, it was introduced by https://github.com/aiidalab/aiidalab-widgets-base/pull/410, the last commit I added to bring the `_configure_computer` back. In the previous case, if no SSH username was provided, the computer still get setup and configured but setting SSH again will not make it work afterward. Because the computer configuration need the SSH username updated.

> Thanks for the new tests, I assume they would fail before the fix.

Yes, it would fail. This is the very positive side to have such a test framework. What a relief!

PR #422, opened 2022-12-17
Tests: Get screenshot at test teardown - approach 2

### LINKED ISSUE #419: When tests fail, the images are not uploaded as artifacts

danielhollas:
This is an alternative PR to #421, see that PR for a detailed problem description. These two PRs should be reviewed together and it should be decided which approach is preferred. 

Here to take a final screenshot, we modify the existing `selenium_driver` fixture. Screenshot name is passed in as an optional parameter when getting the driver.

I don't have a strong preference. I think a like the approach in #421 (having a dedicated new screenshot fixture) a bit more, because `selenium_driver` fixture is already quite busy and here we're making it more complicated. 

Closes #419

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -67,9 +67,13 @@ def notebook_service(docker_ip, docker_services, aiidalab_exec):
 
 
 @pytest.fixture(scope="function")
-def selenium_driver(selenium, notebook_service):
-    def _selenium_driver(nb_path):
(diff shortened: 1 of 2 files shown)


danielhollas:
Looks like approach in #421 is preferred so I will close this.

PR #424, opened 2022-12-30
CI: Configure pre-commit schedule to run quarterly

danielhollas:
I find the constant barrage of pre-commit PRs from various repos distracting / annoying, and not particularly useful. I'd propose to run them only quarterly, as was already done in a couple of repositories, such as `aiidalab-launch` and `aiidalab-docker-stack`. WDYT? If you agree I'll do the same change for `aiidalab` and `aiidalab-home` repos as those are the ones I am currently subscribed to.

https://github.com/aiidalab/aiidalab-launch/pull/143

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,7 @@
 ---
+ci:
+    autoupdate_schedule: quarterly
+
 repos:
     - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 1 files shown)


yakutovicha (approved):
@danielhollas I agree. Updating those things every week is too much. It creates more work for us without delivering too much value.

Please go ahead with updating the other repositories.

PR #427, opened 2023-02-04
ProcessMonitorWidget: Print traceback for callback exceptions

### LINKED ISSUE #386: Print tracebacks for disabled callbacks in ProcessMonitorWidget
Currently when an exception is thrown from a callback in the ProcessMonitorWidget, the exception is swalled, the callback is disabled and a warning is printed, with the exception itself, but without any traceback. This makes it extremely hard to debug issues while developing the app. Two issues

1. Without traceback, it's often hard to know where the exception is coming from
2. In the appmode, the warning, printed via `warnings.warn()` is not shown at all. I am not sure if that's the best design.

danielhollas:
Closes #386  (see the issue for details why this change is needed)

### Before:
obrazek

### After:

obrazek

### NOTES:
1. The warning is hidden in the Appmode. This can easily lead to user confusion, since something doesn't work but the error is hidden.  Not sure if this is the best design, perhaps these exceptions should be raised instead. But that's beyond the point of this PR.
2. This fix could (should?) be applied to the 1.x support branch.


danielhollas:
@yakutovicha here's a code snippet to test this in Jupyter Notebook

```python
from aiida.orm import load_node
from aiidalab_widgets_base import ProcessMonitor

def raise_error():
    raise ValueError("We have a problem")

monitor = ProcessMonitor(
            timeout=0.5,
            callbacks=[
                raise_error,
            ],
)
process_pk = 1217
monitor.value = load_node(process_pk).uuid
```

aiidalab_widgets_base/process.py
```diff
@@ -747,9 +748,9 @@ def _run(funcs):
                         func(process_uuid)
                     else:
                         func()
-                except Exception as error:
+                except Exception:
                     warnings.warn(
-                        f"WARNING: Callback function '{func}' disabled due to error: {error}"
+                        f"WARNING: Callback function {func.__name__!r} disabled due to error:\n{traceback.format_exc()}"
```
yakutovicha, line 753:
```suggestion
                        f"WARNING: The callback function {func.__name__!r} was disabled due to an error:\n{traceback.format_exc()}"
```

yakutovicha (changes requested):
Thanks, @danielhollas. I just have a minor request.

danielhollas:
@yakutovicha thanks, I've accepted your suggestion.

Once this is merged, could you please release a new beta version? I want to use this for better in my app for easier debugging.
Also the last beta version is broken for my app so it would be nice to publish the `configure_tabs` fix.

yakutovicha (approved):

PR #429, opened 2023-02-09
Fix/configure tabs

### LINKED ISSUE #425: `configuration_tabs` option of StructureDataViewer is broken
PR #371 broke the `configuration_tabs` option of the StructureDataViewer.  In the latest release, when this option is passed to the class constructure, no configuration tabs are displayed.

I am pretty sure because of this deleted if statement

https://github.com/aiidalab/aiidalab-widgets-base/pull/371/files#diff-5aac3a07d48497df9963232d6dff4bbeb87f5a6992192cae0ed1eecd7d6573cdL217

Putting the `if configuration_tabs:` statement back fixes the issue for me. 

(Note that this broke my app so putting high urgency on this...)

AndresOrtegaGuerrero:
fixes #425

### DIFF
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -214,7 +214,6 @@ class _StructureDataBaseViewer(ipw.VBox):
         if configuration_tabs is None:
             configuration_tabs = ["Selection", "Appearance", "Cell", "Download"]
         if len(configuration_tabs) != 0:
-            self.selection_tab_idx = 0
             self.configuration_box = ipw.Tab(
                 layout=ipw.Layout(flex="1 1 auto", width="auto")
             )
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/viewers.py
```diff
@@ -213,7 +213,8 @@ def __init__(
         # Constructing configuration box
         if configuration_tabs is None:
             configuration_tabs = ["Selection", "Appearance", "Cell", "Download"]
-
+        if len(configuration_tabs) != 0:
+            self.selection_tab_idx = 0
```
yakutovicha, line 217: I assume 0 is anyways the default, so we might just completely remove the line:
```suggestion
```

yakutovicha (commented):
Thanks, @AndresOrtegaGuerrero. One minor thing.

aiidalab_widgets_base/viewers.py
danielhollas: That's not a correct assumption, `configuration_tab` is a constructor parameter so the Selection Tab index can be whatever, the selection tab might not even be there. 

That said, you've removed the `self.selection_tab_idx` completely in your previous PR, so this line should be removed anyway.

yakutovicha (approved):
Thanks, @AndresOrtegaGuerrero!

PR #430, opened 2023-02-10
register viewer widget with customized key and allow fallback support

unkcpz:


aiidalab_widgets_base/viewers.py
```diff
@@ -71,7 +74,12 @@ def viewer(obj, downloadable=True, **kwargs):
         return obj
 
     try:
-        _viewer = AIIDA_VIEWER_MAPPING[obj.node_type]
+        try:
+            _viewer = AIIDA_VIEWER_MAPPING[(obj.node_type, obj.process_label)]
```
danielhollas, line 78: I am wondering, to make this more general we could use the `obj.label` instead of `obj.process_label`. I am just not sure if that would be a backwards compatible change for QeApp.

unkcpz: Thanks for looking at this. I made this a draft since I am not decide if using `process_label` is a good design. I was thinking to introduce a rule that stores some standard metadata to `extra` attributes for use here.
You mentioned in the meeting yesterday that there is a new field that can be used for mapping the process name in TreeNodeWidget, what is that? Is it specific for process or for more general data node?

danielhollas: No, it's just a hook for renaming the `process_label`, see

https://github.com/ispg-group/aiidalab-ispg/blob/7895ed6ce9aa869e1c3a0803064a22a2d2a83fec/workflows/aiidalab_atmospec_workchain/__init__.py#L109

Implemented here:

https://github.com/aiidateam/aiida-core/pull/5713

It would definitely help to make the node tree more user friendly, but is not needed for this particular feature.

danielhollas:
I think this has been implemented (in a slightly different way) in #541

Issue #426, opened 2023-01-30
Make some changes in AWB 2.0 backwards compatible


danielhollas:
I've been thinking about how the rest of AiiDAlab apps will need to migrate to our new stack and AiiDA-2.0. From our experience with QeApp and my application we know that it is a bit of a pain, to say the least. Therefore I'd suggest we try to make it easier for others to migrate, otherwise we risk Python2-Python3 style split in the ecosystem.

One concrete suggestion specifically for AWB would be to make certain things backwords compatible. 
Concretely, some widgets were changed to use UUID in traitlets instead of AiiDA Nodes. That was a neccessary change, but we can easily make it backwards compatible by preserving the original traitlets, and making a small shim which converts them to UUID.

What do you think @yakutovicha @unkcpz. (LMK if my suggestion is not clear).

danielhollas:
To make this more specific, here's a code snippet to make the ProcessMonitorWidget backwards compatible.

```python
class ProcessMonitor()
    process = traitlets.Instance(ProcessNode, allow_none=True)

    @traitlets.observe("process")
    def _observe_process(self, change):
        # TODO: Perhaps issue a deprecation warning here
        process = change["new"]
        if process is None:
             self.value = None
             return
        self.value = process.uuid
```

@yakutovicha @unkcpz could you take a brief look at this issue if you have a bit of time? Would be nice to discuss on today's meeting.

yakutovicha:
I would prefer not to do that, tbh. Since the work is done in awb (which was the largest chunk of the work), migrating the apps should not be a problem. At this point, maintaining the backwards compatibility feels like an additional hurdle.

danielhollas:
> Since the work is done in awb (which was the largest chunk of the work), migrating the apps should not be a problem.

I am not sure we're understanding each other. The problem is we changed the input traitlets for some of the widgets, which actually makes the migration harder for the Apps developers. So to alleviate this burden I propose this workaround, because I feel there's already quite some work to migrate anyway.

unkcpz:
I think this is what we need to finish before we make the official release. We do this once or if we ignore it now it will haunt there and make widgets not very intuitive to be reused. 
Maybe I am naive, I think the only thing to do is to rename the traitlets. Then from QeApp and @danielhollas's app we update to be compatible with the new interface. 
Before the release of AWB `2.0.0` we still have a chance to do it aggressively.

danielhollas:
[...]

PR #433, opened 2023-02-20
process followers widget need interface updated

unkcpz:
When running `process.ipynb` with a real process, the exception raised that the interface calling `ProcessMonitorWidget` is incorrect.
I update the `ProcessFollowerWidget` interface to use the `value` for process uuid as well. 
This brings the issue https://github.com/aiidalab/aiidalab-widgets-base/issues/426 to more higher priority that we what having all such widget have a new interface which is more designable in my opinion.

### DIFF
--- a/aiidalab_widgets_base/process.py
+++ b/aiidalab_widgets_base/process.py
@@ -230,7 +230,7 @@ class ProcessOutputsWidget(ipw.VBox):
 class ProcessFollowerWidget(ipw.VBox):
     """A Widget that follows a process until finished."""
 
-    process = Instance(ProcessNode, allow_none=True)
+    value = traitlets.Unicode(allow_none=True)
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


danielhollas:
Thanks @unkcpz. I've added a "blocked" label until we discuss this issue more generally. The question to me is whether we want to make further breaking changes or not. The fix in this PR can certainly be done in a way that doesn't break existing apps.

unkcpz:
Okay, no problem. Let's find time to discuss this with @yakutovicha together. I'll separate some pre-commit fixes out and make a separate PR.

danielhollas:
@unkcpz can you create a corresponing issue for this and add it to the 2.0.0 milestone?

unkcpz:
This is fixed in https://github.com/aiidalab/aiidalab-widgets-base/pull/455

PR #435, opened 2023-02-22
Support the inputs namespace for process inputs check widget

### LINKED ISSUE #434: ProcessInputWidget do not support namespace inputs
The namespace input is an `AttributDict` which not supported by the `ProcessInputWidget` since it tries to show the `pk` of the input nodes. 
The workaround will be checking if the input port is the namespace the widget disables the display of the input showing. 
To fix this, we need to support the namespace also for the input ports. I think it make sense to show in the dropdown list the `<namespace>.<embed_port>`. For example, if we have inputs as following:

```
Inputs                PK    Type
--------------------  ----  -------------
pw
    pseudos
        Si            99    UpfData
    code              215   InstalledCode
    parallelization   220   Dict
    structure         218   StructureData
    parameters        232   Dict
clean_workdir         233   Bool
kpoints_distance      221   Float
kpoints_force_parity  222   Bool
max_iterations        223   Int
```

The dropdown list will having `pw.pseudos.Si`, `pw.code` etc..

unkcpz:
fixes #434 

- [x] Full implementation to support namespace
- [x] Import achive with process and smoke test the widgets (Instead we mock a node class with minimal requirements for testing purpose.)


unkcpz:
The implementation is finished, the pre-commit failed by the same reason as https://github.com/aiidalab/aiidalab-widgets-base/pull/433 caused by `warn.warning` clauses. I'll separate the pre-commit fixes into a PR.
For this one, I want to use it as an example to implement the unit test for widgets. I can mock a workchain process as did in aiida-quantumespresso. Pass the process to the widget and interactive with it through the traitlets and detect the change.

.github/workflows/ci.yml
```diff
@@ -89,3 +89,41 @@ jobs:
...
+                python-version: ['3.8', '3.9', '3.10']
+            fail-fast: false
+
+        runs-on: ubuntu-latest
+        timeout-minutes: 30
+
+        steps:
+
+            - name: Check out app
+              uses: actions/checkout@v3
```
yakutovicha, line 109: This is sort of general question. The `pre-commit` action was executed before and this step was there too. SInce `test-package` depends on `pre-commit`, why do we rerun this step then?

If we redo it anyways, what is the point of `needs`?

aiidalab_widgets_base/process.py
```diff
@@ -1,6 +1,8 @@
 """Widgets to work with processes."""
 # pylint: disable=no-self-use
 # Built-in imports
+from __future__ import annotations
```
yakutovicha, line 4: I think this is needed only for python<3.7, no?

tests/test_process.py
```diff
@@ -0,0 +1,176 @@
+import os
+import shutil
+from collections.abc import Mapping
+
+import pytest
+from aiida import orm
+
+pytest_plugins = ["aiida.manage.tests.pytest_fixtures"]  # pylint: disable=invalid-name
```
yakutovicha, line 8: We do not have pyliint.
```suggestion
pytest_plugins = ["aiida.manage.tests.pytest_fixtures"]
```

yakutovicha (changes requested):
Thanks, @unkcpz. I tested it, looks all good. I only have a few minor suggestions and one question.

.github/workflows/ci.yml
unkcpz: From the doc https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds, what I understand is it only requires the job to be passed but not inherit the state/environment of the job. The jobs are still independent of the environment.

aiidalab_widgets_base/process.py
unkcpz: For the `annotations` module itself, it is deprecated on a much later version as show in https://docs.python.org/3/library/__future__.html#id1

Here I use this for the type hints such as `dict[str, str]` rather than `Dict`. I'll give it a test and see if I can drop this import.

.github/workflows/ci.yml
```diff
@@ -89,3 +89,41 @@ jobs:
...
+
+            - name: Install package
+              run: pip install -e .[dev]
+
+            - name: Run pytest
+              run: pytest -v tests
+              env:
+                  TAG: ${{ matrix.tag }}
+
+            # code coverage
```
unkcpz, line 129: @yakutovicha I'll let you decide if you want to include the code coverage from here as well. I personally prefer to only to measure the coverage from only the standard unit test. But it is worth to have a try to get the coverage from the notebook smoke test, which will quickly increase the coverage by the black box test cases.

aiidalab_widgets_base/process.py
yakutovicha: ok, makes sense 👍

.github/workflows/ci.yml
yakutovicha: I propose to complete this in #441

yakutovicha (approved):
Thanks a lot, @unkcpz looks good to me.

Also, special thanks for splitting notebook tests into separate "smoke" tests. I think it makes a lot of sense 👍

PR #441, opened 2023-03-07
Measure the test coverage

### LINKED ISSUE #440: Measure the test coverage
For this library, we should also measure the coverage, similar to aiidalab-launch and eventually aim for test coverage of at least 75-80%. I would consider this issue completed if we reach a coverage of 15%.

yakutovicha:
fixes #440


yakutovicha:
Since the [reported coverage](https://app.codecov.io/gh/aiidalab/aiidalab-widgets-base/tree/feature%2Fcoverage-reports) is well beyond my original goal of 10-15% (as stated in the issue), I am making this PR ready for the review.

tests/test_structures.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # There should be no structures in the database, only the default output "Select structure".
+    assert len(structure_browser_widget.results.options) == 1
+
+    # Store the structure and check that it is listed in the widget.
+    structure_data_object.store()
+    structure_browser_widget.search()
+    assert len(structure_browser_widget.results.options) == 2
+
+    print(structure_browser_widget.results.options)
```
unkcpz, line 73:
```suggestion
```

tests/test_structures.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@pytest.mark.usefixtures("aiida_profile_clean")
+def test_structure_manager_widget(structure_data_object):
+    """Test the `StructureManagerWidget`."""
+    import aiidalab_widgets_base as awb
+
+    structure_manager_widget = awb.StructureManagerWidget(
+        importers=[
+            awb.StructureUploadWidget(title="From computer"),
```
unkcpz, line 13: Since we are testing the widget here in general, it should be possible to test tab names are correctly set as "From computer" and "Basic Editor".

tests/test_structures.py
```diff
@@ -0,0 +1,73 @@
...
+    structure_manager_widget.store_structure()
+    assert structure_manager_widget.structure_node.is_stored
+    assert structure_manager_widget.structure_node.pk is not None
+
+    # Simulate the structure modification.
+    new_structure = structure_manager_widget.structure.copy()
+    new_structure[0].position += [0, 0, 1]
+
+    structure_manager_widget.structure = new_structure
+    assert structure_manager_widget.structure_node.pk is None
```
unkcpz, line 34: I don't understand why this will be None?

unkcpz (changes requested):
@yakutovicha thanks for adding more tests! I have minor requests and things are not clear to me.

tests/test_structures.py
yakutovicha: Because the structure is not the saved one anymore, since it was modified. But I am not sure it is guaranteed to have `pk==None` for an unsaved structure. Probably, a better option would be to do the following:
```suggestion
    assert not structure_manager_widget.structure_node.is_stored
```

tests/test_structures.py
yakutovicha: After looking at the code - not really. Both editors and importers are set to local variables. It's not easily possible to access their values 🤷 .

unkcpz: Okay, then never bother with that.

tests/test_structures.py
unkcpz: I see. I think the change you proposed may not perfect. If I understand correctly if the `new_structure` is stored, the `structure_node` will become the stored node? Such as:

```python
    structure_manager_widget.structure = new_structure
    new_structure.store()
    assert structure_manager_widget.structure_node.is_stored is True
``` 

I think checking pk is fine, please just add a comment on it.

unkcpz (approved):
Looks all good.

tests/test_structures.py
unkcpz: Ah, you already did the change. Then it is fine, I have no preference, the comment is also not very necessary I think.

danielhollas:
@unkcpz @yakutovicha this is great, thanks! :clap:

PR #442, opened 2023-03-07
Resolve new flake8-bugbear error B028

yakutovicha:


danielhollas:
@yakutovicha why wasn't this called earlier? I thought we pin the flake8 version so this shouldn't happen just by itself?

yakutovicha:
> @yakutovicha why wasn't this caught earlier? I thought we pin the flake8 version so this shouldn't happen just by itself?

It's the new version of the flake8 plugin `flake8-bugbear`. We do not fix that one. Shall we?

danielhollas:
> It's the new version of the flake8 plugin flake8-bugbear. We do not fix that one. Shall we?

Yeah, if possible we should perhaps fix the plugin versions to avoid this in the future.

danielhollas (dismissed):

yakutovicha:
> Feel free to merge this and pin the plugin versions later so we don't get distracted.

already done...

> btw: For my app I was considering using ruff, as a replacement for both flake8 and its plugins and isort.

if you think it is a better alternative, please make an issue. I am not sure what are the advantages, but happy to learn more about them.

unkcpz:
Thanks @yakutovicha. This is what I mentioned gonna separate from https://github.com/aiidalab/aiidalab-widgets-base/pull/433#issuecomment-1440446993

yakutovicha:
> Thanks @yakutovicha. This is what I mentioned gonna separate from [#433 (comment)](https://github.com/aiidalab/aiidalab-widgets-base/pull/433#issuecomment-1440446993)

Oh, sorry. I didn't know you've done that already.

unkcpz:
Nonono, I need feel sorry that didn't manage to separate it. Thanks for doing this! 😉

PR #444, opened 2023-03-08
Run pytest inside the docker container

### LINKED ISSUE #443: Run pytest inside the container
In the current testing schema, the AiiDAlab docker container is used to test the behaviour of the widgets. The disadvantage of such an approach is the inability to measure the test coverage.

I propose launching the tests from inside the container. That should simplify the coverage measure.

yakutovicha:
fixes #443

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,75 +7,54 @@ on: [push, pull_request]
 
 jobs:
 
-    pre-commit:
-    # Adapted from: https://github.com/CasperWA/voila-optimade-client
+    build-docker-container:
 
         runs-on: ubuntu-latest
 
         steps:
             - uses: actions/checkout@v3
 
-            - name: Setup Python
-              uses: actions/setup-python@v4
+            - name: Build Docker image
+              uses: docker/build-push-action@v2
               with:
-                  python-version: '3.10'
-                  cache: pip
-                  cache-dependency-path: |
-                      .pre-commit-config.yaml
-                      **/setup.cfg
-                      **/pyproject.toml
-                      **/requirements*.txt
-
-            - uses: pre-commit/action@v2.0.0
+                  context: tests
+                  push: false
+                  tags: aiidalab/full-stack-with-browsers:latest
 
     test-notebooks:
 
-        needs: [pre-commit]
+        needs: [build-docker-container]
 
         strategy:
             matrix:
                 tag: [latest]
-                browser: [Chrome, Firefox]
+                browser: [Chrome]
                 python-version: ['3.10']
             fail-fast: false
 
         runs-on: ubuntu-latest
+        container:
+            image: aiidalab/full-stack-with-browsers:latest
+            env:
+                RMQHOST: messaging
+                TZ: Europe/Zurich
+                DOCKER_STACKS_JUPYTER_CMD: notebook
+                SETUP_DEFAULT_AIIDA_PROFILE: true
+                JUPYTER_TOKEN: my_test_token
+            volumes:
+                - ./:/home/jovyan/apps/aiidalab-widgets-base
         timeout-minutes: 30
 
         steps:
 
-            - name: Check out app
-              uses: actions/checkout@v3
+            - name: Checkout repository
+              uses: actions/checkout@v2
 
-            - name: Set up Python
-              uses: actions/setup-python@v4
-              with:
-                  python-version: ${{ matrix.python-version }}
-                  cache: pip
-                  cache-dependency-path: |
-                      **/setup.cfg
-                      **/pyproject.toml
-                      **/requirements*.txt
+            - name: Enter the repository
+              run: cd apps/aiidalab-widgets-base
 
             - name: Install package
-              run: pip install -e .[dev]
-
-            - name: Set jupyter token env
-              run: echo "JUPYTER_TOKEN=$(openssl rand -hex 32)" >> $GITHUB_ENV
-
-            # The Firefox and its engine geckodrive need do be installed manually to run
-            # selenium tests.
-            - name: Install Firefox
-              uses: browser-actions/setup-firefox@latest
-              with:
-                  firefox-version: '96.0'
-              if: matrix.browser == 'Firefox'
-
-            - name: Install geckodriver
-              run: |
-                  wget -c https://github.com/mozilla/geckodriver/releases/download/v0.30.0/geckodriver-v0.30.0-linux64.tar.gz
-                  tar xf geckodriver-v0.30.0-linux64.tar.gz -C /usr/local/bin
-              if: matrix.browser == 'Firefox'
+              run: pip install -e .[dev,smiles]
 
             - name: Run pytest
               run: pytest -v --driver ${{ matrix.browser }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,9 @@ jobs:
             - name: Checkout repository
               uses: actions/checkout@v2
 
+            - name: Change folder ownder to 1000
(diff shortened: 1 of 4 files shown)


yakutovicha:
Ok, so up to now it is possible to run tests by entering the awb folder and executing the following:

```bash
docker build -t aiidalab/full-stack-with-browsers:latest tests/

JUPYTER_TOKEN=$(openssl rand -hex 32)

DOCKER_ID=$(docker run -it -d --rm -v $(pwd):/home/jovyan/apps/aiidalab-widgets-base -e JUPYTER_TOKEN=${JUPYTER_TOKEN} aiidalab/full-stack-with-browsers:latest)

docker exec -it ${DOCKER_ID} bash -c "cd apps/aiidalab-widgets-base && pip install -e.[dev,smiles]"

docker exec -it ${DOCKER_ID} bash -c "cd apps/aiidalab-widgets-base && pytest --driver Chrome -s"

docker stop ${DOCKER_ID}
```

I haven't yet figured out how to convert this script to github action. But I am going to try that now.

danielhollas:
This is a great effort, but I am just wondering, are we sure we want to introduce this much additional complexity to the current test suite?

btw: When I was thinking about this problem for my own up, I could not wrap my head around how to get code coverage data from a jupyter python kernel. Have you been able to solve that issue?

yakutovicha:
> This is a great effort, but I am just wondering, are we sure we want to introduce this much additional complexity to the current test suite?

I am reducing complexity actually. Everything happens inside the container.

> btw: When I was thinking about this problem for my own app, I could not wrap my head around how to get code coverage data from a jupyter python kernel. Have you been able to solve that issue?

It looks like it is not possible when using selenium. But nbmake does help here, see my PR #441. It executes the notebook cells as a normal python interpreter.

yakutovicha:
But, I have to admit, I struggle with the GitHub actions file. It is simply not possible to easily convert the script above into the GitHub actions and let it run. One gets errors that one even doesn't expect.

danielhollas:
> I am reducing complexity actually. Everything happens inside the container.

I was referring to the fact of having to build extra Docker image and extra Gihub actions. :-) Also, even if this work, without Selenium interaction the code coverage will not be complete anyway right? Since just executing the notebook is not enough, some code paths get executed through the Selenium interactions.

An alternative would be to focus on a more unit-style testing without the notebooks and measure that. Of course that has issues as well. :shrug:

yakutovicha:
Let me explain my intention, because I sort of fail to see why it is more complex.

**Here is what we have now:**
1. We start a docker container by `pytest-selenium`. That makes it impossible to extract the coverage since the pytest is located outside of the container and the jupyter container runs inside of it.
2. While trying to access jupyter notebooks we install Chrome/Firefox drivers. This is unavoidable.


**So here are the options:**
1. We do everything on the GitHub runner machine. But this requires configuring AiiDA, PGSQL, Rabbitmq and all the rest. I don't think it is easy. AiiDAlab image has it all.
2. We do everything inside the container, but that requires installing Chrome/Firefox drivers to it.
 
**Here is what I am trying to do**
1. I moved driver installation from GitHub actions to Dockerfile.
2. I ran the container and execute `pytest` in it. Pytest will execute all the `test_` files and all the notebooks. The only test we do not get coverage from is the selenium one. But that is fine, I believe. We will mimic those by running unit tests (as you mentioned).


So my conclusion. It is fine to keep the selenium tests. But we use them more as smoke tests. The same notebooks run as smoke tests can be run with `nbmake` which will report coverage of launching the widgets. The test coverage interacting with widgets will have to be approximated by unit-test style tests.

It is not clear to me what would be an alternative. But if you have an idea - please share.

danielhollas:
> We do everything inside the container, but that requires installing Chrome/Firefox drivers to it.

This seems like a false dichotomy. I would just separate different style of tests into separate Github jobs. 
If we're not able to get the coverage from selenium tests anyway, I'd keep them exactly as they are right now. No need for extra docker image. Unit tests or nbmake-style test can run in a separate Github workflow. Or am I missing something?

alex-treebeard:
hey sorry for intruding. nbmake maintainer here.

RE the [container permissions issue](https://github.com/aiidalab/aiidalab-widgets-base/pull/444#issuecomment-1462157279), I've had a lot more success using the [Devcontainer CLI](https://github.com/devcontainers/cli) in github actions.

RE [coverage](https://github.com/aiidalab/aiidalab-widgets-base/pull/444#issuecomment-1462237025): I am not aware of anybody producing coverage data via nbmake. My initial belief is you won't see any coverage because under the hood we launch a kernel and message it via the local network. Keen to know if I'm wrong!

RE testing widgets: What you have put in place for this project is very impressive, and I am aware that others are trying to do the same. Do you recommend this pattern of using Selenium? Could we improve it together? ([Issue](https://github.com/treebeardtech/nbmake/issues/99))

unkcpz:
@alex-treebeard Thanks a lot for chiming in and for the very useful advice!

> RE the https://github.com/aiidalab/aiidalab-widgets-base/pull/444#issuecomment-1462157279, I've had a lot more success using the [Devcontainer CLI](https://github.com/devcontainers/cli) in github actions.

I think it deserve to try and it will allow us to drop `pytest-docker` by using this solution. I didn't investigate how the built image can be further used in github CI. Just want to mention that the aiida-core start using this feature in https://github.com/aiidateam/aiida-core/pull/5913, and we can give it a try here as well.

> If we're not able to get the coverage from selenium tests anyway, I'd keep them exactly as they are right now. No need for extra docker image. Unit tests or nbmake-style test can run in a separate Github workflow. Or am I missing something?

For the PR itself, I tend to agree with @danielhollas here that for the test coverage, it makes more sense to focus on unit tests. I make a first attempt in https://github.com/aiidalab/aiidalab-widgets-base/pull/435 where I separate the `tests_notebook` out and implemented a unit test for the `ProcessInputWidget`. It seems to work fine. For this repository specifically which provide the basic widgets as building blocks of other AiiDAlab app, I think it is more proper to have unit type test rather than the smock tests. The ultimate goal is we reach a high test coverage without using the smoke test. The smoke tests are leaving to such as QeApp to having a black box test for integrity, and that is where the `devcontainers` can still be used as mentioned above.

danielhollas:
Hi @alex-treebeard  :wave: 

thanks for insightful comments!

> I am not aware of anybody producing coverage data via nbmake. My initial belief is you won't see any coverage because under the hood we launch a kernel and message it via the local network. Keen to know if I'm wrong!

Yeah, that's what I was afraid of. I did a bit of googling, and found another similar project, [nbval](https://github.com/computationalmodelling/nbval), that claims to be able to generate coverage for code imported to the notebook via [integration with pytest-cov](https://github.com/computationalmodelling/nbval#coverage). Do you think that is something nbmake could support in the future?

> RE testing widgets: What you have put in place for this project is very impressive, and I am aware that others are trying to do the same. Do you recommend this pattern of using Selenium? Could we improve it together? (https://github.com/treebeardtech/nbmake/issues/99)

I don't want to speak for others, but it is a new technology for me. It is great when it works and it already caught some bugs for me, but it is a bit hard to get right. It is very easy to create a flaky test if one is not careful, and the Python API documentation is not the best. It would be great to see another examples where people are doing this. 

We also might be unlucky in that one of the widgets here is doing 3D rendering, which brings its own challenges.

You can find some more tests in my repo:
https://github.com/ispg-group/aiidalab-ispg/tree/main/tests

yakutovicha:
Thanks, everybody, for the insightful comments. Indeed, I was wrong regarding the test coverage produced by `nbmake` - it does not do it, it seems.

At the moment, I can't continue with this idea due to time constraints. Even if we decide to move back to it at some point, this would require careful thinking.

So I am closing the PR.

PR #446, opened 2023-03-14
fix bug report to use the new return type of find_installed_packages

### LINKED ISSUE #445: Recent aiidalab changes broke the github error handler
Recent changes to `aiidalab` package modified the `find_installed_packages` functions. Turns out this function is used in the github error handler in AWB.

https://github.com/aiidalab/aiidalab-widgets-base/blob/7a0e266a1576d6a10423931a6821d3f510f4b000/aiidalab_widgets_base/bug_report.py#L22

The error handler now (ironically) excepts with the following;

```python
Error while generating bug report: 'str' object has no attribute 'name'
```

unkcpz:
fixes #445 

Note: after this PR is merged, it needs to backport to `1.4.x` as well.

### DIFF
--- a/tests/test_bug_report.py
+++ b/tests/test_bug_report.py
@@ -2,12 +2,6 @@ import base64
 import json
 import zlib
 
-# TODO: this is a workaround for a import issue of aiida-core fixtures..
-# from aiidalab_widgets_base.bug_report import (
-#     get_environment_fingerprint,
-#     parse_environment_fingerprint,
-# )
-
 
 def test_fingerprint_parser():
     """Test get_environment_fingerprint function and parse it out."""
--- a/tests/test_bug_report.py
+++ b/tests/test_bug_report.py
@@ -2,14 +2,49 @@ import base64
 import json
 import zlib
 
-from aiidalab_widgets_base.bug_report import (
-    get_environment_fingerprint,
-    parse_environment_fingerprint,
-)
+# TODO: this is a workaround for a import issue of aiida-core fixtures..
+# from aiidalab_widgets_base.bug_report import (
+#     get_environment_fingerprint,
+#     parse_environment_fingerprint,
+# )
 
 
 def test_fingerprint_parser():
     """Test get_environment_fingerprint function and parse it out."""
+    from aiidalab_widgets_base.bug_report import (
+        get_environment_fingerprint,
+        parse_environment_fingerprint,
+    )
+
+    encoding = "utf-8"
+    fingerprint = get_environment_fingerprint(encoding)
+
+    # Parse the fingerprint.
+    data = parse_environment_fingerprint(fingerprint)
+
+    # To test, manually generate the fingerprint and compare it to the output of the parser.
+    json_data = json.dumps(data, separators=(",", ":"))
+    got = base64.urlsafe_b64encode(zlib.compress(json_data.encode(encoding), level=9))
+
+    assert got == fingerprint
+
+
+def test_fingerprint_parser_for_old_aiidalab(monkeypatch):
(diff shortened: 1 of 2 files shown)


danielhollas:
@unkcpz just a quick question, is your fix compatible with both the old and new `aiidalab` package versions? I think it needs to be, otherwise it's a mess for the users.

unkcpz:
> @unkcpz just a quick question, is your fix compatible with both the old and new aiidalab package versions?

Nope, but I see your point. I'll make a change. Thanks for the head up.

unkcpz:
It support the old version of `aiidalab`. But the `test_fingerprint_parser` pass without hitting the correct code path. In the test, it still using the old aiidalab since we didn't release a new version yet.

This is a issue with unit test because of the `load_profile()` we used in `__init__.py`, I don't have very clear clue why that happened but that need the fix from aiida-core I assume. The workaround is moving the import clause into the test functions. @yakutovicha encounter the same issue in his PR, we need investigate more and then open the issue in aiida-core.

For this PR, I think it is ready to review.

unkcpz:
I also move the fixtures in this PR with are duplicate as @yakutovicha did in https://github.com/aiidalab/aiidalab-widgets-base/pull/441/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128
Let's discuss how we manage these on Thursday.

yakutovicha:
> I also move the fixtures in this PR with are duplicate as @yakutovicha did in https://github.com/aiidalab/aiidalab-widgets-base/pull/441/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128
> Let's discuss how we manage these on Thursday.

I wouldn't worry too much about this. If your PR is merged first, I will adapt mine. If mine - then the other way around. It is really about moving some chunk of code, nothing more than this.

tests/test_bug_report.py
```diff
@@ -0,0 +1,58 @@
+import base64
+import json
+import zlib
+
+# TODO: this is a workaround for a import issue of aiida-core fixtures..
+# from aiidalab_widgets_base.bug_report import (
+#     get_environment_fingerprint,
+#     parse_environment_fingerprint,
+# )
```
yakutovicha, line 9: Is this comment still needed?

yakutovicha (commented):
This looks good to me. I let @danielhollas have a final look since you guys were discussing this before. If everything is OK - let's merge it.

danielhollas (changes requested):
I've thought about this a bit more, and I really don't like that we have this random coupling between AWB and the `aiidalab` package, especially since we're importing a function from utils, i.e. not really something one might consider stable public API. 
I think just the very existence of this PR makes this argument self-evident.

I am aware that this might not by _DRY_ or whatever, but let's be pragmatic here and define our own simplified local `find_installed_packages` (without using the Package class). WDYT?

tests/test_bug_report.py
unkcpz, line 9:
```suggestion
```

unkcpz: Yep, removed.

unkcpz:
> I am aware that this might not by DRY or whatever, but let's be pragmatic here and define our own simplified local find_installed_packages (without using the Package class). WDYT?

I was worried about it not DRY, but after trying it is very clean. After all, we just need a very simple help function.

danielhollas (approved):
Thanks @unkcpz. I tested manually that this work with the latest `aiidalab` package.

Will you take care of cherry-picking this on the 1.x branch and releasing a new version?

unkcpz:
> Will you take care of cherry-picking this on the 1.x branch and releasing a new version?

Sure I'll do that now.

PR #455, opened 2023-03-17
Test and update process module

### LINKED ISSUE #454: Test process module

yakutovicha:
fixes #454

### DIFF
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -1,7 +1,32 @@
-from aiida import orm
+import pytest
+from aiida import engine, orm
 
 
-def test_process_inputs(generate_calc_job_node):
+@pytest.mark.usefixtures("aiida_profile_clean")
+def test_submit_button_widget(multiply_add_process_builder_ready):
+    """Test SubmitButtonWidget with a simple `WorkChainNode`"""
+    from aiida.workflows.arithmetic.multiply_add import MultiplyAddWorkChain
+
+    import aiidalab_widgets_base as awb
+
+    def hook(_=None):
+        pass
+
+    def return_inputs():
+        return multiply_add_process_builder_ready
+
+    widget = awb.SubmitButtonWidget(
+        process_class=MultiplyAddWorkChain, inputs_generator=return_inputs
+    )
+
+    widget.on_submitted(hook)
+
+    # Simulate the click on the button.
+    widget.on_btn_submit_press()
+
+
+@pytest.mark.usefixtures("aiida_profile_clean")
+def test_process_inputs_widget(generate_calc_job_node):
     """Test ProcessInputWidget with a simple `CalcJobNode`"""
     from aiidalab_widgets_base.process import ProcessInputsWidget
 
@@ -14,8 +39,8 @@ def test_process_inputs(generate_calc_job_node):
         }
     )
 
-    # test the widget can be instantiated with empty inputs
-    process_input_widget = ProcessInputsWidget(process=None)
+    # Test the widget can be instantiated with empty inputs
+    process_input_widget = ProcessInputsWidget()
 
     process_input_widget = ProcessInputsWidget(process=process)
     input_dropdown = process_input_widget._inputs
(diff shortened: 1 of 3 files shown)


tests/test_process.py
```diff
@@ -1,7 +1,32 @@
...
+        return multiply_add_process_builder_ready
+
+    widget = awb.SubmitButtonWidget(
+        process_class=MultiplyAddWorkChain, inputs_generator=return_inputs
+    )
+
+    widget.on_submitted(hook)
+
+    # Simulate the click on the button.
+    widget.on_btn_submit_press()
```
unkcpz, line 25: I think this will trigger the submit and there will be node in the DB.

tests/test_process.py
```diff
@@ -30,3 +55,141 @@ def test_process_inputs(generate_calc_job_node):
...
+def test_process_report_widget(multiply_add_completed_workchain):
+    """Test ProcessReportWidget with a simple `WorkChainNode`"""
+    from aiidalab_widgets_base.process import ProcessReportWidget
+
+    # Test the widget can be instantiated with empty inputs
+    widget = ProcessReportWidget()
+
+    # Test the widget can be instantiated with a process
+    widget = ProcessReportWidget(process=multiply_add_completed_workchain)
+    widget.update()
```
unkcpz, line 104: Is that possible to test some tangible thing happened here? The same for the tests below, I think just running `update` is sort of meaningless.

unkcpz (changes requested):
Thanks @yakutovicha. The tests for process widget seems not actually test the interaction with widget. I think when those widget gets process traitlets updated, must come with some text of the sub-widget updated.

danielhollas:
Just a quick general comment, could you devise a test that would uncover the issue in #433. You can then mark it with `xfail` and we fix it in another PR, but I think it's important to think it here to uncover other potential problems.

unkcpz:
@danielhollas thanks for the head up. I open an issue https://github.com/aiidalab/aiidalab-widgets-base/issues/456
It can be tested and fixed here.

tests/test_process.py
yakutovicha: Right, thanks. I fixed that in [7423f60](https://github.com/aiidalab/aiidalab-widgets-base/pull/455/commits/7423f60bf1e436c6a34d75263cfe31f3c9d4d113) 👍

tests/test_process.py
yakutovicha: That is correct. Is it a problem that should be fixed?

yakutovicha:
> Just a quick general comment, could you devise a test that would uncover the issue in #433. You can then mark it with `xfail` and we fix it in another PR, but I think it's important to think about it here to uncover other potential problems.

> @danielhollas thanks for the head up. I open an issue #456 It can be tested and fixed here.

Thanks, guys. Indeed, I make those series of PRs as sort of a side product of manual testing. The goal is to make sure that at least the basics are tested.

For more in-depth stuff I would divert to separate issues/PRs so that I can focus more on the specific widget that requires special attention. Just to avoid delays with a release.

yakutovicha:
> Thanks @yakutovicha. The tests for process widget seems not actually test the interaction with widget. I think when those widget gets process traitlets updated, must come with some text of the sub-widget updated.

In the places where it makes more sense, I updated some of the widgets to follow the process status "live".

tests/test_process.py
unkcpz: I don't think there is any problem here, just think we can add an assert to test the button is clicked and the process is triggered.

tests/test_process.py
unkcpz: Sorry, you misunderstand me here. I was trying to ask you to add an assert clause to test the widget is changed after the `update` is called. The widget value should be update in terms of the process information.

For instance, the `ProcessReportWidget` here, after call `widget.update()`, the `self.value` is updated and it will contain some information about the input process.

unkcpz (changes requested):
thanks @yakutovicha! Just minor request to add some assert to the widget value, it is helpful to guarantee the `update` methods of widgets is run as expected.

tests/test_process.py
yakutovicha: done.

tests/test_process.py
yakutovicha: Ok, done. I changed the other widgets, however. The process doesn't produce a report here, so the message stays the same.

unkcpz (approved):
Thanks, @yakutovicha, looks all good to me.

PR #458, opened 2023-03-20
Test and update databases module

### LINKED ISSUE #457: Test databases module

yakutovicha:
fixes #457

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,7 +20,7 @@ classifiers =
 packages = find:
 install_requires =
     PyCifRW~=4.4
-    aiida-core[atomic_tools]>=2.1,<3
+    aiida-core>=2.1,<3
     aiidalab>=21.11.2
     aiidalab-eln>=0.1.2,~=0.1
     ansi2html~=1.6
@@ -30,6 +30,7 @@ install_requires =
     ipytree~=0.2
     ipywidgets~=7.7
     more-itertools~=8.0
+    pymysql~=0.10
     nglview~=3.0
     optimade-client==2022.9.19
     spglib~=1.14
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,7 +20,7 @@ classifiers =
 packages = find:
 install_requires =
     PyCifRW~=4.4
-    aiida-core[atomic-tools]>=2.1,<3
+    aiida-core[atomic_tools]>=2.1,<3
     aiidalab>=21.11.2
     aiidalab-eln>=0.1.2,~=0.1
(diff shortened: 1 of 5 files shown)


setup.cfg
```diff
@@ -20,7 +20,7 @@ classifiers =
 packages = find:
 install_requires =
     PyCifRW~=4.4
-    aiida-core>=2.1,<3
+    aiida-core[atomic-tools]>=2.1,<3
```
yakutovicha, line 23: Despite I originally wrote `atomic_tools`, the `setup-cfg-fmt` hook had a different opinion on what should it be. Let's see if it still works.

yakutovicha: I just removed this annoying hook.

setup.cfg
```diff
@@ -20,7 +20,7 @@ classifiers =
 packages = find:
 install_requires =
     PyCifRW~=4.4
-    aiida-core>=2.1,<3
+    aiida-core[atomic_tools]>=2.1,<3
```
danielhollas, line 23: Please let's not do this. This is a big dependency, and pymatgen was giving us troubles before. If this is because of ase, please add ASE separately.

setup.cfg
unkcpz: okay, let's remove it, I encountered it once in qeapp. But why you add these extra dependencies? Do we really need everything here? I think the `PyCifRW` is included in this extra, so if you add this, you need remove `PyCifRW` above. I personally disagree with adding these extra dependencies which include `pymatgen` which bring a lot of dependencies issue for other apps.

unkcpz (approved):
Thanks @yakutovicha, all good from my side.

tests/test_databases.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # Initiate the widget with no arguments.
+    widget = ComputationalResourcesDatabaseWidget()
+    assert "merlin.psi.ch" in widget.database
+
+    # Initialize the widget with default_calc_job_plugin="cp2k"
+    widget = ComputationalResourcesDatabaseWidget(default_calc_job_plugin="cp2k")
+    assert (
+        "merlin.psi.ch" not in widget.database
+    )  # Merlin does not have CP2K installed.
```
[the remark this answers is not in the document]
danielhollas: You could maybe instead select a non-existing plugin instead of cp2k and assert that widget.database is empty (this happens for example for `orca` currently). And for cp2k just assert that it is non-empty.

danielhollas (approved):
lgtm after a very cursory review.

PR #466, opened 2023-03-22
Test and update the `wizard` module

### LINKED ISSUE #343: WizardAppWidgetStep excepts under some conditions
My app excepts under certain conditions. I haven't yet determine the exact conditions, but here's part of the stack trace:

```python
~/apps/aiidalab-ispg/aiidalab_ispg/steps.py in _update_state(self)
    644     def _update_state(self):
    645         if self.process is None:
--> 646             self.state = self.State.INIT
    647             return
    648 
.
.
.
~/apps/aiidalab-widgets-base/aiidalab_widgets_base/wizard.py in _update_step_state(self, _)
    188             self._update_titles()
    189             self._update_buttons()
--> 190             self._consider_auto_advance()
    191 
    192     @traitlets.observe("selected_index")

~/apps/aiidalab-widgets-base/aiidalab_widgets_base/wizard.py in _consider_auto_advance(self, _)
    175         with self.hold_trait_notifications():
    176             index = self.accordion.selected_index
--> 177             last_step_selected = index + 1 == len(self.accordion.children)
    178             selected_widget = self.accordion.children[index]
    179             if (

TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```

I'll try to determine the exact conditions so we can reproduce this, but from the exception the fix seems straightforward.

### LINKED ISSUE #461: Test the `wizard` module

yakutovicha:
fixes #461
fixes #343

### DIFF
--- a/aiidalab_widgets_base/wizard.py
+++ b/aiidalab_widgets_base/wizard.py
@@ -4,12 +4,12 @@ Authors:
 
     * Carl Simon Adorf <simon.adorf@epfl.ch>
 """
-from enum import Enum
-from threading import Thread
-from time import sleep, time
+import enum
+import threading
+import time
 
 import ipywidgets as ipw
-import traitlets
+import traitlets as tl
 
 
 class AtLeastTwoStepsWizardError(ValueError):
@@ -21,10 +21,10 @@ class AtLeastTwoStepsWizardError(ValueError):
         )
 
 
-class WizardAppWidgetStep(traitlets.HasTraits):
+class WizardAppWidgetStep(tl.HasTraits):
     "One step of a WizardAppWidget."
 
-    class State(Enum):
+    class State(enum.Enum):
         """Each step is always in one specific state.
 
         The state is used to determine:
@@ -68,8 +68,8 @@ class WizardAppWidgetStep(traitlets.HasTraits):
         # All error states have negative codes
         FAIL = -1  # the step has unrecoverably failed
 
-    state = traitlets.UseEnum(State)
-    auto_advance = traitlets.Bool()
+    state = tl.UseEnum(State)
+    auto_advance = tl.Bool()
 
     def can_reset(self):
         return hasattr(self, "reset")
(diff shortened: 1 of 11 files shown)


aiidalab_widgets_base/wizard.py
```diff
@@ -118,10 +118,11 @@ def __init__(self, steps, **kwargs):
 
         def spinner_thread():
             while True:
-                sleep(0.1)
+                time.sleep(0.1)
                 self._update_titles()
 
-        Thread(target=spinner_thread).start()
+        if not testing:  # pytest is hanging if we start a thread
+            threading.Thread(target=spinner_thread).start()
```
unkcpz, line 125:
```suggestion
        self.spinner_thread = threading.Thread(target=spinner_thread).start()
```

The thread can be stop and joined in the test explicitly. It is also good to leave a handler attribute to get control of inner thread.
You

unkcpz (changes requested):
@yakutovicha thanks! only a minor request.
Is that possible to address https://github.com/aiidalab/aiidalab-widgets-base/issues/343 here?

tests/test_wizard.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Check state after finishing the first step.
+    assert s1.state == s1.State.SUCCESS
+    assert s2.state == s2.State.READY
+    assert widget.accordion.selected_index == 1
+
+    # Check state after resetting the app.
+    widget.reset()
+    assert s1.state == s1.State.INIT
+    assert s2.state == s2.State.INIT
```
unkcpz, line 76:
```suggestion
    assert s2.state == s2.State.INIT
    assert widget.back_button.disabled is True
```

aiidalab_widgets_base/wizard.py
unkcpz: Did you try `self.spinner_thread.join()` this will wait for the thread to stop and finish. We did it here https://github.com/aiidalab/aiidalab/blob/50498b08ea77937cb8af0e8d753446dd17b6d637/tests/test_appclass.py#L87-L104

yakutovicha: > Did you try `self.spinner_thread.join()` this will wait for the thread to stop and finish. We did it here https://github.com/aiidalab/aiidalab/blob/50498b08ea77937cb8af0e8d753446dd17b6d637/tests/test_appclass.py#L87-L104 we can try together.

The problem is that the current loop is infinite:

```python
while True:
    ....
```

So there is no reason for it to stop. And calling `_stop()` didn't stop it.

unkcpz (approved):
Thanks Looks all good to me.

PR #471, opened 2023-03-24
Restore backwards compatibility for some widgets

### LINKED ISSUE #426: Make some changes in AWB 2.0 backwards compatible

yakutovicha:
fixes #426

### DIFF
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -265,3 +265,4 @@ def test_process_nodes_tree_widget(multiply_add_completed_workchain):
     from aiidalab_widgets_base.process import ProcessNodesTreeWidget
 
     ProcessNodesTreeWidget(value=multiply_add_completed_workchain.uuid)
+    ProcessNodesTreeWidget(process=multiply_add_completed_workchain)
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -222,6 +222,43 @@ def test_process_list_widget(multiply_add_completed_workchain):
 
 
 @pytest.mark.usefixtures("aiida_profile_clean")
+def test_process_monitor(
+    multiply_add_process_builder_ready, daemon_client, await_for_process_completeness
+):
+    """Test ProcessMonitor with a simple `WorkChainNode`"""
+    from aiidalab_widgets_base.process import ProcessMonitor
+
+    ProcessMonitor()
+
+    # Stopping the daemon and submitting the process.
+    daemon_client.stop_daemon(wait=True)
+    process = engine.submit(multiply_add_process_builder_ready)
+
+    test_variable = False
+
+    def f():
+        nonlocal test_variable
+        test_variable = True
+
+    widget = ProcessMonitor(value=process.uuid, callbacks=[f])
+
+    # Starting the daemon and waiting for the process to complete.
+    daemon_client.start_daemon()
+    await_for_process_completeness(process)
+
+    widget.join()  # Make sure the thread is finished.
+
+    assert test_variable
+
+    # Now do the same with the process trait instead of value.
+    daemon_client.stop_daemon(wait=True)
+    process = engine.submit(multiply_add_process_builder_ready)
+    ProcessMonitor(process=process)
+    daemon_client.start_daemon()
+    await_for_process_completeness(process)
+
+
+@pytest.mark.usefixtures("aiida_profile_clean")
 def test_process_nodes_tree_widget(multiply_add_completed_workchain):
     """Test ProcessNodesTreeWidget with a simple `WorkChainNode`"""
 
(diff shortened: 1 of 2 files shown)


tests/test_process.py
```diff
@@ -221,10 +221,48 @@ def test_process_list_widget(multiply_add_completed_workchain):
...
+
+    assert test_variable
+
+    # Now do the same with the process trait instead of value.
+    daemon_client.stop_daemon(wait=True)
+    process = engine.submit(multiply_add_process_builder_ready)
+    ProcessMonitor(process=process)
+    daemon_client.start_daemon()
+    await_for_process_completeness(process)
+
```
unkcpz, line 259:
```suggestion
    daemon_client.stop_daemon(wait=True)
```

If use `daemon_client`, need to stop daemon explicitly after the test.

unkcpz (changes requested):
Thanks @yakutovicha for taking care of this. The changes are good. 
Although there is an issue caused by using threading with daemon as we discussed offline, I propose we ignore it for the moment.

```
03/25/2023 11:26:31 AM <2172657> sqlalchemy.pool.impl.QueuePool: [ERROR] Exception during reset or similar
  File "/opt/conda/lib/python3.9/site-packages/sqlalchemy/pool/base.py", line 753, in _finalize_fairy
    fairy._reset(pool)
  File "/opt/conda/lib/python3.9/site-packages/sqlalchemy/pool/base.py", line 1004, in _reset
    pool._dialect.do_rollback(self)
  File "/opt/conda/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 683, in do_rollback
    dbapi_connection.rollback()
psycopg2.OperationalError: server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
```

I spend some time try to figure it out and narrow down it with a minimal reproduce case. It can be reproduced when submit the process using test daemon, and having a separate threading query the database. 

The test can be like this:

```python
def test_process_monitor(
    multiply_add_process_builder_ready, daemon_client, await_for_process_completeness
):
    """Test ProcessMonitor with a simple `WorkChainNode`"""
    from aiidalab_widgets_base.process import ProcessMonitor

    process = engine.submit(multiply_add_process_builder_ready)
    widget = ProcessMonitor(value=process.uuid)
```

The `_monitor_process` method of `ProcessMonitor` just need to have a query as:

```python
    def _monitor_process(self, process_uuid):
        assert process_uuid is not None
        process = orm.load_node(process_uuid)
```

I assume this still comes from aiida in general not thread-safe, I think maybe we need help from Chris to dig the root out. I'll ask him and set a meeting for that, but this has to be after the 2.0 release of AWB.

danielhollas:
@yakutovicha I will take a closer look at this, please wait before merging, thanks.

danielhollas:
>  Although there is an issue caused by using threading with daemon as we discussed offline, I propose we ignore it for the moment.

@unkcpz can you elaborate on this? Is this problem present also on master or only on this PR? If the latter, we should not merge this.
What aiida-core version have you seen it? I think there have been some issues with the daemon for version < 2.2.2

danielhollas (commented):
tbh I am not very happy about the code here (and I realize I was the one pushing for this change, my apologies). The circular nature of the `value` and `process` traits makes the code very confusing and error prone imo. My original idea was to have a one-way flow from `process` to `value`, but that is not great either since they would not be in sync which might lead to bugs as well.

danielhollas:
@yakutovicha that is fine with me, since you anyway said we have currently control over all apps in the appstore.

Btw: the new test in this PR might be worthwhile merging.

I am travelling today back from Switzerland, should be available more tomorrow.

unkcpz:
> @unkcpz can you elaborate on this? Is this problem present also on master or only on this PR? If the latter, we should not merge this.
What aiida-core version have you seen it? I think there have been some issues with the daemon for version < 2.2.2

The error is shown in  https://github.com/aiidalab/aiidalab-widgets-base/pull/471#pullrequestreview-1357821001 and can be reproduce with the latest docker stack which running with aiida-core 2.2.2. 

It is not this PR which introduces the issue, it was there already. The tests will not failed but the error messages popped up showing the DB session is not properly shut down after the query. @yakutovicha and I suspect this might be the issue that causes the memory leak of the container we run for the moment.

unkcpz:
Seems we don't want to continue on this route. 
So @yakutovicha can you update the PR to leave only the test as @danielhollas suggested?

yakutovicha:
> Seems we don't want to continue on this route. So @yakutovicha can you update the PR to leave only the test as @danielhollas suggested?

yep, I just opened PR #478 with the test only.

PR #475, opened 2023-03-31
New Feature : Editor Adding tags to structure

AndresOrtegaGuerrero:
This PR implements a new Editor AddingTagsEditor

The idea behind the editor is to enable the user to set different kinds of atoms in a structure
The widget uses the ase.Atoms tags. 

The editor allows defining more kinds in a structure
for example, a structure can have Fe, O, and Fe1 (selected atoms)

This editor (or the idea behind it) is to enable the user to define new kinds, these new kinds can later be set to a specific magnetization for new kinds in the StructureData (For example in the AiiDAlab QE app).

### DIFF
--- a/aiidalab_widgets_base/__init__.py
+++ b/aiidalab_widgets_base/__init__.py
@@ -25,6 +25,7 @@ from .process import (
     SubmitButtonWidget,
 )
 from .structures import (
+    AddingTagsEditor,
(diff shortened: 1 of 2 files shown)


AndresOrtegaGuerrero:
This editor might be relevant in QeApp

PR #478, opened 2023-04-26
Add test for the `ProcessMonitor` widget.

yakutovicha:


tests/test_process.py
```diff
@@ -221,6 +221,37 @@ def test_process_list_widget(multiply_add_completed_workchain):
...
+    widget = ProcessMonitor(value=process.uuid, callbacks=[f])
+
+    # Starting the daemon and waiting for the process to complete.
+    daemon_client.start_daemon()
+    await_for_process_completeness(process)
+
+    widget.join()  # Make sure the thread is finished.
+
+    assert test_variable
+    daemon_client.stop_daemon(wait=True)
```
unkcpz, line 252: Seems this will close the daemon for the test session and lead to the failure of other tests that use the daemon. Maybe for every tests that use the daemon using `started_daemon_client` fixture? As documented in https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/plugins.html#plugin-test-fixtures

yakutovicha: I need the `daemon_client` fixture as I stop and start it a few times during the test. I updated the PR.

unkcpz (approved):
Looks all good. Thanks!

tests/test_process.py
yakutovicha: `daemon_client` has it all, so I just used it. 🤷

PR #479, opened 2023-04-26
Remove spinning Thread from WizzardAppWidgetStep

danielhollas:
WizzardAppWidget uses an infinite background thread to continously update titles of the Accordion widget. It looks like the only reason for this is to implement a dynamic "spinner".
This thread just needlesly consumes server resources for a very little benefit. Implementing a "spinner" in this way is silly.

More importantly, it looks like this background thread is the main reason for the memory leak we're seeing, see https://github.com/aiidalab/issues/issues/13

Here I remove the thread in favour of a static unicode character, in line with all the other icons. It would be great if we could use some dynamic icons, e.g. fa-spinner as used in the SmilesWidget, but it looks like the Accordion.set_title() function does not parse HTML. Ultimately I think having static icon is fine, and users can implement something more dynamic in the child widgets.

TODO:
 - [x] Extend the existing unit test. Most importantly, we should test that the titles are updated when the `state` trait of the child widget changes.
 - [x] Is there a better Unicode character that we could use?

### DIFF
--- a/aiidalab_widgets_base/wizard.py
+++ b/aiidalab_widgets_base/wizard.py
@@ -188,11 +188,13 @@ class WizardAppWidget(ipw.VBox):
             self.accordion.children[idx] for idx in range(len(self.accordion.children))
         ]
 
-        if all(step.can_reset() for step in steps):
-            return True
+        if not all(step.can_reset() for step in steps):
+            return False
 
         if any(step.state is not WizardAppWidgetStep.State.INIT for step in steps):
             return True
+
+        return False
 
     def _update_buttons(self):
         with self.hold_trait_notifications():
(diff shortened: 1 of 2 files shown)


danielhollas:
@unkcpz @yakutovicha I will be on holiday till Monday. Feel free to pick up this PR.

yakutovicha:
> [ ] Is there a better Unicode character that we could use?

I think the Unicode character you picked up is good enough 👍

unkcpz:
Super nice! thanks a lot, @danielhollas. I'll do the review after @yakutovicha implementation of the unit test.

unkcpz:
Super nice! thanks a lot, @danielhollas. I'll do the review after @yakutovicha implementation of the unit test.

yakutovicha:
@unkcpz if you are ok with the current "busy" Unicode character, please tick the box. I am ok with it.

Otherwise, the PR is ready for your review.

danielhollas:
Thank you @yakutovicha @unkcpz.
Because the memory leak is such a critical issue, I would suggest to backport this to the 1.4 support branch.

unkcpz:
>  I would suggest to backport this to the 1.4 support branch.

I'll take care of this after this one gets merged.

PR #480, opened 2023-04-27
ProcessMonitor: Increase default timeout

danielhollas:
ProcessMonitor has a background thread that polls the process status. The default interval is now 0.1 seconds which is quite aggressive, and may potentially lead to memory leak as discussed on https://github.com/aiidalab/issues/issue/13

Here I propose to increase the default to 1.0s, since the typical AiiDA workflow will anyway take quite a bit longer. If that seems to large, perhaps 0.5s might be a good compromise? In any case, widgets can override this by passing the timeout parameter to the constructor.


yakutovicha (approved):
The 1.0-second timeout looks good to me. Moreover, users can change it if they want to.

@unkcpz would you agree?

unkcpz (approved):
Agree, 1s looks good. Merged.

PR #12, opened 2019-03-10
Computer and code setup

yakutovicha:
Here is my work on computer and code setup.

In brief what I've implemented:

1) In `CodeDropdown widget` I added a `refresh` button and `setup another` link.

2) `setup another` link will bring the user to code setup notebook, which is based on `AiiDACodeSetup` widget. The notebook also accepts initial values of the parameters through URL:

```
from aiidalab_widgets_base.codes import AiiDACodeSetup, extract_aiidacodesetup_arguments

parsed_url = urlparse.parse_qs(urlparse.urlsplit(jupyter_notebook_url).query)
args = extract_aiidacodesetup_arguments(parsed_url)

code = AiiDACodeSetup(**args)
display(code)
```

3)  `AiiDACodeSetup` widget also contains `ComputerDropdown` widget that allows to select computers that are already setup. 

4) The `ComputerDropdown` widget also contains a link to setup a new computer.

5) The computer setup notebook is based on  `SshComputerSetup` and `AiidaComputerSetup` widgets. They are used to configure password-free access to remote machine and to configure the computer with AiiDA. Both widgets accept initial values for the parameters (in same way as  `AiiDACodeSetup` does).

PR #481, opened 2023-05-15
Remove leftover load profile

unkcpz:
The `load_profile()` in `__init__.py` to load the profile in module scope was introduced by PR https://github.com/aiidalab/aiidalab-widgets-base/pull/12. It is not needed anymore since the computer and code widgets are updated and can update the dropdown dynamically. 

More problematically that this makes the profile loaded in the unit tests and overrides the test profile which makes it has to put the module load inside every test. There is also the problem that if the user want to test the widgets in a newly create python environment that doesn't have an AiiDA profile setup impossible. 

- [x] move modules import from test to file in front.

Before meeting this, we should make sure that apps which depend on AWB, load profiles in the notebooks. We checkbox the apps if they do:

- [x] [Aurora app](https://github.com/epfl-theos/aiidalab-aurora) (doesn't depend on AWB)
- [ ] [Molecules app](https://github.com/nanotech-empa/aiidalab-empa-molecules)
- [x] [LSMO app](https://github.com/lsmo-epfl/aiidalab-epfl-lsmo.git) (not migrated)
- [ ] [Martket place - use case 3](https://github.com/daniel-sintef/aiidalab-uc3)
- [x] [Graphene nanoribbons](https://github.com/nanotech-empa/aiidalab-empa-nanoribbons)
- [ ] [Quantum ESPRESSO app](https://github.com/aiidalab/aiidalab-qe)
- [x] [Scanning Probe Microscopy](https://github.com/nanotech-empa/aiidalab-empa-scanning-probe) (deprecated)
- [ ] [SSSP toolbox](https://github.com/aiidalab/aiidalab-sssp)
- [x] [On-Surface Chemistry](https://github.com/nanotech-empa/aiidalab-empa-surfaces)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,11 +5,6 @@ name: continuous-integration
 
 on: [push, pull_request]
 
-# https://docs.github.com/en/actions/using-jobs/using-concurrency
-concurrency:
-    # only cancel in-progress jobs or runs for the current workflow - matches against branch & tags
-    group: ${{ github.workflow }}-${{ github.ref }}
-    cancel-in-progress: true
 
 jobs:
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,12 @@ name: continuous-integration
 
 on: [push, pull_request]
 
+# https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+    # only cancel in-progress jobs or runs for the current workflow - matches against branch & tags
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
+
 jobs:
 
     pre-commit:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,7 +126,8 @@ jobs:
               run: pip install -e .[dev,smiles]
 
             - name: Run pytest
-              run: pytest -v tests --cov
+              # use TESTING=true to disable the automatic load of the aiida profile
+              run: TESTING=true pytest -v tests --cov
               env:
                   TAG: ${{ matrix.tag }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,8 +126,7 @@ jobs:
               run: pip install -e .[dev,smiles]
 
             - name: Run pytest
-              # use TESTING=true to disable the automatic load of the aiida profile
-              run: TESTING=true pytest -v tests --cov
+              run: pytest -v tests --cov
               env:
                   TAG: ${{ matrix.tag }}
 
--- a/aiidalab_widgets_base/__init__.py
+++ b/aiidalab_widgets_base/__init__.py
@@ -15,13 +15,13 @@ load_profile();</pre>
 if get_profile() is None:
     # if no profile is loaded, load the default profile and raise a deprecation warning
     from aiida import load_profile
-    from IPython.display import display, HTML
-    
+    from IPython.display import HTML, display
+
     load_profile()
-    
+
     profile = get_profile()
     assert profile is not None, "Failed to load the default profile"
-    
+
     # raise a deprecation warning
     warning = HTML(_WARNING_TEMPLATE.format(profile=profile.name, version="v3.0.0"))
     display(warning)
--- a/aiidalab_widgets_base/__init__.py
+++ b/aiidalab_widgets_base/__init__.py
@@ -1,6 +1,4 @@
 """Reusable widgets for AiiDAlab applications."""
-import os
-
 from aiida.manage import get_profile
 
 _WARNING_TEMPLATE = """
@@ -13,15 +11,25 @@ load_profile();</pre>
 </div>
 """
 
-# use the environment variable to control the automatic profile loading
-# so it can be disabled for testing, e.g. in the CI
-# use `TESTING=1 pytest` to disable the automatic profile loading
-TESTING = os.environ.get("TESTING", "False").lower() in ["true", "1"]
+
+# We only detect profile and throw a warning if it is on the notebook
+# It is not necessary to do this in the unit tests
+def is_running_in_jupyter():
+    try:
+        from IPython import get_ipython
+
+        if get_ipython() is not None:
+            return True
+        else:
+            return False
+    except NameError:
+        return False
+
 
 # load the default profile if no profile is loaded, and raise a deprecation warning
 # this is a temporary solution to avoid breaking existing notebooks
 # this will be removed in the next major release
-if not TESTING and get_profile() is None:
+if is_running_in_jupyter() and get_profile() is None:
     # if no profile is loaded, load the default profile and raise a deprecation warning
     from aiida import load_profile
     from IPython.display import HTML, display
--- a/aiidalab_widgets_base/__init__.py
+++ b/aiidalab_widgets_base/__init__.py
@@ -1,4 +1,6 @@
 """Reusable widgets for AiiDAlab applications."""
+import os
+
 from aiida.manage import get_profile
 
 _WARNING_TEMPLATE = """
@@ -11,8 +13,15 @@ load_profile();</pre>
 </div>
 """
 
+# use the environment variable to control the automatic profile loading
+# so it can be disabled for testing, e.g. in the CI
+# use `TESTING=1 pytest` to disable the automatic profile loading
+TESTING = os.environ.get("TESTING", "False").lower() in ["true", "1"]
 
-if get_profile() is None:
+# load the default profile if no profile is loaded, and raise a deprecation warning
+# this is a temporary solution to avoid breaking existing notebooks
+# this will be removed in the next major release
+if not TESTING and get_profile() is None:
     # if no profile is loaded, load the default profile and raise a deprecation warning
     from aiida import load_profile
     from IPython.display import HTML, display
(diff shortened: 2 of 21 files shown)


unkcpz:
well, there is one thing needs to change for the existing notebooks or apps the `load_profile` need to be explicitly called from app. Which also means the apps need to be adopted.
I think it is a more proper way to do it, as instructed in the aiida doc https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/installation.html?highlight=jupyter%20notebook#using-aiida-in-jupyter. 

pinning @yakutovicha @danielhollas for comments.

danielhollas (changes requested):
I like this change. In particular, this will allow in the future to support multiple AiiDA profiles. 
However, as you note this will break most of the existing apps so we should not take this lightly. It's a pity that we did not do it before the 2.0 release. Should we wait with this for 3.0?

On a practical side, I would propose to add a code in `aiidalab_widgets_base/__init__.py` that would check whether the AiiDA profile is loaded and throws a `RuntimeError` if that is not the case. This error should at least make it very clear to the developers what needs to be done.

unkcpz:
>  Should we wait with this for 3.0?

I don't think this will happen soon.  @yakutovicha we still have control of most of app, do you think it is doable to make this change? I'd argue this can be a feature that is worth adding with a minor release.

> I would propose to add a code in aiidalab_widgets_base/__init__.py that would check whether the AiiDA profile is loaded and throws a RuntimeError if that is not the case. 

I didn't see the point of doing this. This is to prevent the loaded profile being overrided by the `load_profile` of `aiidalab_widgets_base/__init__.py`? Changing/overriding the profile in a notebook is not a problem I guess?

danielhollas:
> I didn't see the point of doing this. This is to prevent the loaded profile being overrided by the load_profile of aiidalab_widgets_base/__init__.py?

The point is only to provide a clear error message in case the developer forgets to call `load_profile()` before loading AWB. I do not know what happens now, perhaps the existing error that one gets is clear enough? 

The question I had not  considered is whether there are widgets which do not require loaded AiiDA profile? In which case we should not have this runtime check I guess.

yakutovicha:
> I don't think this will happen soon. @yakutovicha we still have control of most of app, do you think it is doable to make this change? I'd argue this can be a feature that is worth adding with a minor release.

I don't think it is a big problem to do that. In most of our notebooks, we have `%aiida` magic which does that already.

At the moment, the only fear I have is that some of the modules of AWB that require AiiDA classes will fail to load. Maybe this is ok (in the end, for exactly that reason we launch `verdi shell` instead of just `ipython`).

yakutovicha:
So I would probably go with this PR taking the suggestion of @danielhollas to through `RuntimeError` in case the profile is not loaded.

unkcpz:
> load_profile() before loading AWB. I do not know what happens now, perhaps the existing error that one gets is clear enough?

As I did try in lastest two commits, if the check is put in the `__init__.py` it will run when the module is imported, and the exception will then raise if the profile is not loaded before `import aiidalab_widgets_base`. I think the current error is clear enough for AiiDA users. Therefore I rollback the change.

yakutovicha:
@unkcpz, is there anything left here to be done?

danielhollas:
Are we in a rush for this? This is a hard breaking change for essentially
all apps that use AWB.

Dne st 12. 7. 2023 9:13 uživatel Aliaksandr Yakutovich <
***@***.***> napsal:

unkcpz:
> Are we in a rush for this? This is a hard breaking change for essentially

Not in a rush and yes it is a change that will break other apps. So I tag it to milestone `2.1.0`. But it needs to happen sometime, so the question to me is what is the blocker for not making this move? At least we still have all control of all apps and we are in the stage of migrating things, once all app migration in EMPA is finished or your app is released, there will be more resistance to pushing breaking changes, I guess?

> @unkcpz, is there anything left here to be done?

Yep, It is ready. I'll address your requests.

unkcpz:
> We can, for example, list all the apps and tick-box them if they do.

This is a good idea, can you create a list of apps and tick boxes for EPMA apps? I'll check other apps.

yakutovicha:
> > We can, for example, list all the apps and tick-box them if they do.
> 
> This is a good idea, can you create a list of apps and tick boxes for EPMA apps? I'll check other apps.

Sure, I will put it in the top comment.

danielhollas:
> Not in a rush and yes it is a change that will break other apps. So I tag it to milestone 2.1.0. But it needs to happen sometime, so the question to me is what is the blocker for not making this move? At least we still have all control of all apps and we are in the stage of migrating things, once all app migration in EMPA is finished or your app is released, there will be more resistance to pushing breaking changes, I guess?

Note that I've already sent my app to some external collaborators so I'd really like things to be stable. At the very least, for breaking things like this I'd like us to get back into a habit of having a reasonable deprecation period.

The good news here is that we can first make all the existing apps ready for this change, by making sure they load the profile in the notebook (loading profile twice is okay). 

I would propose to first do:

1. In AWB `__init__`, check whether profile is loaded. If it is, do nothing.
2. If a profile is not loaded, load the default profile and emit a deprecation warning.

In this way, we can release this change in a fully backwards compatible manner (even in a patch release). At the same time it would already allow us to do all the things we wanted (e.g. load non-default profiles, easier unit testing).

WDYT @unkcpz @yakutovicha ?

yakutovicha:
> I would propose to first do:
> 
> 1. In AWB `__init__`, check whether profile is loaded. If it is, do nothing.
> 2. If a profile is not loaded, load the default profile and emit a deprecation warning.

@unkcpz I assume you take care of that, right?

unkcpz:
Yes, I'll take care of this one.

danielhollas:
@unkcpz thanks! Since this is going to be backward compatible for now, I've moved it to 2.0.2 release.

aiidalab_widgets_base/__init__.py
```diff
@@ -1,7 +1,39 @@
...
+    <pre style="background-color: #f0f0f0; padding: 10px; border: 1px solid #ccc; font-family: 'Courier New', monospace;">
 from aiida import load_profile
+load_profile();</pre>
+</div>
+"""
+
+# use the environment variable to control the automatic profile loading
+# so it can be disabled for testing, e.g. in the CI
+# use `TESTING=1 pytest` to disable the automatic profile loading
+TESTING = os.environ.get("TESTING", "False").lower() in ["true", "1"]
```
danielhollas, line 19: I am not a fan of this solution. Could we instead detect whether we run in the Jupyter notebook context (perhaps checking whether ipython is loaded or something like that?). That way we should be able to detect when we're unit testing and not load the profile.

danielhollas (changes requested):
The warning message is very nice! Since it is rather intrusive, we should at least bump the minor version when we release this.

aiidalab_widgets_base/__init__.py
unkcpz: I also adapt suggestions from @yakutovicha, the import is with module namespaces now.

danielhollas: Great! I like the solution with `is_running_in_jupyter`. :+1:

notebooks/aiida_datatypes_viewers.ipynb
```diff
@@ -7,6 +7,17 @@
     "# Displaying AiiDA data types"
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from aiida import load_profile\n",
```
danielhollas, line 16: Can you please revert this change, so that we test that the default profile is loaded in __init__ still?
Also, you could then extend the `test_aiida_datatypes_viewers` test in `tests_notebooks/test_notebooks.py` to check that the deprecation warning is displayed.

danielhollas (changes requested):
@unkcpz thanks, this looks great now. I've tested this locally. I only have one request for testing the new behaviour.

This should be released in 2.1 version. So before merging, we should release 2.0.2. There are still some tasks in that milestone. I've taken care of one in #502.

.github/workflows/ci.yml
```diff
@@ -5,6 +5,12 @@ name: continuous-integration
 
 on: [push, pull_request]
 
+# https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
```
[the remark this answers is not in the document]
unkcpz: Is not needed here, but it is useful CI workflow configuration. Every time the new commit of a PR is submitted, it will stop the previously running test, this what we usually want and this save the time.

notebooks/aiida_datatypes_viewers.ipynb
unkcpz: Good point, I'll try to add a test.

yakutovicha (commented):
LGTM,

The only thing I would consider is whether to propose to the developers an alternative solution using ipython magic:

```python
%load_ext aiida
%aiida
 ```

**Comment by unkcpz** [comment]
> The only thing I would consider is whether to propose to the developers an alternative solution using ipython magic:

```
%load_ext aiida
%aiida
```

Yes, we can, but I remember there are issues with this magic, although all are fixed (https://github.com/search?q=repo%3Aaiidateam%2Faiida-core+ipython+magic&type=issues) but it is less robust than the `load_profile` I'd say. So I would recommend using the load_profile.

yakutovicha (approved):
Great work, @unkcpz!

unkcpz:
Thanks for the review! This will go to `v2.1` so anyone who makes the release should remember to cherry-pick commits when doing the release.

Issue #486, opened 2023-06-21
Update spglib to >=2.0


AndresOrtegaGuerrero:
This version causes conflit with newer versions of pymatgen

PR #487, opened 2023-06-22
updating spglib to 2.0.2

AndresOrtegaGuerrero:
This PR address #486

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,11 +34,7 @@ install_requires =
     pymysql~=0.9
     nglview~=3.0
     optimade-client==2022.9.19
-<<<<<<< HEAD
     spglib>=1.14,<3
-=======
-    spglib~=2.0
->>>>>>> d73b3fa74cc3b999055b2b2ae3311b1aa1105a72
     vapory~=0.1.2
 python_requires = >=3.8
 include_package_data = True
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,7 @@ install_requires =
     pymysql~=0.9
     nglview~=3.0
     optimade-client==2022.9.19
-    spglib~=1.14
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -34,7 +34,7 @@ install_requires =
     pymysql~=0.9
     nglview~=3.0
     optimade-client==2022.9.19
-    spglib~=1.14
+    spglib~=2.0.2
```
yakutovicha, line 37: I would do `~2.0`. Versions such as `2.1`, `2.2`, and `2.100` should still be ok.

```suggestion
    spglib~=2.0
```

yakutovicha (changes requested):
suggestion

setup.cfg
danielhollas: Agree, I think in general we should not pin to specific versions, otherwise we'll be running into this issue over and over again.

danielhollas (commented):
Just asking the obvious question, since the spglib is used in the viewer, has this been tested?

yakutovicha (dismissed):
LGTM!

danielhollas:
Just a question, does this mean that when this is merged and released, that the new AWB version will be incompatible with old QeApp versions?

If that is the case, I would suggest to relax the constraint here and allow both old and new spglib versions here, since it does not seem to affect the AWB code anyway.

AndresOrtegaGuerrero:
@danielhollas I should do this by just `spglib >=1.14` ?

yakutovicha:
> Just a question, does this mean that when this is merged and released, that the new AWB version will be incompatible with old QeApp versions?

I am confused, why QE app? I checked - QE app doesn't depend on `spglib`.

yakutovicha (approved):

Issue #485, opened 2023-06-20
Logic in viewers doesnt allow periodicity different than True , True , True 


AndresOrtegaGuerrero:
The logic in https://github.com/aiidalab/aiidalab-widgets-base/blob/0d9fee60086da122490eadb28b45a856263454d5/aiidalab_widgets_base/viewers.py#L321 doesnt allow pcb different from True , True , True

PR #488, opened 2023-06-24
allow structures with different periodicity, add periodicity

AndresOrtegaGuerrero:
This PR address #485 
The user can see the cell parameters of  StructureData with pbc != True, True, True
The PR also add a HTML line to display the periodicity of the structure

I did three test, 

1. add a xyz file (molecule) it doensnt show any cell information
2. add a StructureData with pbc as `True,True,False` `True, False, False` obtaining "xy" and "x"
3.  using the examples where all should be "xyz"

I also included a test in test_viewers.

### DIFF
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -318,7 +318,7 @@ class _StructureDataBaseViewer(ipw.VBox):
     @tl.observe("cell")
     def _observe_cell(self, _=None):
         # only update cell info when it is a 3D structure.
-        if self.cell and all(self.structure.pbc):
+        if self.cell and any(self.structure.pbc):
             self.cell_a.value = "<i><b>a</b></i>: {:.4f} {:.4f} {:.4f}".format(
                 *self.cell.array[0]
             )
@@ -348,8 +348,18 @@ class _StructureDataBaseViewer(ipw.VBox):
                 spglib_structure, symprec=1e-5, angle_tolerance=1.0
             )
 
+            periodicity_map = {
+                (True, True, True): 'xyz',
+                (True, False, False): 'x',
+                (False, True, False): 'y',
+                (False, False, True): 'z',
+                (True, True, False): 'xy',
+                (True, False, True): 'xz',
+                (False, True, True): 'yz',
+            }
             self.cell_spacegroup.value = f"Spacegroup: {symmetry_dataset['international']} (No.{symmetry_dataset['number']})"
             self.cell_hall.value = f"Hall: {symmetry_dataset['hall']} (No.{symmetry_dataset['hall_number']})"
+            self.periodicity.value = f"Periodicity: {periodicity_map[tuple(self.structure.pbc)]}"
         else:
             self.cell_a.value = "<i><b>a</b></i>:"
             self.cell_b.value = "<i><b>b</b></i>:"
@@ -378,6 +388,7 @@ class _StructureDataBaseViewer(ipw.VBox):
 
         self.cell_spacegroup = ipw.HTML()
         self.cell_hall = ipw.HTML()
+        self.periodicity = ipw.HTML()
 
         self._observe_cell()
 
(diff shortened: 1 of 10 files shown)


danielhollas (changes requested):
The code looks okay to me, but it would be great if you could add a unit test that would check the value of `self.periodicity` for different structures.

I don't really deal with solids so also pinging @yakutovicha for review and testing.

aiidalab_widgets_base/viewers.py
```diff
@@ -348,8 +348,20 @@ def _observe_cell(self, _=None):
                 spglib_structure, symprec=1e-5, angle_tolerance=1.0
             )
 
+            periodicity_map = {
```
danielhollas, line 351: Nice! :+1:

AndresOrtegaGuerrero: @danielhollas  Thank you!,  I will include the  test

tests/test_structures.py
```diff
@@ -23,6 +23,9 @@ def test_structure_manager_widget(structure_data_object):
     assert structure_manager_widget.structure is not None
     assert isinstance(structure_manager_widget.structure, ase.Atoms)
 
+    # Text the periodicity of the structure
```
danielhollas, line 26:
```suggestion
    # Test the structure periodicity
```

danielhollas (dismissed):
Thanks @AndresOrtegaGuerrero for adding the test! This looks great to me, but I haven't tested it manually, leaving that to @yakutovicha, since I do not deal with solids. :-)

aiidalab_widgets_base/viewers.py
```diff
@@ -318,7 +318,7 @@ def change_camera(change):
     @tl.observe("cell")
     def _observe_cell(self, _=None):
         # only update cell info when it is a 3D structure.
-        if self.cell and all(self.structure.pbc):
+        if self.cell and any(self.structure.pbc):
```
unkcpz, line 321: The function `_observe_cell` is for the 3D structure with the cell, so from my understanding, even the structure that is non-periodic can be in the cell and need to go through this cell. 
In this case, an additional mapping is required from `False, False, False` to non-periodic maybe represented as `-`.

danielhollas: I don't understand, why would a non-periodic structure have a cell? As a workaround for gas phase molecules in plane wave codes?

unkcpz: Yep, for the plane-wave basis. I won't regard it as a workaround, it may be not as accurate as Gaussian basis but a quite standard and common calculation in "our" daily life 😄

danielhollas: Haha, sure, didn't mean to offend. :D

AndresOrtegaGuerrero: @unkcpz  So perhaps the method should be modified? one thing is to check if the structure has a cell (and display or not the cell parameters defined), and another one should be to inspect what periodicity has defined.

aiidalab_widgets_base/viewers.py
```diff
@@ -348,8 +348,21 @@ def _observe_cell(self, _=None):
...
+                (True, False, True): "xz",
+                (False, True, True): "yz",
+                (False, False, False): "-",
+            }
             self.cell_spacegroup.value = f"Spacegroup: {symmetry_dataset['international']} (No.{symmetry_dataset['number']})"
             self.cell_hall.value = f"Hall: {symmetry_dataset['hall']} (No.{symmetry_dataset['hall_number']})"
+            self.periodicity.value = (
+                f"Periodicity: {periodicity_map[tuple(self.structure.pbc)]}"
+            )
         else:
```
yakutovicha, line 366: Could you please reset the following widgets here as well:

```python
            self.cell_spacegroup.value = ""
            self.cell_hall.value = ""
            self.periodicity.value = ""
```

Otherwise, if no structure is selected, the values of the previous structure are still shown.

yakutovicha (changes requested):
Thanks, @AndresOrtegaGuerrero. One little request.

unkcpz (approved):
@AndresOrtegaGuerrero thanks! All looks good to me.

yakutovicha (approved):
Tested - all good. Thanks a lot, @AndresOrtegaGuerrero!

PR #493, opened 2023-06-27
Remove dynamic version attribute from setup.cfg

danielhollas:
Dynamic version attributes are not yet supported in the `aiidalab` package. 
Since AWB is currently an App, and not just a library, we need to use a static version.

Here's what you currently get for AWB 2.0.0
```console
$ aiidalab list
App name               Version              Path
---------------------  -------------------  ---------------------------------------
aiidalab-widgets-base  AppVersion.UNKNOWN•  /home/jovyan/apps/aiidalab-widgets-base
home                   v23.03.1             /home/jovyan/apps/home
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = aiidalab_widgets_base
-version = attr: aiidalab_widgets_base.__version__
+version = 2.0.0
 description = Reusable widgets for AiiDAlab applications.
 long_description = file: README.md
 long_description_content_type = text/markdown
@@ -84,3 +84,5 @@ push = True
 [bumpver:file_patterns]
 aiidalab_widgets_base/__init__.py =
     __version__ = "{pep440_version}"
+setup.cfg =
+    version = {pep440_version}


danielhollas:
@yakutovicha ah, interesting. My outputs is what I see is when I pull master for AWB development. In other words, when the app is checked out at a commit that is not tagged. 

So I don't know if this is worth fixing (but I guess it does not hurt either since we're using bumpver anyway). What do you think @yakutovicha

yakutovicha:
> @yakutovicha ah, interesting. My outputs is what I see is when I pull master for AWB development. In other words, when the app is checked out at a commit that is not tagged.

The output makes sense then

> So I don't know if this is worth fixing (but I guess it does not hurt either since we're using bumpver anyway). What do you think @yakutovicha

Honestly, I will probably forget about this and one day revert to what it was (if no one will stop me), as it is a bit shorter and somewhat more automatic. So I am slightly negative.

But if you want this change, go ahead - it won't hurt either.

danielhollas:
I'll close this, not worth the churn. Thanks!

PR #495, opened 2023-07-07
StructureViewer: Print 3 decimal places for distance and angles

danielhollas:
Print more significant digits when printing Distance or angles in the Selection tab.

obrazek


superstar54 (approved):
LGTM!

PR #502, opened 2023-08-16
Remove Transport dropdown in computer setup, assume SSH

### LINKED ISSUE #417: Cannot setup localhost computer
Just tried to setup a localhost computer, with transport `core.local` and scheduler `core.direct`, but I got:

image

danielhollas:
It is not clear whether AiiDA Computer setup ever worked with any transport other than SSH,
but in any case the widget is clearly designed with SSH in mind so unless a clear need is found we can simply remove the 'Transport type" dropdown. Note the the value of this widget has not been used anywhere in the code, which is telling.

The localhost computer is setup in the AiiDAlab Docker image by default, and in the unlikely case that another one needs to be setup, such user probably knows how to set it up via verdi command line interface.

Fixes #417

### DIFF
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -733,9 +733,8 @@ class AiidaComputerSetup(ipw.VBox):
         )
 
         # Transport type.
-        self.transport = ipw.Dropdown(
-            value="core.local",
-            options=plugins.entry_point.get_entry_point_names("aiida.transports"),
+        self.transport = ipw.HTML(
+            value="SSH",
             description="Transport type:",
             style=STYLE,
         )
@@ -941,7 +940,6 @@ class AiidaComputerSetup(ipw.VBox):
         self.mpirun_command.value = "mpirun -n {tot_num_mpiprocs}"
         self.default_memory_per_machine_widget.value = ""
         self.mpiprocs_per_machine.value = 1
-        self.transport.value = "core.ssh"
         self.safe_interval.value = 30.0
         self.scheduler.value = "core.slurm"
         self.shebang.value = "#!/usr/bin/env bash"


yakutovicha (commented):
I think this is good for now.

However, later I am introducing the [`ssh_light`](https://github.com/aiidateam/aiida-core/pull/6083) which will revert things to `Dropdown`. Then, there will be `firecrest`. It would take some more work to adopt them, though.

Therefore, for now, it is good to keep it as `HTML` to signify that widget is not dealing with other cases.

I would only request to keep the `self.transport.value = "core.ssh"` line and set the original value to `core.ssh` too.

It was issued later on. Also, settings downloaded from the [code database](https://github.com/aiidateam/aiida-code-registry) override the value. Herein, `SSH` as the value may cause trouble.

danielhollas:
> However, later I am introducing the https://github.com/aiidateam/aiida-core/pull/6083 which will revert things to Dropdown. Then, there will be firecrest. It would take some more work to adopt them, though.

Ah, that's good to know. I also found out that the transport value is indeed passed to the computer setup function, I did not notice that before since it is accessed via `getattrs`

https://github.com/aiidalab/aiidalab-widgets-base/blob/413c26c76ec438b9c70afe3437420923f738e50b/aiidalab_widgets_base/computational_resources.py#L898

> I would only request to keep the self.transport.value = "core.ssh" line and set the original value to core.ssh too.
> It was issued later on. Also, settings downloaded from the [code database](https://github.com/aiidateam/aiida-code-registry) override the value. Herein, SSH as the value may cause trouble.

Another good point. As a side note, I am not a fan of this design, `core.ssh` is not very user friendly, we're essentially exposing the user to internal AiiDA details. But that's for another time...

Given all of the above, I've reverted the original change and instead fix the underlying bug and made the `core.local` transport work. I also added a unit test. @yakutovicha could you please test locally?

aiidalab_widgets_base/computational_resources.py
```diff
@@ -858,6 +858,17 @@ def _configure_computer(self, computer: orm.Computer):
...
         return True
 
+    def _configure_computer_local(self, computer: orm.Computer):
+        """Configure the computer with local transport"""
+        authparams = {
+            "use_login_shell": self.use_login_shell.value,
+            "safe_interval": self.safe_interval.value,
+        }
+        user = orm.User.collection.get_default()
+        computer.configure(user=user, **authparams)
```
danielhollas, line 868: Note that we need to catch any potential errors here. Since this was not done in the original function, I am keeping that for separate PR.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -902,7 +913,14 @@ def on_setup_computer(self, _=None):
...
         try:
             computer = computer_builder.new()
-            self._configure_computer(computer)
+            if self.transport.value == "core.ssh":
+                self._configure_computer_ssh(computer)
+            if self.transport.value == "core.local":
+                self._configure_computer_local(computer)
+            else:
+                self.message = (
+                    f"WARNING: Transport value {self.transport.value} not recognized!"
```
danielhollas, line 922: I think it would be safer to throw an error here. I guess a computer that is not configured is not usable?

yakutovicha: > I guess a computer that is not configured is not usable?

One can configure it later, of course. But that is not the purpose of the widget. The widget should create a computer that is ready to be used.

> I think it would be safer to throw an error here.

I agree. I made a suggestion below.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -902,7 +913,14 @@ def on_setup_computer(self, _=None):
...
+                )
```
yakutovicha, line 923:
```suggestion
                raise common.ValidationError(f"transport value {self.transport.value} not recognized!")
```

yakutovicha (commented):
Thanks, @danielhollas. As you've suggested, let's not allow creating a computer if its configuration is not working. The rest is good, I tested the widget locally and it works as expected 👍 .

aiidalab_widgets_base/computational_resources.py
danielhollas: Agreed, donw

aiidalab_widgets_base/computational_resources.py
```diff
@@ -902,16 +936,15 @@ def on_setup_computer(self, _=None):
         )
         try:
             computer = computer_builder.new()
-            self._configure_computer(computer)
+            self._configure_computer(computer, self.transport.value)
+            computer.store()
```
danielhollas, line 940: I've moved this here since it can also fail.

Strangely, when a user provides empty workdir, the store command raises ValidationError, but the computer is still stored. That to me seems like `aiida-core` bug.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -902,16 +936,15 @@ def on_setup_computer(self, _=None):
...
         except (
             ComputerBuilder.ComputerValidationError,
             common.exceptions.ValidationError,
             RuntimeError,
         ) as err:
-            self.message = f"Failed to setup computer {type(err).__name__}: {err}"
+            self.message = f"Computer setup failed! {type(err).__name__}: {err}"
```
danielhollas, line 946: Made this shorter, since long error messages do not fit the widget. We'll need a better solution for this.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -937,7 +970,7 @@ def _reset(self):
         self.label.value = ""
         self.hostname.value = ""
         self.description.value = ""
-        self.work_dir.value = ""
+        self.work_dir.value = "/home/{username}/aiida_run"
```
danielhollas, line 973: work_dir cannot be empty. Should we provide some default value like this or not??

CC @yakutovicha @unkcpz

danielhollas (commented):
@yakutovicha I've made some more improvements, can you take a look?

aiidalab_widgets_base/computational_resources.py
```diff
@@ -817,8 +817,17 @@ def observe_memory_per_machine(change):
...
-    def _configure_computer(self, computer: orm.Computer):
-        """Configure the computer"""
+    def _configure_computer(self, computer: orm.Computer, transport: str):
+        if transport == "core.ssh":
+            self._configure_computer_ssh(computer)
+        elif transport == "core.local":
+            self._configure_computer_local(computer)
+        else:
+            msg = f"invalid transport type '{transport}'"
+            raise common.exceptions.ValidationError(msg)
```
yakutovicha, line 827: `ValidationError` is a part of the `common`'s API:
```suggestion
            raise common.ValidationError(msg)
```

aiidalab_widgets_base/computational_resources.py
```diff
@@ -858,6 +866,17 @@ def _configure_computer(self, computer: orm.Computer):
 
         return True
 
+    def _configure_computer_local(self, computer: orm.Computer):
+        """Configure the computer with local transport"""
+        authparams = {
+            "use_login_shell": self.use_login_shell.value,
+            "safe_interval": self.safe_interval.value,
+        }
+        user = orm.User.collection.get_default()
```
yakutovicha, line 875: Maybe do that inside `_configure_computer` method and then pass the `user` argument to `_configure_computer_local` and `_configure_computer_ssh`? We will then use it also for the other transport types.

yakutovicha (changes requested):
Thanks, @danielhollas. I suggested a few minor changes. I also tested it, all good.

yakutovicha (approved):
I tested the latest changes - all good 👍. Ship it 🚢 !

aiidalab_widgets_base/computational_resources.py
danielhollas: I decided to revert this change. I think this needs to be a conscious and informed decision from the  user. On some HPCs, setting this to home directory could potentially lead to excessive network communication.

danielhollas:
Great, thanks @yakutovicha!

PR #504, opened 2023-08-24
StructureViewer Download tab: pass file format explicitly

danielhollas:
The change in #501 broke my application because it changed the `file_format` traitlet, on which I was relying. 

I could update my app but it would be a bit complicated to support both old and new version, so I thought I'll make a small change here if that is okay with you @yakutovicha.

Incidentally, I also fixed a small bug where `_prepare_payload` function was called twice.


yakutovicha (approved):
Thanks, @danielhollas - all good. And sorry for breaking your app.

PR #509, opened 2023-09-03
Remove warnings in unit tests

danielhollas:
Fixes #490.

TODO: Test things

I extracted some fixes that were actual bugs into separate PRs, #512 and #513. ~~This is blocked until they are merged~~ Now unblocked

PR #512, opened 2023-09-07
Fix Warning coming from ProcessFollowerWidget

danielhollas:
This was missed when we changed the ProcessMonitor to use Process UUID as its traitlet. I discovered this when I was trying to get rid of warnings in unit tests in #509 

@yakutovicha @unkcpz I am not very familiar with this Widget, could you please test it?

### DIFF
--- a/aiidalab_widgets_base/process.py
+++ b/aiidalab_widgets_base/process.py
@@ -302,7 +302,7 @@ class ProcessFollowerWidget(ipw.VBox):
 
         if self._monitor is None:
             self._monitor = ProcessMonitor(
-                process=self.process,
+                value=self.process.uuid,
                 callbacks=[self.update],
                 on_sealed=self._run_after_completed,
                 timeout=self.update_interval,
--- a/aiidalab_widgets_base/process.py
+++ b/aiidalab_widgets_base/process.py
@@ -261,7 +261,7 @@ class ProcessFollowerWidget(ipw.VBox):
         self,
         process=None,
         followers=None,
-        update_interval=1.0,
+        update_interval=0.1,
         path_to_root="../",
         **kwargs,
     ):
--- a/aiidalab_widgets_base/process.py
+++ b/aiidalab_widgets_base/process.py
@@ -261,7 +261,7 @@ class ProcessFollowerWidget(ipw.VBox):
         self,
         process=None,
         followers=None,
-        update_interval=0.1,
+        update_interval=1.0,
         path_to_root="../",
         **kwargs,
     ):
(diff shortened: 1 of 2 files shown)


aiidalab_widgets_base/process.py
```diff
@@ -261,7 +261,7 @@ def __init__(
         self,
         process=None,
         followers=None,
-        update_interval=0.1,
+        update_interval=1.0,
```
danielhollas, line 264: Seems like a more reasonable default here.

unkcpz: For the remote running process where the interval cooling time usually > 10s, `1.0` seems better, but we set localhost where `safe-interval=0.0` (https://github.com/aiidalab/aiidalab-docker-stack/blob/6cfb093b7dd91272258023d6d45e92da47986ddc/stack/base/before-notebook.d/40_prepare-aiida.sh#L63). So it makes sense to set to `0.1` to see the quick update of the calcjob.

aiidalab_widgets_base/process.py
```diff
@@ -302,7 +302,7 @@ def follow(self, detach=False):
 
         if self._monitor is None:
             self._monitor = ProcessMonitor(
-                process=self.process,
+                value=self.process.uuid,
```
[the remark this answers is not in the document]
danielhollas: ah, my mistake, I did not check the class constructor. I removed this line. 

> I think the proper way is to make the process_uuid an argument of __init__ and remove the ipw.dlink below.

I don't think we can remove the `dlink`, it is needed when the `process` traitlet of ProcessFollowerWidget changes.

I do agree that we should add a possibility of passing the UUID through the constructor (but since the traitlet is called `value` it should be named value. This would make this consistent with how ipywidgets are designed. But note that almost none of the widgets in this file are done this way so this is for a longer discussion.

aiidalab_widgets_base/process.py
danielhollas: Hmm, I am not sure I agree. `0.1` can in my experience put a lot of unnecessary CPU load on the server. Also, typical AiiDA ab initio jobs will run quite long, surely not much can happen within 0.1? Note that we've already made this default decision for the `ProcessMonitor` class, so not sure it make sense to have a different default here.

yakutovicha:
@unkcpz, since you are reviewing it, I will un-assign myself. I trust you guys :)

aiidalab_widgets_base/process.py
unkcpz: I see~ Okay, let's set it to 1.0s.

unkcpz (approved):
Looks all good, thanks!

aiidalab_widgets_base/process.py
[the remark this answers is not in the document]
danielhollas: I just realized that in the end this was not a bug, since this wrong constructor parameter was simply ignored as an extra keyword argument (and because we have the `dlink` below). That's why this was not a runtime error but just a warning in pytest. So should be safe to merge.

PR #513, opened 2023-09-07
Fix .ssh/config file not being closed

danielhollas:
Turns out we were not closing the `.ssh/config` file, which was uncovered by looking at pytest warnings when running unit tests in #509.

Also, the existing parsing logic was very brittle and did not take into account a possible variations (e.g. capitalization, extra white space etc.) So I switched to a proper parser from aiida-core, that we already use in this file anyway.
Here's a good page describing the syntax of ssh config file. https://www.ssh.com/academy/ssh/config

I also discovered other possible bugs in the code, left them as TODO for future PRs and opened #516 to track them.


aiidalab_widgets_base/computational_resources.py
```diff
@@ -425,8 +425,15 @@ def _is_in_config(self):
         fpath = Path.home() / ".ssh" / "config"
         if not fpath.exists():
             return False
-        cfglines = open(fpath).read().split("\n")
-        return "Host " + self.hostname.value in cfglines
+        sshcfg = parse_sshconfig(self.hostname.value)
+        # Confusingly, parse_sshconfig returns a dict with a hostname
+        # even if it is not in the config file.
```
yakutovicha, line 430: I am not sure why it is confusing. This is what SSH does itself. If the `Hostname` is not specified, the value is taken directly from the `Host`

aiidalab_widgets_base/computational_resources.py
```diff
@@ -425,8 +425,15 @@ def _is_in_config(self):
...
+        if len(sshcfg.keys()) < 2:
+            return False
```
yakutovicha, line 432: I believe this condition is redundant. The hostname will be always present in the dictionary (even if the host is not present in the `.ssh/config` file. The other required key is the username, which is checked in the next line. So I would omit this check entirely.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -477,8 +484,12 @@ def _on_setup_ssh_button_pressed(self, _=None):
 
             # Write private key in ~/.ssh/ and use the name of upload file,
             # if exist, generate random string and append to filename then override current name.
+            # TODO(danielhollas): I don't think this works as intended. If there is an existing private key,
+            # the new filename is never propagated to the caller here.
             self._add_private_key(private_key_abs_fname, private_key_content)
 
+        # TODO(danielhollas): I am not sure this is correct. What if the user wants
+        # to overwrite the private key? The configuration would never be written.
```
yakutovicha, line 492: Given that the primary focus of the PR isn't on this topic, could we move this information to an issue instead? It seems unnecessary to have multiple locations where information about potential problems can be accessed.

yakutovicha (changes requested):
Thanks, @danielhollas. A few suggestions.

yakutovicha:
@danielhollas please re-request my review when the PR is ready.

aiidalab_widgets_base/computational_resources.py
danielhollas: I am not sure what you mean by `Host`. The function is called `parse_sshconfig` and we pass it the hostname that we're interested in. I would expect that if the given hostname is not in the config file, that this function returns None or something like that. Anyway, I've reworded the comment a bit , but I would like to keep it to make things clearer.

aiidalab_widgets_base/computational_resources.py
danielhollas: Yeah, I thought about this as well. I've removed it.

aiidalab_widgets_base/computational_resources.py
danielhollas: I've opened #516 to track these issues. However, I would like to keep the TODOs in the code. I think it is quite important because if somebody wanders into this code later, they shouldn't need to know about every open issue on Github. I've added the link to the GH issue so that these things are crosslinked.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -421,12 +421,17 @@ def _can_login(self):
...
         fpath = Path.home() / ".ssh" / "config"
         if not fpath.exists():
             return False
-        cfglines = open(fpath).read().split("\n")
-        return "Host " + self.hostname.value in cfglines
+        sshcfg = parse_sshconfig(self.hostname.value)
+        # NOTE: parse_sshconfig returns a dict with a hostname
+        # even if it is not in the config file.
+        # We require at least the user to be specified.
+        if "user" not in sshcfg:
```
unkcpz, line 432: Maybe it is worth setting the `self.message` with the warning.

unkcpz (approved):
Thanks @danielhollas. The change looks good for me, one minor request.

aiidalab_widgets_base/computational_resources.py
danielhollas: I don't think so. This function is meant to check if the hostname is in config, but it is not an error if it is not there, since the user is typically setting it up when this function is called.

unkcpz: I'll let @yakutovicha have another look at it, all good for me.

yakutovicha (approved):
LGTM!

aiidalab_widgets_base/computational_resources.py
yakutovicha: > I am not sure what you mean by `Host`.

I was referring to the ssh configuration. Suppose it looks like that.
```
Host my_computer
    HostName my_computer.com
    User me
```

In that context, if the `HostName` (`my_computer.com`) is not set in the `~/.ssh/config`, it is taken directly from the `Host` (`my_computer`). Consequently, if nothing is provided at all, the ssh command `ssh my_computer` (where `my_computer` is the `Host`) would imply that you are requesting a `Hostname` `my_computer`.

So, it is not surprising to me that if you request an SSH configuration for a host "whatever", the `Hostname` setting can be identified already based on your request.

That is how I would interpret the behaviour of the function.

danielhollas: Interesting, I was unaware of the difference between Host and Hostname. Thanks!

PR #514, opened 2023-09-08
DOC: re-strucuture the doc folder

unkcpz:


unkcpz:
Looks better now https://aiidalab-widgets-base--514.org.readthedocs.build/en/514/

docs/source/conf.py
```diff
@@ -46,7 +41,7 @@
 
 # General information about the project.
 project = "aiidalab-widgets-base"
-copyright_first_year = "2020"
+copyright_first_year = "2023"
```
yakutovicha, line 44: Shouldn't it be the year the documentation was created?

yakutovicha (commented):
Very minor question, otherwise good 👍

docs/source/conf.py
unkcpz: You are correct, it didn't notice "first_year" :laughing:

unkcpz, line 44:
```suggestion
copyright_first_year = "2020"
```

yakutovicha (approved):
I actually need those changes very much, because I run into the same issues that you fixed here. So let's merge it 😎 

Thanks, @unkcpz!

PR #518, opened 2023-09-26
Remove openbabel dependency

danielhollas:
As discussed in https://github.com/aiidalab/aiidalab-docker-stack/issues/388 we want to remove openbabel dependency from AWB because it cannot be installed by pip. We use it here only as a fallback to RDKit in SmilesWidget. 

In my experience, it is not needed once the RDkit SMILES generation is more stable. @cpignedoli I would be interested if you have any SMILES that needed Openbabel. Please open a separate issue if you have any that still don't work for the current code. Thanks!

This also includes a little cleanup of the canonicalization code.

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -685,13 +685,6 @@ class SmilesWidget(ipw.VBox):
 
     def __init__(self, title=""):
         self.title = title
-
-        try:
-            from openbabel import openbabel  # noqa: F401
-            from openbabel import pybel  # noqa: F401
-        except ImportError:
-            self.disable_openbabel = True
-
         try:  # noqa: TC101
             from rdkit import Chem  # noqa: F401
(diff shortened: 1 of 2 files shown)


cpignedoli:
Good point @danielhollas indeed I do not find anymore cases that justify openbabel so from my side we can remove it and if I find some new problematic cases we can think about how to handle them.

I tested an old molecule I had that gave me problems but also with openbabel seems not to work.

I add it here as another problematic case not working at present:
[H][C]1C2=CC=CC3=C2C(C4=C1C=CC5=C4C6=C(C7=C([C]([H])C8=CC=CC9=CC=CC7=C98)C=C%10)C%10=CC=C6C=C5)=CC=C3

a correct (not best conformer) interpretation of this helicoidal molecule is provided by Mathematica:

image

yakutovicha (approved):
Thanks @danielhollas, I tested it locally and things seem to work fine 👍

PR #522, opened 2023-10-04
Structure viewer for 0d hide the cell tab

### LINKED ISSUE #351: Hide the Cell tab when the structure is not periodic
The `StructureDataviewer` will show the cell tab if the cell tab is initialized when creating the viewer. 
It makes no sense to show the cell information for the non-periodic structure. 
So when the structure detected is not 0-D structure the `cell` tab should be completely hidden. 

But for the 1-D and 2-D structures, there are options here, either we can hide the cell tab completely or we can show partial information with the cell.

unkcpz:
fixes #351 

- [ ] unit test on the newly added feature.

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -376,32 +376,6 @@ class StructureUploadWidget(ipw.VBox):
             children=[self.file_upload, supported_formats, self._status_message]
         )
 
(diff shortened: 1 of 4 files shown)


danielhollas:
I am going to close this for now. Ultimately I think hiding the tab might be more confusing than it's worth it.

PR #526, opened 2023-10-13
Bump traitlets to v5.9

danielhollas:
New version of traitlets contains optimizations that significantly speed up loading of the app.

As discussed in https://github.com/danielhollas/aiidalab-widgets-base/pull/new/deps/bump-traitlets


danielhollas:
Hmm, the notebook tests are failing during `pip install`, but only with Firefox which is very very strange. Need to investigate more, but unfortunately I don't see the error in the output, the installation seems to finish fine.

unkcpz (commented):
> but only with Firefox which is very very strange. 

I saw chrome notebook test also failed. 
What is that `free -m` for?

danielhollas:
> I saw chrome notebook test also failed. What is that free -m for?

Yeah, it's very annoying. :anguished: The original errors indicated that the pip process got killed by the OS OOM killer at the very end of the installation (there was no error message and the process return exit code 137, i.e. SIGKILL). I wanted to see how much memory is available inside the container. But now I see errors coming from the actual selenium tests, so perhaps there are real issues. Need to investigate more, but will probably not get to it today.

I hoped this would go smoother. :sob:

unkcpz:
Maybe I can try to use the same way as we now test the notebook of QeApp, the image is build and the package is installed inside the container. It then makes the container test easy to reproduce locally. 
If there is no need to consider multi-arch and dockerhub uploaded, then the CI for the docker image build should be simple.

danielhollas:
@unkcpz not sure what you mean, but we already do install AWB inside the container when running the notebook tests, see fixtures in the conftest.

I got ill over the weekend so will not be able to work on this today, and probably in the next few days. Feel free to continue. Imo the first thing is to just try installing AWB from this branch locally and verify that it works.(I should have done that myself). (and perhaps monitor pips use of memory during install).

unkcpz:
Here is how qe app build the qe image with the things installed https://github.com/aiidalab/aiidalab-qe/blob/fc36106f315b2c46ec50995b048b4aa9bf5e5cd3/.github/workflows/docker-build-test-upload.yml#L48-L62

then there is not need to install the app through `conftest.py`. See https://github.com/aiidalab/aiidalab-qe/blob/fc36106f315b2c46ec50995b048b4aa9bf5e5cd3/tests_integration/conftest.py#L70-L93

unkcpz:
Hehe, seems to work now. Time is the master of solving problems.

.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
             matrix:
                 tag: [latest]
                 browser: [Chrome, Firefox]
-                python-version: ['3.10']
+                python-version: ['3.9']
```
unkcpz, line 46:
```suggestion
                python-version: ['3.10']
```

unkcpz: let me try to change this back to 3.10 see if it still fine.

tests_notebooks/docker-compose.yml
```diff
@@ -16,3 +16,7 @@ services:
             - ..:/home/jovyan/apps/aiidalab-widgets-base
         ports:
             - 8998:8888
+        deploy:
+            resources:
+                limits:
+                    memory: 2000M
```
unkcpz, line 22: Not sure if is this still required. I'd try to avoid adding it if not needed.

unkcpz (changes requested):
Notebook tests all pass 🥂 ! Just one minor request, see if it is correct.

unkcpz:
Emm... Now it is the unit test that stuck. I ran locally and it stuck at `tests/test_databases.py::test_cod_query_widget`, but it happened to the master branch as well, thus nothing related to the changes in this PR.

tests_notebooks/docker-compose.yml
danielhollas, line 22: Yeah, I don't think memory was the issue.

```suggestion
```

unkcpz (approved):
All tests passed. (finally :smile: )

danielhollas:
Okay, let's try this out. @unkcpz could you please release a new beta version? It would be great if you could also test this together with aiida-core 2.4

PR #527, opened 2023-10-17
Restructure to build image and run integration test on it

unkcpz:


unkcpz:
I have to admit this seems not that useful as I thought after I did the change. For the qeapp, the benefits are straightforward because the image is the final goal and using it to do the integration test comes as a positive side effect. 
Here, in order to run the integration test, the image has to be built first, which is not that bad although increasing the complexity and has small benefits that the installation of the package happened in the build phase which from my perspective easy to control than installed by conftest. 

I would avoid to go this way.

unkcpz:
I close this, because it is not a good idea to use such a complex way to build and test notebook.

PR #535, opened 2023-11-06
Draw bonds computed by ASE

### LINKED ISSUE #48: nglview: figure out why bonds are not always shown by deffault

yakutovicha:
fixes #48
supercedes #377 

In nglview if a system is too big bonds will not be computed.
I bypass the "ball+stick" representation defining it as "spacefill" plus shapes that are computed via ase connectivity matrix.

I would not draw also the atom spheres through  shapes otherwise it would be a problem to pick atoms with mouse selection

Despite I tested it also with ~1000 atoms it is meant to work on smaller system sizes.

The image below is outdated, just the "ball+stick" representation is needed to produce the image

image (15)

To Do:

- [ ] Consider using Enum for the dropdown widgets.


yakutovicha:
> [ ] Consider using Enum for the dropdown widgets.

This item is not yet implemented, but I think we can discuss it first during today's meeting and then decide what to do.

The rest is ready, please review.

aiidalab_widgets_base/viewers.py
```diff
@@ -554,6 +555,59 @@ def _observe_all_representations(self, change):
...
+                tuple(v2),
+                tuple(color),
+                radius,
+            )
+
+        # Choose the correct function for computing the cylinder.
+        compute_cylinder = povray_cylinder if povray else cylinder
+
+        if len(structure) > 1:
+            ii, jj, bond_lengths = ase.neighborlist.neighbor_list(
```
superstar54, line 596: It's a distance vector, not the length.
```suggestion
            ii, jj, bond_vectors = ase.neighborlist.neighbor_list(
```

aiidalab_widgets_base/viewers.py
```diff
@@ -554,6 +555,59 @@ def _observe_all_representations(self, change):
...
+                "ijD", structure, cutoff, self_interaction=False
+            )
+            for id1, id2, bond_length in zip(ii, jj, bond_lengths):
+                atom1, atom2 = structure[id1], structure[id2]
+
+                # Bond parameters
+                v1 = np.array([atom1.x, atom1.y, atom1.z])
+                half_bond_point = v1 + bond_length * Radius[atom1.symbol] / (
+                    Radius[atom1.symbol] + Radius[atom2.symbol]
+                )
```
superstar54, line 606: The math here is wrong. You need both the `vector` and the `length` to get the middle point.

aiidalab_widgets_base/viewers.py
```diff
@@ -1175,18 +1195,31 @@ def _observe_displayed_structure(self, change):
...
+                        # Add bonds if ball+stick representation is used.
+                        if representation.type.value == "ball+stick":
+                            bonds += self._compute_bonds(
+                                self.displayed_structure[indices],
+                                representation.size.value * 0.2,
+                                representation.color.value,
+                            )
                 self._viewer.set_representations(nglview_params, component=0)
                 self._viewer.add_unitcell()
+                self._viewer._add_shape(set(bonds), name="bonds")
```
superstar54, line 1220: Even though it is not obvious in the GUI, but you add new shapes for the bonds, and haven't removed the default bond from the nglview itself, thus the two bonds overlap.

superstar54 (changes requested):
Hi @yakutovicha  Thanks for the work. It's important for the large structure. I made a few change requests.

Also two comments:
- better to add a test for the `_compute_bonds` function.
- There is a variable called `uuid`, which is the `Unique identifier for the representation.`. At first glance, I mixed it with the AiiDA node uuid. I suggest using another name, e.g. `style_id`.

.github/workflows/ci.yml
```diff
@@ -119,6 +119,9 @@ jobs:
             - name: Check out app
               uses: actions/checkout@v3
 
+            - name: Install povray
+              run: sudo apt-get install povray
```
unkcpz, line 123: Is the `povray` in the `full-stack`? I remember you proposed to add it?

aiidalab_widgets_base/dicts.py
```diff
@@ -221,3 +221,21 @@
     "Hs": 1.75,
     "Mt": 1.75,
 }
+
+RGB_colors = {
```
unkcpz, line 225: The value is used as a tuple by converting from a list, so why not define the value as tuple directly?

aiidalab_widgets_base/viewers.py
```diff
@@ -233,7 +234,7 @@ def atoms_in_representaion(self, structure: ase.Atoms | None = None):
         if structure:
             if self.uuid in structure.arrays:
                 return structure.arrays[self.uuid] >= self.atom_show_threshold
```
unkcpz, line 236: Not from this PR, but since you are here. 
Because python if..else if short-circuting, I think this condition can be simplified to:

```suggestion
        if structure and self.uuid in structure.arrays:
            return structure.arrays[self.uuid] >= self.atom_show_threshold
```

aiidalab_widgets_base/viewers.py
```diff
@@ -249,8 +250,8 @@ def nglview_parameters(self, indices):
         }
         if self.type.value == "ball+stick":
             nglview_parameters_dict["params"]["aspectRatio"] = self.size.value
-        else:
-            nglview_parameters_dict["params"]["radiusScale"] = 0.1 * self.size.value
+        elif self.type.value == "spacefill":
```
unkcpz, line 253: Maybe add an else for not support type value and raise exception?

.github/workflows/ci.yml
danielhollas: Yes, povray is in the image. But this is needed here for the unit tests, since they do not run in the container.

danielhollas (commented):
You already got two great reviews so I am leaving this one alone. :-)

aiidalab_widgets_base/dicts.py
yakutovicha: Thanks, done in https://github.com/aiidalab/aiidalab-widgets-base/pull/535/commits/011ec6e180875d7aadd2f90e06fd151a79a8399e. I also renamed RGB_colors to RGB_COLORS, since it is a constant.

aiidalab_widgets_base/viewers.py
```diff
@@ -554,6 +555,59 @@ def _observe_all_representations(self, change):
...
+
+        if len(structure) > 1:
+            ii, jj, bond_lengths = ase.neighborlist.neighbor_list(
+                "ijD", structure, cutoff, self_interaction=False
+            )
+            for id1, id2, bond_length in zip(ii, jj, bond_lengths):
+                atom1, atom2 = structure[id1], structure[id2]
+
+                # Bond parameters
+                v1 = np.array([atom1.x, atom1.y, atom1.z])
```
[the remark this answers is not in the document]
yakutovicha: good catch, thanks!

aiidalab_widgets_base/viewers.py
yakutovicha: I won't be able to test it, since `self.type` is a Dropdown that doesn't have another choice.

yakutovicha:
> * better to add a test for the `_compute_bonds` function.

This function is executed during text, you can check the[ codecov report](https://app.codecov.io/gh/aiidalab/aiidalab-widgets-base/pull/535/blob/aiidalab_widgets_base/viewers.py), for instance.

> * There is a variable called `uuid`, which is the `Unique identifier for the representation.`. At first glance, I mixed it with the AiiDA node uuid. I suggest using another name, e.g. `style_id`.

Done in https://github.com/aiidalab/aiidalab-widgets-base/pull/535/commits/09b3704af2bfc196f9f99bfe0359cc63142d85d4, thanks!

aiidalab_widgets_base/viewers.py
yakutovicha: Thanks, changed in https://github.com/aiidalab/aiidalab-widgets-base/pull/535/commits/6ff62b473b3227557d6e6549eb5193ee7bafb7c7

aiidalab_widgets_base/viewers.py
yakutovicha: The naming here is confusing, I guess. This is to place the point where the bond changes colour. We find this point using the proportion of the atomic radius of the atoms.

I renamed the variable to `intermediate_point` in https://github.com/aiidalab/aiidalab-widgets-base/pull/535/commits/d4f0a063dc36c9c6409db73c2a6d326fa78e3fdc

aiidalab_widgets_base/viewers.py
```diff
@@ -233,7 +234,7 @@ def atoms_in_representaion(self, structure: ase.Atoms | None = None):
         if structure:
             if self.uuid in structure.arrays:
                 return structure.arrays[self.uuid] >= self.atom_show_threshold
-        return None
+        return []
```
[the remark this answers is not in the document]
yakutovicha: One can, I just tried:
```python
In [1]: import numpy as np

In [2]: np.array([])
Out[2]: array([], dtype=float64)
```

aiidalab_widgets_base/viewers.py
unkcpz: Fair. `else` is not always necessary :)

unkcpz:
I guess test failed because of dockerhub rate limit, https://github.com/aiidalab/aiidalab-widgets-base/pull/539 should fix it.

aiidalab_widgets_base/viewers.py
yakutovicha: That was a hard nut to crash, but I think I managed in https://github.com/aiidalab/aiidalab-widgets-base/pull/535/commits/066aa4a5a567b79c726b04d0a002b2b64e5f6792. What do you think @cpignedoli?

yakutovicha:
@cpignedoli please comment on where all the numbers are coming from.

aiidalab_widgets_base/viewers.py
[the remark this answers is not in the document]
superstar54: sorry that my comment is not clear. I mean these lines
```
missing_atoms = np.zeros(self.natoms)
missing_atoms += rep.atoms_in_representaion(self.structure)
```
In case of an empty list
```
>>> import numpy as np
>>> missing_atoms = np.zeros(4)
>>> missing_atoms += []
  File "<stdin>", line 1, in <module>
ValueError: operands could not be broadcast together with shapes (4,) (0,) (4,) 
>>> 
```

aiidalab_widgets_base/viewers.py
superstar54: > We find this point using the proportion of the atomic radius of the atoms.

No. The correct math is:
```
# get the bond_vector from the `neighbor_list` too.
normal = np.linalg.norm(bond_vector)
r1 = Radius[atom1.symbol]
r2 = Radius[atom2.symbol]
intermediate_point = v1 + r1*normal + (bond_length - r1 - r2)/2*normal
# you can use the short term:  intermediate_point = v1 + (bond_length + r1 - r2)/2*normal
```

superstar54:
Hi @yakutovicha, Thanks for the update! I am still worried about the test on the bond.

> This function is executed during text, you can check the[ codecov report](https://app.codecov.io/gh/aiidalab/aiidalab-widgets-base/pull/535/blob/aiidalab_widgets_base/viewers.py), for instance.

It's executed, but the result isn't verified. I would suggest running this `function` and checking the result `bonds` using a simple structure as input, e.g., H2O. See my comment on the math to calculate the `intermediate_point`.

aiidalab_widgets_base/viewers.py
superstar54: After setting different radii for different atoms in this commit, the wrong `intermediate_point` becomes obvious. Before, all atoms have the same radius in the `ball+stick` model, thus `intermediate_point` is  (only) correct in that case.

aiidalab_widgets_base/viewers.py
```diff
@@ -554,6 +555,59 @@ def _observe_all_representations(self, change):
         if change["new"]:
             self._all_representations[-1].viewer_class = self
 
+    def _compute_bonds(self, structure, radius=1.0, color="element", povray=False):
+        """Create an array of bonds for the structure."""
+
+        import ase.neighborlist
+
+        radius = radius / 5
```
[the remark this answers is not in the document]
cpignedoli: I set these numbers more than one year ago, I was just tuning them to have a "nice" aspect ratio. Feel free to change and motivate

aiidalab_widgets_base/viewers.py
```diff
@@ -554,6 +555,59 @@ def _observe_all_representations(self, change):
...
+
+        cutoff = ase.neighborlist.natural_cutoffs(structure, mult=1.09)
```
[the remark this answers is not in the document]
cpignedoli: Same as above, while testing on a number of systems in some cases I had bonds where I did not 'want' them in in other cases I did not have bonds and I wanted them. This 1.09 was a good compromise. Feel free to change to a value that produces even better results.

aiidalab_widgets_base/viewers.py
cpignedoli: I am not sure we do not remove default bonds: we never use the ngl representation "ball+stik" we always use spacefill (no bonds) and add bond shapes in the case we use the "fake ball+stik"

cpignedoli: v1 + bond_vector * Radius[atom1.symbol] / (
                    Radius[atom1.symbol] + Radius[atom2.symbol]
bond_vector goes from atom1 to atom2.
we will add to atom1 a vector bond_v*r1/(r1+r2) and to atom2 a vector, in opposite direction (or from intermediate point to atom2), bond_v*r2/(r1+r2) in total: bond_v*(r1+r2)/(r1+r2) = bond_v and the two portions of bonds will have length proportional to the size of the spheres

aiidalab_widgets_base/viewers.py
yakutovicha: I don't understand what is wrong with our approach.

`pos1` - location of the first atom
`pos2` - location of the second atom
`bond_vector` is a vector connecting the first atom and the second atom.
`r1` - is the vdW radius of the first atom
`r2` - is the vdW radius of the first atom

The intermediate point is a percentage of the vector r1 relative its contribution to the sum of two vdW radii.

`bond_vector = pos2 - pos1`

`atom1_contribution = r1 / (r1 + r2)`
`atom2_contribution = r2 / (r1 + r2)`

`intermediate_point = pos1 + bond_vector * atom1_contribution`

or

`intermediate_point = pos2 + bond_vector * atom2_contribution`

`atom1_contribution + atom2_contribution = r1 / (r1 + r2) + r2 / (r1 + r2) = (r1 + r2) / (r1 + f2)`

I'd appreciate it if you could take a moment to double-check our work before mentioning any concerns about the math. Thank you!

aiidalab_widgets_base/viewers.py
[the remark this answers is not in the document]
yakutovicha: Great, thanks. I see now what could be the problem. Fixed in https://github.com/aiidalab/aiidalab-widgets-base/pull/535/commits/fc815dc6619b91e0e7b30620233751f2ab85ac5d

aiidalab_widgets_base/viewers.py
yakutovicha: For the record, we discussed with @superstar54 and @cpignedoli, where we agreed to cut the bond in the middle of the visible part.

yakutovicha: done in https://github.com/aiidalab/aiidalab-widgets-base/pull/535/commits/9ee0f991412f276cc77bb4890c717020e01fbe25

aiidalab_widgets_base/viewers.py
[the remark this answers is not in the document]
yakutovicha: done in https://github.com/aiidalab/aiidalab-widgets-base/pull/535/commits/6e547fc5957793efe8450003be81ffb371e3d7bc

aiidalab_widgets_base/viewers.py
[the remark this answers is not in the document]
yakutovicha: done in https://github.com/aiidalab/aiidalab-widgets-base/pull/535/commits/6e547fc5957793efe8450003be81ffb371e3d7bc

yakutovicha:
Test for the bond function is added in https://github.com/aiidalab/aiidalab-widgets-base/pull/535/commits/d780b5a1c13065aa2bb8dfb225bc978c12724c2b

unkcpz (dismissed):
My requests are all addressed, you have my approval. Thanks for the nice work! @yakutovicha

aiidalab_widgets_base/viewers.py
```diff
@@ -554,9 +555,63 @@ def _observe_all_representations(self, change):
...
+
+        # The radius is scaled by 0.04 to have a better visual appearance.
+        radius = radius * 0.04
+
+        # The value 1.09 is chosen based on our experience. It is a good compromise between showing too many bonds
+        # and not showing bonds that should be there.
+        cutoff = ase.neighborlist.natural_cutoffs(structure, mult=1.09)
+        bonds = []
+
+        def povray_cylinder(v1, v2, radius, atom, color):
```
superstar54, line 571: the arg `color` is not used.

aiidalab_widgets_base/viewers.py
```diff
@@ -554,9 +555,63 @@ def _observe_all_representations(self, change):
...
+                radius,
+            )
+
+        # Choose the correct function for computing the cylinder.
+        compute_cylinder = povray_cylinder if povray else cylinder
+
+        if len(structure) > 1:
+            ii, _, bond_vectors = ase.neighborlist.neighbor_list(
+                "ijD", structure, cutoff, self_interaction=False
+            )
```
superstar54, line 601:
```suggestion
            ii, bond_vectors = ase.neighborlist.neighbor_list(
                "iD", structure, cutoff, self_interaction=False
            )
```

superstar54 (changes requested):
Hi @yakutovicha, Thanks for the update. Since `bond` is an important component in the `viewer`, and this PR is supposed to handle large systems with > 1000 bonds. So I added a comment on the performance.

superstar54:
@unkcpz @yakutovicha I moved the `if` outside the `for` loop, which makes the code 1 time faster. Please review.

aiidalab_widgets_base/viewers.py
```diff
@@ -554,9 +555,74 @@ def _observe_all_representations(self, change):
...
             self._all_representations[-1].viewer_class = self
 
+    def _compute_bonds(self, structure, radius=1.0, color="element", povray=False):
+        """Create an array of bonds for the structure."""
+
+        import ase.neighborlist
+
+        bonds = []
+        if len(structure) <= 1:
+            return bonds
```
unkcpz, line 565:
```suggestion
        if len(structure) <= 1:
            return []
```

aiidalab_widgets_base/viewers.py
```diff
@@ -554,9 +555,74 @@ def _observe_all_representations(self, change):
         if change["new"]:
             self._all_representations[-1].viewer_class = self
 
+    def _compute_bonds(self, structure, radius=1.0, color="element", povray=False):
+        """Create an array of bonds for the structure."""
```
unkcpz, line 559: Create an array? The return value is a list.

aiidalab_widgets_base/viewers.py
```diff
@@ -554,9 +555,74 @@ def _observe_all_representations(self, change):
...
+            "iD", structure, cutoff, self_interaction=False
+        )
+        nb = len(ii)
+        # bond start position
+        v1 = structure.positions[ii]
+        # middle position
+        v2 = v1 + bond_vectors * 0.5
+
+        # Choose the correct way for computing the cylinder.
+        def povray_cylinder(v1, v2, radius, color):
```
unkcpz, line 583: I agree using function makes it more clean for me. Can you move it out as `staticmethod` instead of nesting it here? If it is only used in the class, change the name to `_povray_cylinder`. Same for `cylinder` -> `_cylinder` and move it out.

unkcpz (changes requested):
Some small requests should be easy to address. @superstar54

aiidalab_widgets_base/viewers.py
superstar54: Thanks, I changed the inner function into a private helper function.

superstar54:
Hi @unkcpz, thanks for the review. I made the changes as you suggested. Please have another look.

unkcpz (approved):
Thanks! @superstar54 Looks all good.

superstar54 (approved):

PR #536, opened 2023-11-08
Add timeout to cod query test to avoid test hang

unkcpz:
Temporary solution to timeout the cod query widget test.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,6 +49,7 @@ dev =
     pytest-cov~=4.0
     pytest-docker~=1.0
     pytest-selenium~=4.0
+    pytest-timeout~=2.2
     selenium~=4.7.0
     webdriver-manager~=3.8
 smiles =
--- a/tests/test_databases.py
+++ b/tests/test_databases.py
@@ -1,9 +1,12 @@
 import ase
+import pytest
 
 import aiidalab_widgets_base as awb
 from aiidalab_widgets_base.databases import ComputationalResourcesDatabaseWidget
 
 
+# timeout for the test
+@pytest.mark.timeout(30)
 def test_cod_query_widget():
     """Test the COD query widget."""
 


danielhollas:
Hmm, I would think that the timeout needs to be placed after the widget is initialized inside the test?

unkcpz:
I want to timeout the test itself to not block the other tests, although it will failed in timeout which is expected for this case, after the cod is back, the test will pass.

danielhollas:
I don't understand. Why is it expected that the test should timeout? We ideally want to fix the test itself no? Surely when the test times out it should fail the test suite, even if all other tests pass.

unkcpz:
> Why is it expected that the test should timeout?

I don't think we can "fix" it, the problem is for the cod server side. Once the server is working, the test will pass. So if we find only this test failed and the cod server is not accessible, we can still continue with merging the PRs.

danielhollas:
>  the problem is for the cod server side. Once the server is working,

Okay, could you please open an issue and explain in detail what the problem is? In the meantime, I would suggest to simply skip the test (`@pytest.mark.skip` I think). That's a much easier and more robust solution than the timeout.

yakutovicha:
The test is passing now, btw.

unkcpz:
@yakutovicha thanks!
I think it is not a bad idea to add the timeout, then we don't stuck the test in the future if the cod server not accessible again in the future (although quite rare I hope). If skip it, then it requires to change the test by PRs. If you think it not worth to do like this I'll close the PR.

yakutovicha:
I think it is not a bad idea to add the timeout, then we don't stuck the test in the future if the cod server not accessible again in the future (although quite rare I hope).

So as far as I can see, adding the timeout doesn't solve the test problem itself, but just speeds up the test failure in case of trouble. Which is not a bad thing, in the end.

To resolve the issue we might monkey-patch the output of the request sent to the COD, but that is for another PR. Not sure if this is a good idea, though. Because if COD changes something, then the problem might go unnoticed.

danielhollas:
Sure, let's try it, but the tests are currently failing??

unkcpz:
I didn't retrigger the test after COD was alive. All passed now :)

danielhollas (approved):

PR #537, opened 2023-11-10
Allow set key_filename/key_policy in computer configure

unkcpz:
There are some values for authinfo setting that are hardcoded. It should be able to be set from computer-configure dict. This is required by the CSCS MFA configure setup. The key filename is set in configure to `cscs-key` but not passed to the `AuthInfo`.

### DIFF
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -839,7 +839,19 @@ class AiidaComputerSetup(ipw.VBox):
             raise common.ValidationError(msg)
 
     def _configure_computer_ssh(self, computer: orm.Computer, user: orm.User):
-        """Configure the computer with SSH transport"""
+        """Configure the computer with SSH transport
+
+        There are three sources of authparams information:
+        - the SSH config file
+        - the computer_configure dictionary
+        - the default values
+
+        Priority is given to the computer_configure dictionary, then to the SSH config file, then to the default values.
+        At the moment, there is no overlap between the SSH config file and the computer_configure dictionary.
+
+        The proxyjump and proxycommend can be read from both the SSH config file and the computer_configure dictionary,
+        since the SSH config file is generated from the computer_configure dictionary in `SshComputerSetup._write_ssh_config`.
+        """
         sshcfg = aiida_ssh_plugin.parse_sshconfig(self.hostname.value)
         authparams = {
             "port": int(sshcfg.get("port", 22)),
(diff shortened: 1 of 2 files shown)


aiidalab_widgets_base/computational_resources.py
```diff
@@ -871,10 +883,17 @@ def _configure_computer_ssh(self, computer: orm.Computer, user: orm.User):
...
+            authparams["proxy_jump"] = self.computer_configure["proxy_jump"]
+
+        if "proxy_command" in self.computer_configure:
+            authparams["proxy_command"] = self.computer_configure["proxy_command"]
+
+        if "key_filename" in self.computer_configure:
+            authparams["key_filename"] = self.computer_configure["key_filename"]
+
+        if "key_policy" in self.computer_configure:
+            authparams["key_policy"] = self.computer_configure["key_policy"]
```
unkcpz, line 896: Such if clause can be simplified and moved to the `authparams`, but since there are still three sources for the setup, maybe good to keep it as it is?

aiidalab_widgets_base/computational_resources.py
```diff
@@ -871,10 +883,19 @@ def _configure_computer_ssh(self, computer: orm.Computer, user: orm.User):
...
+        if "proxy_jump" in self.computer_configure:
+            authparams["proxy_jump"] = self.computer_configure["proxy_jump"]
+
+        if "proxy_command" in self.computer_configure:
+            authparams["proxy_command"] = self.computer_configure["proxy_command"]
+
+        if "key_filename" in self.computer_configure:
+            authparams["key_filename"] = os.path.expanduser(
+                self.computer_configure["key_filename"]
+            )
```
yakutovicha, line 895: Small nitpick. I am trying to convert to Pathlib everywhere. If you could do something like [this](https://stackoverflow.com/questions/57248819/how-to-use-pathlib-path-expanduser-and-amend-and-use-a-posixpath-value) here - it would be awesome.

yakutovicha: There is another occurrence of `os.path.expanduser` in this file. Please change there too.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -839,7 +839,19 @@ def _configure_computer(self, computer: orm.Computer, transport: str):
             raise common.ValidationError(msg)
 
     def _configure_computer_ssh(self, computer: orm.Computer, user: orm.User):
-        """Configure the computer with SSH transport"""
+        """Configure the computer with SSH transport
+
+        There are three sources of authparams information:
```
yakutovicha, line 844:
```suggestion
        There are three sources of authparams information ordered by priority increase:
```

aiidalab_widgets_base/computational_resources.py
```diff
@@ -839,7 +839,19 @@ def _configure_computer(self, computer: orm.Computer, transport: str):
...
+        - the SSH config file
+        - the computer_configure dictionary
+        - the default values
```
yakutovicha, line 847:
```suggestion
        - the default values
        - the SSH config file
        - the computer_configure dictionary
```

aiidalab_widgets_base/computational_resources.py
```diff
@@ -839,7 +839,19 @@ def _configure_computer(self, computer: orm.Computer, transport: str):
...
+
+        Priority is given to the computer_configure dictionary, then to the SSH config file, then to the default values.
```
yakutovicha, line 849:
```suggestion
```

yakutovicha (changes requested):
Thanks, @unkcpz, I have a bunch of little requests.

aiidalab_widgets_base/computational_resources.py
unkcpz: Yes, changed. I agree that moving to using `Path` is better.

yakutovicha (approved):
LGTM!

PR #538, opened 2023-11-10
Fix the fail tests of resource setup after the resource database updated

unkcpz:


tests/test_computational_resources.py
```diff
@@ -584,7 +584,7 @@ def test_resource_setup_widget_for_password_configure(monkeypatch, tmp_path):
     # and the computer/code setup widget will be updated accordingly.
     w.comp_resources_database.domain_selector.value = "merlin.psi.ch"
     w.comp_resources_database.computer_selector.value = "cpu"
-    w.comp_resources_database.code_selector.value = "QE-7.0-exe-template"
+    w.comp_resources_database.code_selector.value = "QuantumESPRESSO-7.2"
```
unkcpz, line 587:
```suggestion
    w.comp_resources_database.code_selector.value = "QuantumESPRESSO-7.0"
```

unkcpz:
Offline in-person review by @superstar54, and he is on the train now. Since it breaks all the tests, I merge it so as not to block other PRs.

PR #541, opened 2023-11-20
Allow register different viewers for different WorkChain.

### LINKED ISSUE #540: Can not register different viewers for different WorkChain.
In the `AiidaNodeViewWidget`, it use the `node_type` as the [key](https://github.com/aiidalab/aiidalab-widgets-base/blob/aba0bdb3dc7a7e1af268145a0fd643194e681523/aiidalab_widgets_base/viewers.py#L51-L54) to select the `viewer` for an AiiDA node. This works fine for the `data` node, but will fail in the case of a `process` node. For example,  the `node_type` is the same for all `WorkChainNode`s, thus, one can not register different viewers for different WorkChain.

Better to use the `process_label` for the `process` node

superstar54:
Fix #540 , use `process_type` as the node viewer's key.

Here is an example, https://github.com/aiidalab/aiidalab-qe/pull/559


danielhollas:
I think this will need some discussion, since it would be a breaking change?

superstar54:
Now it has backward capacity. I agree we need more discussion on it.

danielhollas:
Thanks @superstar54. Can you take a look at the failing tests? Looks like the failures might not be related to this PR, but should be fixed regardless.

superstar54:
The failed test is caused by a [update](https://github.com/aiidateam/aiida-resource-registry/commits/main) in the template in `aiida-resource-registry` repo, which add a new `version` field.
 I fixed it.

danielhollas:
@superstar54 thank you for fixing the test. Could you please make a separate PR with that fix? I assume the test failures would pop up on other PRs as well?

Does this PR need to be merged for 2.1.0 version?

danielhollas (changes requested):
I've thought about this a bit more and it seems like a good feature to have. 
The only thing missing here is a unit test for the new behaviour.

superstar54:
> The only thing missing here is a unit test for the new behaviour.

Hi @danielhollas , I added a unit test for this behavior. Could you please have a look?

danielhollas (approved):
@superstar54 perfect, thanks!

PR #545, opened 2023-12-18
Doc: render the widgets in notebook

### LINKED ISSUE #533: Render notebooks for the documentation
To have some visual help, we should enable the notebook rendering when building the documentation. This can be achieved by setting the [nb-config](https://myst-nb.readthedocs.io/en/latest/computation/execute.html#notebook-execution-modes) key found in [conf.py](https://github.com/aiidalab/aiidalab-widgets-base/blob/eeb00a352ed77beadcc38fed66ddfaa5e6d24131/docs/source/conf.py#L30) to `force`.

The issue here is that AiiDA will have to be installed on the documentation server, as some notebooks require access to the AiiDA objects.

unkcpz:
fixes https://github.com/aiidalab/aiidalab-widgets-base/issues/533

Try to bring this in but a bit tricky for the layout and some widgets are timeout in rendering. Need more time to fix one by one.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -16,7 +16,7 @@ sphinx:
     configuration: docs/source/conf.py
     builder: html
     # Let the build fail if there are any warnings
-    fail_on_warning: true
+    #fail_on_warning: true
 
 # Optional but recommended, declare the Python requirements required
 # to build your documentation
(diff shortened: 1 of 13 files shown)


unkcpz:
After release of `aiida-core==2.5.0`, it support temporary sqlite profile, we can use that to import a profile from archive instead of using complex commands to set it up.

danielhollas:
Going to close this since most of the changes in this PR are stale and not applicable anymore.

PR #549, opened 2024-01-22
Set structure traitlet to None before assigning new structure

### LINKED ISSUE #381: Structure importers: cannot import the same structure twice.

danielhollas:
Cherry-picked from #383 by @cpignedoli 

Fixes https://github.com/aiidalab/aiidalab-widgets-base/issues/381

This fixes the setting of self.structure = None when a action button is pressed for an importer ("SMILES" and "import from computer")

Thanks, @danielhollas, I tried self.structure = None just in observe("input_structure") but this does not work still, I keep the modification also in that observe.

TODO(DH): Need to test this


aiidalab_widgets_base/structures.py
```diff
@@ -327,6 +327,7 @@ def _observe_input_structure(self, change):
         # If the `input_structure` trait is set to Atoms object, then the `structure` trait should be set to it as well.
         self.history = []
 
+        self.structure = None
```
danielhollas, line 330: I think all changes in this PR need to be done within the `with self.hold_trait_notifications():` context manager

```suggestion
        with self.hold_trait_notifications():
            self.structure = None
```

danielhollas:
Based on my test and reading of the code this PR is no longer needed

PR #553, opened 2024-02-02
Replace optimade-client with ipyoptimade

unkcpz:
~~Able to install with aiida-core v2.5.0 but turns out there are quite a lot things we need to fix.~~

As discussed (see the thread in this PR), will only replace the optimade-client with `ipyoptimade~=0.1` which is the version right after the migration without the support to the pydantic v2 (which comes in version `v0.2`).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -56,7 +56,7 @@ dev =
     selenium~=4.7.0
     webdriver-manager~=3.8
 optimade =
-    optimade-client==2022.9.19
+    ipyoptimade~=0.1
 smiles =
     rdkit>=2021.09.2
     scikit-learn~=1.0.0
--- a/aiidalab_widgets_base/databases.py
+++ b/aiidalab_widgets_base/databases.py
@@ -157,7 +157,7 @@ class OptimadeQueryWidget(ipw.VBox):
         **kwargs,
     ) -> None:
         try:
-            from optimade_client import default_parameters, query_filter, query_provider
+            from ipyoptimade import default_parameters, query_filter, query_provider
         except ImportError:
             super().__init__(
                 [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ filterwarnings = [
     'ignore:Creating AiiDA configuration:UserWarning:aiida',
     'ignore:crystal system:UserWarning:ase.io.cif',
     'ignore::DeprecationWarning:ase.atoms',
-    # TODO: This comes from a transitive dependency of optimade_client
+    # TODO: This comes from a transitive dependency of ipyoptimade
     # Remove this when this issue is addressed:
     # https://github.com/CasperWA/ipywidgets-extended/issues/85
     'ignore::DeprecationWarning:ipywidgets_extended',


setup.cfg
```diff
@@ -20,7 +20,7 @@ classifiers =
 packages = find:
 install_requires =
     PyCifRW~=4.4
-    aiida-core>=2.1,<3
+    aiida-core>=2.5,<3
```
danielhollas, line 23: We should be very careful about bumping the minimum aiida-core version since it might cause issues for downstream apps. If at all possible, please make the ipyoptimade compatible with earlier aiida-core versions. Ideally, we should allow pip to pick the correct version based on which aiida-core is already installed in the image.

danielhollas (commented):
Thanks for working on this!
I do think we have to workaround the pydantic versions somehow to support earlier aiida-core versions

setup.cfg
unkcpz: Good point, I think very hard and here is some thought.

`ipyoptimade` has nothing to do with `aiida-core` but with `pydantic`.
- `ipyoptimade<0.2.0` which right after I made the migration, it only support `pydantic==v1.0`
- `ipyoptimade==0.2.0` is the version that support pydantic v2.

So if we need something that need `aiida-core==2.5.0` then need `pydantic` v2 is the environment and thus need to use `ipyoptimade==0.2.0`. 
The problem now comes with that the use of optimade widget comes from `aiidalab-widgets-base` and it is another base dependencies in an AiiDAlab environment.

One workaround is not use optimade widget from `aiidalab-widgets-base` but from `ipyoptimade` directly.Then nothing need to change here. 
Maybe another workaround worth to try is what you mentioned, that the `ipyoptimade` is an extra dependency. 

In summary, I think we can not guarantee that one app not break another app. But we need to keep the bottom line that we need to try our best the new version of aiidalab docker-stack will not break the app.

unkcpz: It would be very helpful if you have a clear "break path" like which update of which app or which component would cause the break of which part of an app.

danielhollas: I think our general goal with AWB is to be as independent of various other core dependencies as possible, to avoid coupling between versions of different dependencies. 

>   ipyoptimade<0.2.0 which right after I made the migration, it only support pydantic==v1.0
    ipyoptimade==0.2.0 is the version that support pydantic v2.

This is great that there are diifferent versions that support different pydantic versions! In the best case scenario, we should allow both versions in AWB and let `pip` to pick the correct version. Whether that will work in practice is unclear to me, since pip can ignore the existing aiida-core version constraints. This is another reason why we really should publish AWB as conda package and install it together with `aiida-core` in the base image.

setup.cfg
```diff
@@ -56,7 +56,7 @@ dev =
     selenium~=4.7.0
     webdriver-manager~=3.8
 optimade =
-    optimade-client==2022.9.19
+    ipyoptimade~=0.2
```
danielhollas, line 59: @unkcpz I would first suggest to use version ~=0.1, which should be compatible with current AiiDAlab image. We can then release new AWB version and only then bump it for the next version. WDYT?

unkcpz: Yes, it is a good idea. 
I also think about to add a CI workflow to test on aiida-core==2.1. From AWB point of view, the aiida-core is a very base dependencies and deserve to test on different version. Just like we make python version matrix CIs for the package. 

I am think to 
- drop python 3.8 and test on python 3.9/3.10. 
- test on aiida-core 2.1/2.5 with python 3.9 (since we use py39 for our docker image at the moment). But this require to skip some tests with optimade I guess.

danielhollas: I am not sure pip will always pick the correct version, especially for Apps that don't declare dependency on aiida-core. As first step I think it is safer to do `~0.1.0`, and then test things vigorously. 

Adding an extra CI workflow for aiida-core versions sounds good. Do you mean that you'd pick an image based on the aiida version tag?

danielhollas: Dropping python 3.8 seems fine, although I'd prefer to have that as a separate PR for future reference.

unkcpz: > Do you mean that you'd pick an image based on the aiida version tag?

Both, for the notebook test, pick an image on the aiida version tag. 
Also add a unittest CI workflow with pinning aiida-core version to `~2.1`

danielhollas (approved):
@unkcpz could you test what happens when you bump the aiida-core version in GHA to 2.5.0?

Otherwise I think this is good to go. Since aiida-core <2.5 versions did not depend on pydantic at all, I think this should be safe. I also double checked that no other packages besides pydantic depend on pydantic.

.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
             matrix:
                 browser: [Chrome, Firefox]
                 python-version: ['3.10']
-                aiida-core-version: [2.1.2, 2.4.3]          # test on the latest and the oldest supported version
+                aiida-core-version: [2.1.2, 2.5.0]          # test on the latest and the oldest supported version
```
danielhollas, line 46: We don't have that image yet so this was bound to fail.

danielhollas (commented):
Okay, cool. Good to see that the installation with 2.5 succeeded at least.

.github/workflows/ci.yml
danielhollas: Let's revert back to v2.4.3 and merge this first. I think then we might proceed with releasing the image.

Once this is merged, maybe you could make a alpha pre-release of AWB (2.2.0a0) which we can then use for both testing the image and QeApp (which needs to be updated to specify optimade extra dependency.

unkcpz: Releasing the image of bumping aiida-core to 2.5.0 you mean? But the test will fail since awb is not yet support 2.5.0. 

This is a chicken egg problem. I think we should make a release of awb without changing the aiida-core version in notebook CI test. Make the release and then update the image.

danielhollas: Sorry for not being clear. I agree we can't release the image just yet, and we therefore need to keep the 2.4.3 version the the notebook tests here.  Here are the steps that are needed afaik:

1. Merge this PR
2. Release a pre-release version of AWB
3. Make QeApp work with the pre-released version. Verify that it works in the axisting aiidalab images.
4. Release new version of AWB, v2.2.0
5. Release new version of QeApp
6. Release the new image with aiida v2.5

unkcpz: Thanks for clarification. But I think it will not working for the step 6.

> Release the new image with aiida v2.5

This is impossible if AWB not yet have pydantic v2 supported. 

aiidalab docker stack full stack test:
aiida-core 2.5.0 -> pydantic v2 -> aiidalab-qe (pydantic v2 support) -> AWB (pydantic v2 support)

So if not yet AWB pydantic v2 supported 

For AWB test, it depend on the image with pydantic v2 supported, but that image depend on the this as shown above.

unkcpz: But anyway, before step 6, I think it is all doable. Let's get this merged and when proceed with can see the potential problem more clear and able to find the solution.

danielhollas (approved):

PR #554, opened 2024-02-02
Move optimade dependency as extra dependency

unkcpz:


unkcpz:
Here is the attempt of moving optimade as an extra effort https://github.com/aiidalab/aiidalab-widgets-base/pull/554. I think we want it anyway. 

The test failed expect because the aiida-core is update to `2.5.0`. I am a bit confused on what we really try to compatible with all these versions pinned?

unkcpz:
A real case in my mind is if the user have a workable AiiDAlab environment with `aiida-core==2.4.0`, if they upgrade the `aiidalab-widgets-base` with the pinning of `aiida-core>2.1,<3`, the `aiida-core` will not be updated to `aiida-core==2.5`.

unkcpz:
after #556

danielhollas:
Looking at the `OptimadeQueryWidget`, it seems pretty self-contained and not dependent on other AWB widgets. Therefore, I think it would make sense to push it out of AWB to the new `ipyoptimade` package.
There might be downsides but imo it's worth exploring. 

It would also make it easier for AiiDAlabs Apps to migrate to this package, since it would be more independent of the AWB version (e.g. one could switch to the new `ipyoptimade` widget even with AWB 2.1.0. (unless there are some optimade dependency clashes).

danielhollas:
There's actually a long discussion about whether to include this OPTIMADE widget in AWB on this PR:

https://github.com/aiidalab/aiidalab-widgets-base/pull/131

I think we need to read through the arguments there and decide if they are still valid (especially given that we want to integrate AWB in the Docker image).

tests_notebooks/conftest.py
```diff
@@ -61,7 +61,7 @@ def notebook_service(docker_ip, docker_services, aiidalab_exec):
     aiidalab_exec("chmod -R a+rw /home/jovyan/apps/aiidalab-widgets-base", user="root")
 
     # Install AWB with extra dependencies for SmilesWidget
-    aiidalab_exec("pip install --no-cache-dir .[smiles]")
+    aiidalab_exec("pip install --no-cache-dir .[smiles, optimade]")
```
danielhollas, line 64:
```suggestion
    aiidalab_exec("pip install --no-cache-dir .[smiles,optimade]")
```

unkcpz: Yep, that fixed the tests. Thanks!

danielhollas (commented):
I think this moves us in the right direction so I think this is good to merge now and iterate on it. 

@unkcpz just one request, could you please provide a screenshot of what happens when you try to load QeApp with this version but the `optimade_client` package is not available?

unkcpz:
image

This is how it looks like. I made a change in the last commit to use `<code>optimade-client</code>` to format it.

danielhollas (dismissed):
Thanks. I wonder if more formatting would be warranted (e.g. maybe using the bootstrap error class or something like that to get a reddish box around it. Currently it's not immediately clear that it's an error.

Leaving that up to you.

unkcpz:
I agree, it is not so clear. I change it to 

image

aiidalab_widgets_base/databases.py
```diff
@@ -156,7 +156,22 @@ def __init__(
...
+        try:
+            from optimade_client import default_parameters, query_filter, query_provider
+        except ImportError:
+            super().__init__(
+                [
+                    ipw.HTML(
+                        """
+                        <div style="margin: 10px 0; padding: 10px; background-color: #ffcc00; color: black; border-left: 6px solid #ffeb3b;">
+                            <i class="fa fa-exclamation-triangle" aria-hidden="true"></i>
+                            &nbsp;This widget requires the <code>optimade-client</code> package to be installed.
```
danielhollas, line 168:
```suggestion
                            &nbsp;This widget requires the <code>optimade-client</code> package to be installed.<br>
```
To make the box narrower..
 (Or limit width by CSS)

danielhollas (approved):
Perfect, thanks!

aiidalab_widgets_base/databases.py
unkcpz: Wait... I revert the change, it looks ugly 😸 

image

unkcpz:
I quite like the appearance. I feel comfortable when the whole line filled, so I merge it ^^

PR #555, opened 2024-02-02
Remove deprecated load_documentation_profile for doc build

unkcpz:


danielhollas:
I am confused. Why the does the doc build depend on aiida-core profile?

unkcpz:
Haha, you are right! The profile was for API build, but it is not needed anymore. I remove it and the build still passed.

danielhollas (approved):
Awesome, thanks!

PR #557, opened 2024-02-21
Drop support for py38

unkcpz:
The support for Python versions follows the same timeline as that of numpy, defined in NEP 029. According to that schema, Python 3.8 is no longer support as of April 14 2023 and so official support is also dropped in AWB.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ install_requires =
     nglview~=3.0
     spglib>=1.14,<3
     vapory~=0.1.2
-python_requires = >=3.8
+python_requires = >=3.9
 include_package_data = True
 zip_safe = False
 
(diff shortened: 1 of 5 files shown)


setup.cfg
```diff
@@ -36,7 +36,7 @@ install_requires =
     nglview~=3.0
     spglib>=1.14,<3
     vapory~=0.1.2
-python_requires = >=3.8
+python_requires = >=3.9
```
danielhollas, line 39: Please also update `pyupgrade` configuration in `.pre-commit-config.yaml`.

danielhollas (changes requested):
More importantly, the current Docker stack based on Jupyter docker image is currently Python 3.9, and it was never actually 3.8, so this is a safe deprecation. It would only affect people that are still on the old old docker stack, and those should really just migrate to the new one.

Thanks @unkcpz !

danielhollas (approved):

PR #558, opened 2024-02-21
Add test to latest and oldest supported aiida-core

unkcpz:


unkcpz:
The unittest with python==3.9 and aiida-core==2.1.2 failed because of PyYAML issue https://github.com/yaml/pyyaml/issues/723

So in fact the oldest support version is `2.2.0` where the PyYAML bumped to `~6.0`, https://github.com/aiidateam/aiida-core/commit/09adc188e46d5d2265a921f6726d94d19a72d93b

danielhollas:
Ugh, that's annoying. I'll take a look, thanks!

danielhollas:
I checked on Dockerhub and we released an image with aiida 2.2 10 months ago. So let's just test with this version and not waste too much time on this. Can always  improve this in the future if needed.

.github/workflows/ci.yml
```diff
@@ -109,6 +109,7 @@ jobs:
             matrix:
                 tag: [latest]
                 python-version: ['3.9', '3.10']
+                aiida-core-version: [2.1.2, 2.4.3]          # test on the latest and the oldest supported version
```
danielhollas, line 112:
```suggestion
                aiida-core-version: [2.2.2, 2.4.3]          # test on the latest and the oldest supported version
```

danielhollas (changes requested):

tests_notebooks/docker-compose.yml
```diff
@@ -4,7 +4,13 @@ version: '3.4'
...
 
     aiidalab:
-        image: ghcr.io/aiidalab/full-stack:${TAG:-latest}
+        # TODO: after the oldest aiida-core supported version is v2.4.0,
+        # move to use ghcr.io registry to pull the image, the dockerhub registry has
+        # a rate limit of 100 pulls per 6 hours.
+        # aiida-core < 2.4.0 is not push to ghcr.io because of CI issues
+        # The CI is fixed in https://github.com/aiidalab/aiidalab-docker-stack/pull/390 and
+        # from aiida-core 2.4.0 the images are pushed to ghcr.io.
+        image: aiidalab/full-stack:${TAG:-latest}
```
danielhollas, line 13: Thank you for the  comment, very good for future reference

```suggestion
        # TODO: When we drop support of images with aiida-core<2.4.0, we should go back to using
        # the ghcr.io registry to pull the image,
        # since the dockerhub registry has a rate limit of 100 pulls per 6 hours.
        # Images with aiida-core<2.4.0 were not pushed to ghcr.io because of CI issues,
        # which were fixed in https://github.com/aiidalab/aiidalab-docker-stack/pull/390
        image: aiidalab/full-stack:${TAG:-latest}
```

.github/workflows/ci.yml
```diff
@@ -109,6 +109,7 @@ jobs:
             matrix:
                 tag: [latest]
```
danielhollas, line 110:
```suggestion
```
The tag variable is not used in the `test-package1` workflow.

.github/workflows/ci.yml
```diff
@@ -41,9 +41,9 @@ jobs:
 
         strategy:
             matrix:
-                tag: [latest]
                 browser: [Chrome, Firefox]
                 python-version: ['3.10']
+                aiida-core-version: [2.1.2, 2.4.3]          # test on the latest and the oldest supported version
```
danielhollas, line 46: Here version `2.1.2` actually works since it is pre-installed in the image. So we have at least some coverage.

danielhollas (approved):
Tests are passing now. :tada:  Very happy with this change, will make us more confident to not break backwards compatibility. Thanks @unkcpz!

Couple small fixes but otherwise good to go.

unkcpz:
Thanks for the review @danielhollas

PR #562, opened 2024-02-27
Remove obsolete files

danielhollas:
As it says on the tin --- none of these files seem to be used anymore.


unkcpz:
I agree. But no idea what those files were for, better to have @yakutovicha double check on it.

__init__.py
yakutovicha, line 1: This one was used in the infancy of AiiDAlab to make an app importable. Not needed anymore.

install.sh
yakutovicha, line 1: This is from the old days when the jupyter extensions had to be activated after an install. In this particular case, we don't even use the external `fileupload` widget anymore

miscellaneous/images/computer.png
yakutovicha, line 1: The computer setup widget that used this image was removed [here](https://github.com/aiidalab/aiidalab-widgets-base/pull/344/files#diff-41129b0f1db994ae59e278b44419d6edd35ba16df9ae0b4123634e40ba1de293).

eln_import.ipynb
yakutovicha, line 1: An old link to keep backwards compatibility.

yakutovicha (approved):
Added some history in the comments

PR #564, opened 2024-02-27
Remove the push and tag capability of bumpver

unkcpz:
The push and tag can be triggered manually by hand. 
The https://github.com/aiidalab/aiidalab-widgets-base/blob/master/.github/workflows/publish.yml will manage the publish to PyPI.


danielhollas:
I've converted to draft until we settle on a solution. For reference, here's the discussion on Discourse https://aiida.discourse.group/t/standard-way-of-making-release/278/7

PR #566, opened 2024-02-27
Use Ruff

danielhollas:
Besides flake8 plugins that we already had, I've added ruff-specific rules (`RUF`). Especially the one that detects unnecessary list comprehension is quite useful and made a couple of places more readable.

I've disabled a couple of rules that would take more work to resolve all.


.pre-commit-config.yaml
```diff
@@ -12,49 +12,24 @@ repos:
...
-      rev: 6.1.0
-      hooks:
-          - id: flake8
-            args: [--count, --show-source, --statistics]
-            additional_dependencies:
-                - flake8-bugbear==23.2.13
-                - flake8-builtins==2.1.0
-                - flake8-comprehensions==3.10.1
-                - flake8-debugger==4.1.2
-                - flake8-logging-format==0.9.0
```
danielhollas, line 36: I did not activate these two in ruff. We do not use logging in AWB and the flake8-debugger seems rather niche.

aiidalab_widgets_base/databases.py
```diff
@@ -152,8 +152,8 @@ class :class:`aiidalab_widgets_base.structures.StructureManagerWidget`.
 
     def __init__(
         self,
-        embedded: bool = True,
-        title: str = None,
```
danielhollas, line 156: This type was not correct (missing Optional). We're currently not doing any actual typechecking so it's better to just remove it, wrong types are worse than no types.

aiidalab_widgets_base/viewers.py
```diff
@@ -1049,7 +1049,7 @@ def apply_displayed_selection(self, _=None):
         else:
             self.wrong_syntax.layout.visibility = "visible"
 
-    def download(self, change=None):  # pylint: disable=unused-argument
+    def download(self, _change=None):
```
unkcpz, line 1052: I remember in other place this is just put as `_`.

aiidalab_widgets_base/export.py
```diff
@@ -40,8 +40,6 @@ def export_aiida_subgraph(self, change=None):  # pylint: disable=unused-argument
             document.body.appendChild(link);
             link.click();
             document.body.removeChild(link);
-            """.format(
-                payload=payload, filename=f"export_{self.process.id}.aiida"
-            )
+            """.format(payload=payload, filename=f"export_{self.process.id}.aiida")
```
unkcpz, line 43: I think this can be a f-string

```python
f"""
var link = document.createElement('a');
link.href = "data:;base64,{payload}";
link.download = "export_{self.process.id}.aiida";
document.body.appendChild(link);
link.click();
document.body.removeChild(link);
"""
```

unkcpz (changes requested):
Thanks @danielhollas. Just to minor requests.

aiidalab_widgets_base/viewers.py
danielhollas: Done.

aiidalab_widgets_base/export.py
danielhollas: It can, but the current `.format` is more readable imo. Because we're interpolating a string that represents Javascript code, it would be quite hard to see which variables and where are they inserted.

PR #568, opened 2024-02-27
New publish workflow

danielhollas:
The biggest fix here is to stop a release to PyPI if the tag does not start with `v`.

We now also automatically skip the publish if we are in a fork (it would fail anyway since the secrets are not available on forks).
I've verified the new behaviour on my fork.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -47,9 +47,7 @@ jobs:
     publish-test:
 
         name: Build and publish on TestPyPI
-        if: >
-            startsWith(github.ref, 'refs/heads/release/') ||
-            startsWith(github.ref, 'refs/tags/')
+        if: startsWith(github.ref, 'refs/heads/release/')
 
         needs: [build]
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -94,16 +91,15 @@ jobs:
...
             - uses: softprops/action-gh-release@v0.1.15
               name: Create release
-              if: startsWith(github.ref, 'refs/tags/v')
               with:
                   files: |
                       dist/*
                   generate_release_notes: true
-
-            - name: Publish distribution on PyPI
-              uses: pypa/gh-action-pypi-publish@release/v1
```
danielhollas, line 106: It certainly looks like we would publish to PyPI even if the tag did not start with `v`. I think the new code is correct. Also I've switched the order of PyPI publish and GitHub release. This was, if the PyPI release fails for any reason, we do not create a release on Github.

unkcpz (approved):

PR #572, opened 2024-03-01
Only trigger publish on release.published event

danielhollas:
We noticed that making a release triggered multiple different `release` events: `created`, `prereleased`, `published`. So we need to tweak the workflow file so that the workflow only run for the `release.publish` event.

https://github.com/aiidalab/aiidalab-widgets-base/actions?query=event%3Arelease

Per GHA docs: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

EDIT: ~~I am not sure if the fix here works, need to test on another repo. Marking as draft for now~~

EDIT2: Never mind, turns out I can't read the docs properly. I fixed the syntax and checked on my repo, this seems to work correctly now.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -9,7 +9,7 @@ on:
         # before it is pushed to the actual PyPI index.
             - release/**
     release:
-        published:
+        types: [published]
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


unkcpz (approved):
Looks all good, Thanks! @danielhollas
Now we have everything ready, can you write a wiki page for how to release. We can then ask Sasha to confirm with it and we can implement it to other packages as well.

danielhollas:
Yep! On my todo-list for today or Wednesday. Thanks!

PR #576, opened 2024-03-05
mimic the behavior of nglview and update the cell and positions

### LINKED ISSUE #575: Bonds are wrong
From @cpignedoli .
Steps to reproduce. In QE App, select the “Silicon (diamond)” example structure, and click on “convert to primitive cell”. At every click, it completely changes the bonds, and they are wrong: see two screenshots below.
Screenshot from 2024-03-05 12-32-07

superstar54:
Fix #575 

After inspecting the source code of nglview, I found that nglview loads the structure through saves and loads a temp "pdb" file. This process could change the cell and positions of the structure.

This cause problems
- The structure shown in the viewer is rotated (both cell and position); this is not obvious because the XYZ axis is not shown, so the user will not know the structure is rotated. **This will not affect the structure used for the real calculation.**
- We use the original structure to calculate the bond's positions. Thus, there are mismatches between the bond and atoms.

This PR fixes the second problem. I transferred the `displayed_structure` into a standard form, with the `a` axes aligned with the X axis, so that the `displayer_structure` will not be rotated anymore when saving to a `pdb` file. Thus the viewer will show the correct `displayed_structure.` In this case, the calculated bonds will match the atoms.


danielhollas:
> After inspecting the source code of nglview, I found that nglview loads the structure through saves and loads a temp "pdb" file.

Ooof, that's horrifying. Can you please clarify when is this happening? Is it happening for any kind of structure (e.g. even without a cell?)? Or only for in certain cases?

superstar54:
> Ooof, that's horrifying. Can you please clarify when is this happening? Is it happening for any kind of structure (e.g. even without a cell?)? Or only for in certain cases?

The `ASEStructure` use it, which, unfortunately, means it is used for all structures inside AiiDAlab
https://github.com/nglviewer/nglview/blob/12a75bc6f0cea8fbc0f75ffc68117d010ce96cc9/nglview/adaptor.py#L137

load structure into nglview:
https://github.com/nglviewer/nglview/blob/12a75bc6f0cea8fbc0f75ffc68117d010ce96cc9/nglview/widget.py#L1241

danielhollas:
Was this a problem in version 2.0, or has this been introduced with the new features in 2.1. This workaround seems really icky. 

> The ASEStructure uses it, which, unfortunately, means it is used for all structures inside AiiDAlab

If that is the case, why is this fix only needed when the supercell changes?

superstar54:
> Was this a problem in version 2.0, or has this been introduced with the new features in 2.1.

This has existed from the very beginning (before awb 1.0). But it is not a problem, because user will not notice if both cell and position are rotated. And we are not exporting the rotated structure from the view. However, it becomes a problem when we draw bonds manually #535 .

danielhollas:
> This is the only entry point where we set the structure in the viewer. 

Isn't that rather the `structure` traitlet? Confusing indeed, but okay for now. 

TBH I am still not very happy with the solution here....We're writing a file every time a structure is updated, that's not going to help performance. Is there a way to somehow get the correct coordinates from the nglview (even if we needed to access its internal state), and pass them around to whenever we compute the bonds? That seems like a cleaner design and makes more sense in terms of data flow.

superstar54:
> Isn't that rather the `structure` traitlet? Confusing indeed, but okay for now.

The `displayed_structure` is used in the nglview.
https://github.com/aiidalab/aiidalab-widgets-base/blob/8190dc08ab9a25bae84a991cad165a1ff3d05d1d/aiidalab_widgets_base/viewers.py#L1197

> TBH I am still not very happy with the solution here....We're writing a file every time a structure is updated, that's not going to help performance. Is there a way to somehow get the correct coordinates from the nglview (even if we needed to access its internal state), and pass them around to whenever we compute the bonds?

Good point. I didn't find a way to get data from nglview. But I manager to get the correct cell and positions by rotating the cell into a standard form.

aiidalab_widgets_base/viewers.py
```diff
@@ -1136,7 +1136,14 @@ def _observe_supercell(self, _=None):
...
             )  # To make sure the structure is always updated.
-            self.set_trait("displayed_structure", self.structure.repeat(self.supercell))
+            # because nglview loads the structure through save and load a temp "pdb" file
+            # thus change the cell and positions into standard form.
+            # we need to mimic this process and update the cell and positions
+            pdb_structure = self.structure.copy()
+            pdb_structure.set_cell(
+                self.structure.cell.standard_form()[0], scale_atoms=True
+            )
+            self.set_trait("displayed_structure", pdb_structure.repeat(self.supercell))
```
danielhollas, line 1146:
```suggestion
            # nglview converts the ASE structure internally
            # by saving and loading from temporary "pdb" file.
            # In this process, the cell and positions are converted to the standard form.
            # Because we use the `displayed_structure` trait outside of nglview
            # (e.g. to calculate and display custom bonds),
            # we need to do the same conversion here as well.
            oriented_structure = self.structure.copy()
            oriented_structure.set_cell(
                self.structure.cell.standard_form()[0], scale_atoms=True
            )
            self.set_trait("displayed_structure", oriented_structure.repeat(self.supercell))
```

aiidalab_widgets_base/viewers.py
```diff
@@ -1136,7 +1136,14 @@ def _observe_supercell(self, _=None):
             self.set_trait(
                 "displayed_structure", None
             )  # To make sure the structure is always updated.
-            self.set_trait("displayed_structure", self.structure.repeat(self.supercell))
+            # because nglview loads the structure through save and load a temp "pdb" file
+            # thus change the cell and positions into standard form.
```
danielhollas, line 1140: Would it possible to add a sentence what the "standard form" actually is?
(I know this is probably basic knowledge, but not for people outside of solid state).

danielhollas (changes requested):
@superstar54 Thanks! This looks much better. Just suggesting some re-wording.

@cpignedoli could you test if this branch fixes the issue you saw?

aiidalab_widgets_base/viewers.py
cpignedoli: dear @superstar54 @danielhollas I tested and to me seems working.
I am a little bit scared of converting back and forth from pdb in case coordinates could lose accuracy but the tests I did seem to have no accuracy problems.

I wanted to update the files of the examples but I will do it on a separate PR as soon as I manage

danielhollas:
> I am a little bit scared of converting back and forth from pdb in case coordinates could lose accuracy but the tests I did seem to have no accuracy problems.

I agree that it is extremely fishy, but note that this conversion only happens inside of nglview, so it only affects what you see on the screen, not the data that we have internally in aiidalab widgets. 

> I wanted to update the files of the examples but I will do it on a separate PR as soon as I manage

Why do you want to update the examples?

cpignedoli:
> Why do you want to update the examples?
We are trying to have a demo server as clean as possible, starting from examples where lattice parameters are all, for example, exp ones, the cell is primitive with symmetric axes settings, the name is clear (e.g. alpha/beta-quartz in place of SiO2)

we have a preliminary list of issues we would need to fix, I will share with you a google document

>it only affects what you see on the screen, not the data that we have internally in aiidalab widgets.

indeed all tests I did show no issues, if I download a structure after doing back and forth from primitive to conventional, the coordinates are translated+rotated,wrt teh initial input geo, to the very last digit

aiidalab_widgets_base/viewers.py
superstar54: I added the description on "standard form" and refined the comment for clarity.

superstar54:
Hi @danielhollas , thanks for the review! I made the changes as you suggested. Could you please take another look?

aiidalab_widgets_base/viewers.py
```diff
@@ -1136,7 +1136,16 @@ def _observe_supercell(self, _=None):
...
+            # nglview displays structures by first saving them to a temporary "pdb" file, which necessitates
+            # converting the unit cell and atomic positions into a standard form where the a-axis aligns along the x-axis.
+            # This transformation can cause discrepancies between the atom positions and custom bonds calculated from the original structure.
+            # To mitigate this, we transform the "displayed_structure" into the standard form prior to rendering in nglview.
+            # This ensures that nglview's internal handling does not further modify the structure unexpectedly.
+            pdb_structure = self.structure.copy()
+            pdb_structure.set_cell(
+                self.structure.cell.standard_form()[0], scale_atoms=True
+            )
+            self.set_trait("displayed_structure", pdb_structure.repeat(self.supercell))
```
danielhollas, line 1148:
```suggestion
            oriented_structure = self.structure.copy()
            oriented_structure.set_cell(
                self.structure.cell.standard_form()[0], scale_atoms=True
            )
            self.set_trait("displayed_structure", oriented_structure.repeat(self.supercell))
```

danielhollas (approved):
Just one variable naming suggestion, otherwise looks good. Thanks for updating the comment!

aiidalab_widgets_base/viewers.py
superstar54: thanks! I change it to `standard_structure`.

danielhollas:
@superstar54 this feels like an important bugfix and seems like the issue is present in version 2.1.0. Should we backport it to 2.1.x branch? And release a bugfix version?

superstar54:
Hi @danielhollas , I'm not very familiar with release management, but if you believe it is good to do this, please go ahead with the necessary steps. Thanks for handling this!

PR #578, opened 2024-03-17
CI: Use uv installer instead of pip (take 2)

danielhollas:
[uv](https://github.com/astral-sh/uv/issues/2155l) is the new cool kid in the town of Python packaging, from the creators of ruff. It currently serves as a (much) faster drop-in replacement for pip. 

For AWB, uv _without_ cache runs in ~10s, compared to 1m10s for pip **with** cache. :exploding_head: So overall speedup around 30% for our workflows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
               # the pytest-docker dependency.
               # To add the dependencies to the container, we need to add them to the
               # installation command in the `test_notebooks/conftest.py`.
-              run: uv pip install --system "aiidalab_widgets_base[dev] @ ."
+              run: uv pip install --system -e .[dev]
 
             - name: Set jupyter token env
               run: echo "JUPYTER_TOKEN=$(openssl rand -hex 32)" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -105,14 +103,18 @@ jobs:
...
 
             - name: Install package
-              run: pip install -e .[dev,smiles,optimade] aiida-core==${{ matrix.aiida-core-version }}
+            # NOTE: uv (unlike pip) does not compile python to bytecode after install.
+            # This uncovered a lot of SyntaxError(s) in the vapory package,
+            # since pip swallows these by default (WTH?).
+            # Ideally, these would be fixed, but vapory is largely unmaintained,
+            # so here we simply keep the pip behaviour with the --compile flag.
+            # See https://github.com/astral-sh/uv/issues/1928#issuecomment-1968857514
+              run: uv pip install --compile --system "aiidalab_widgets_base[dev,smiles,optimade] @ ." aiida-core==${{ matrix.aiida-core-version }}
```
danielhollas, line 117: Note: `uv` currently does not allow the simpler syntax ".[dev,smiles,optimade]`. But it will in the future.

unkcpz: I guess beside this, all other stuff will be the same? It is not another package manage but simply an installer I assume? Which means the installed packages can be inspected by the normal pip command, correct?

danielhollas: Yes, `uv` is only an installer for now, for the most part you simply substitute `pip ...` with `uv pip ...`. There are difference in behaviour, as explained in `uv`s README, and not everything is implemented yet, but it should be stable enough for the CI to use (and we pin the version).

danielhollas: >  Note: uv currently does not allow the simpler syntax .[dev,smiles,optimade]. But it will in the future.

This simpler syntax is actually allowed for editable installs, so I switched to editable install.

unkcpz (approved):
Looks all good, thanks! Nice stuff @danielhollas

PR #579, opened 2024-03-21
Fix download of binary files in FolderDataViewer

### LINKED ISSUE #350: FolderDataViewer chokes on binary files
I was trying to browse the content if the Retrieved folder of an ORCA calculation in the NodeTreeWidget, but when I tried to open and download a binary wavefunction file, I got a Unicode decoding error. 

image

Here's part of the stack trace
```python
~/apps/aiidalab-widgets-base/aiidalab_widgets_base/viewers.py in change_file_view(self, change)
   1057     def change_file_view(self, change=None):  # pylint: disable=unused-argument
   1058         with self._folder.open(self.files.value) as fobj:
-> 1059             self.text.value = fobj.read()
   1060 
   1061     def download(self, change=None):  # pylint: disable=unused-argument

/opt/conda/lib/python3.7/codecs.py in decode(self, input, final)
    320         # decode input (taking the buffer into account)
    321         data = self.buffer + input
--> 322         (result, consumed) = self._buffer_decode(data, self.errors, final)
    323         # keep undecoded input until the next call
    324         self.buffer = data[consumed:]

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf8 in position 8: invalid start byte
```

And here's the stack trace if I try to download the file

```python
UnicodeDecodeError                        Traceback (most recent call last)
~/apps/aiidalab-widgets-base/aiidalab_widgets_base/viewers.py in download(self, change)
   1064 
   1065         payload = base64.b64encode(
-> 1066             self._folder.get_object_content(self.files.value).encode()
   1067         ).decode()
   1068         javas = Javascript(

~/.local/lib/python3.7/site-packages/aiida/orm/nodes/node.py in get_object_content(self, path, mode, key)
    545             warnings.warn("from v2.0 only the modes 'r' and 'rb' will be accepted", AiidaDeprecationWarning)  # pylint: disable=no-member
    546 
--> 547         return self._repository.get_object_content(path, mode)
    548 
    549     def put_object_from_tree(

~/.local/lib/python3.7/site-packages/aiida/orm/utils/_repository.py in get_object_content(self, key, mode)
    136         """
    137         with self.open(key, mode=mode) as handle:
--> 138             return handle.read()
    139 
    140     def put_object_from_tree(self, path, key=None, contents_only=True, force=False):

/opt/conda/lib/python3.7/codecs.py in decode(self, input, final)
    320         # decode input (taking the buffer into account)
    321         data = self.buffer + input
--> 322         (result, consumed) = self._buffer_decode(data, self.errors, final)
    323         # keep undecoded input until the next call
    324         self.buffer = data[consumed:]

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf8 in position 8: invalid start byte
```

We need to handle these errors and do the appropriate thing for binary filles.

danielhollas:
Retrieved folder of CalcJobs may contain binary files. FolderDataViewer would except if the user tried to display / download such file. 

Fixes #350

### DIFF
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -1521,20 +1521,25 @@ class FolderDataViewer(ipw.VBox):
         super().__init__(children, **kwargs)
 
     def change_file_view(self, _=None):
-        with self._folder.base.repository.open(self.files.value) as fobj:
-            self.text.value = fobj.read()
+        try:
+            with self._folder.base.repository.open(self.files.value) as fobj:
+                self.text.value = fobj.read()
+        except UnicodeDecodeError:
+            self.text.value = "[Binary file, preview not available]"
 
     def download(self, _=None):
-        """Prepare for downloading."""
+        """Download selected file."""
         from IPython.display import Javascript
 
-        payload = base64.b64encode(
-            self._folder.get_object_content(self.files.value).encode()
-        ).decode()
+        # TODO: Preparing large files for download might take a while.
+        # Can we do a streaming solution?
+        raw_bytes = self._folder.get_object_content(self.files.value, "rb")
+        base64_payload = base64.b64encode(raw_bytes).decode()
+
         javas = Javascript(
             f"""
             var link = document.createElement('a');
-            link.href = "data:;base64,{payload}"
+            link.href = "data:;base64,{base64_payload}"
             link.download = "{self.files.value}"
             document.body.appendChild(link);
             link.click();
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -264,12 +264,15 @@ def folder_data_object():
     """Return a `FolderData` object."""
     FolderData = plugins.DataFactory("core.folder")  # noqa: N806
     folder_data = FolderData()
-    with io.StringIO("content of test1 filelike") as fobj:
+    with io.StringIO("content of test1.txt") as fobj:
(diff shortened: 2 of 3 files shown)


danielhollas:
@sphuber would mind taking a quick look at this one?

Essentially, we're trying to download files from the file repository via the FolderData interface, without knowing upfront whether it is a binary or text file. I am wondering if there is a better approach than I took here.

sphuber:
> @sphuber would mind taking a quick look at this one?
> 
> Essentially, we're trying to download files from the file repository via the FolderData interface, without knowing upfront whether it is a binary or text file. I am wondering if there is a better approach than I took here.

Seems reasonable to me. AiiDA doesn't store any information about the encoding in the repository. It just works with byte streams. It is up to the client to know and how to deal with it.

aiidalab_widgets_base/viewers.py
```diff
@@ -1521,20 +1521,25 @@ def __init__(self, folder, downloadable=True, **kwargs):
...
     def download(self, _=None):
-        """Prepare for downloading."""
+        """Download selected file."""
         from IPython.display import Javascript
 
-        payload = base64.b64encode(
-            self._folder.get_object_content(self.files.value).encode()
-        ).decode()
+        # TODO: Preparing large files for download might take a while.
+        # Can we do a streaming solution?
```
superstar54, line 1535: Could you open an issue on this?

superstar54 (approved):
LGTM, thanks for the work!

PR #582, opened 2024-04-09
DO NOT MERGE: Test new aiida pytest fixtures

danielhollas:


danielhollas:
I'll close this for now since we'll be unable to use this while we are testing against older aiida-core version, sadly.

PR #583, opened 2024-04-10
Build NodeTree on demand

superstar54:
Fix https://github.com/aiidalab/aiidalab-widgets-base/issues/584

This PR introduces the idea that: only build NodeTree on demand.
- do not build tree recursively, only show the top-level
- build the sub-tree when users select it, and expand it.

## Demo
Here I used the result of the IR/Raman plugin (from @AndresOrtegaGuerrero ) as a example. Before, it's almost impossible to load it in the NodeTree.

Using this PR:

https://github.com/aiidalab/aiidalab-widgets-base/assets/11457659/18aa487e-a6f6-429d-88b2-0b66ac9337ef

### DIFF
--- a/aiidalab_widgets_base/nodes.py
+++ b/aiidalab_widgets_base/nodes.py
@@ -189,9 +189,19 @@ class NodesTreeWidget(ipw.Output):
                 name = calc_info(node)
             else:
                 name = str(node)
-        return cls.NODE_TYPE.get(type(node), UnknownTypeTreeNode)(
+        tree_node = cls.NODE_TYPE.get(type(node), UnknownTypeTreeNode)(
             pk=node.pk, name=name, **kwargs
         )
+        # Set the style based on the process state of the node
+        if isinstance(node, orm.ProcessNode):
+            process_state = (
+                engine.ProcessState.EXCEPTED if node.is_failed else node.process_state
+            )
+            tree_node.icon_style = cls.PROCESS_STATE_STYLE.get(
+                process_state, cls.PROCESS_STATE_STYLE_DEFAULT
+            )
+
+        return tree_node
 
     @classmethod
     def _find_called(cls, root):
--- a/aiidalab_widgets_base/nodes.py
+++ b/aiidalab_widgets_base/nodes.py
@@ -141,6 +141,10 @@ class NodesTreeWidget(ipw.Output):
             display(self._tree)
 
     def _observe_tree_selected_nodes(self, change):
+        for node in change["new"]:
+            if hasattr(node, "pk"):
+                # find the selected node and build the tree from it, so that users can expand and explore the tree
+                self._build_tree(self.find_node(node.pk))
         return self.set_trait(
             "selected_nodes",
             tuple(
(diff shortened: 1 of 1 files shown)


danielhollas:
@superstar54 this is amazing, thanks so much on working on this. :clap: I badly need it for my app as well so I will definitely take a look and test it with my app, hopefully tomorrow.

It's nice to see that there was not more code needed for this; I was a bit afraid that it will be more difficult to implement. 
Did you notice any downsides / weird corner cases in your implementation?

giovannipizzi:
That's fantastic, great! If Daniel checks it it would be great and then please merge it! Pinging @cpignedoli as well.

As a note (on existing code, not what you changed): is `_default_openend` a typo for `_default_opened` or it's correct?

aiidalab_widgets_base/nodes.py
```diff
@@ -141,6 +141,15 @@ def _refresh_output(self):
...
     def _observe_tree_selected_nodes(self, change):
+        for node in change["new"]:
+            # find the selected node and build the tree from it, so that users can expand and explore the tree
+            node_pk = (
+                node.parent_pk
+                if isinstance(node, AiidaOutputsTreeNode)
+                else getattr(node, "pk", None)
+            )
+
+            self._build_tree(self.find_node(node_pk, getattr(node, "namespaces", None)))
```
danielhollas, line 152: From just looking at this code, I wonder how is it ensured that the tree is build only once? i.e. if a user selects / deselects the same node multiple times, is it going to trigger the build_tree multiple times or will the existing nodes be reused somehow?

superstar54: Thanks for the review. It's designed to build the tree every time the user selects it, for two reasons:
- if the job is running, it will ensure the user sees the latest node tree.
- since it only builds the first level of the child nodes, it should be quick even for a large number of child nodes.

superstar54:
> As a note (on existing code, not what you changed): is _default_openend a typo for _default_opened or it's correct?

:+1: It's a typo. I fixed it.

AndresOrtegaGuerrero:
@superstar54 I noticed the change in color of the WorkChains with finished[0] from green to gray, 
image
Technically it doesnt affect the performance, or the visualization of the elements of the three but i was just wondering if it should be address or ignored ?

cpignedoli:
@superstar54 this is great. I tested on my docker for something small just to confirm it works. In addiiton, I checked with @AndresOrtegaGuerrero on a large Raman case, this PR seems to solve our problems.

superstar54:
> I noticed the change in color of the WorkChains with finished[0] from green to gray,

Hi @AndresOrtegaGuerrero , thanks for pointing out this. it should be fixed in [795af3f](https://github.com/aiidalab/aiidalab-widgets-base/pull/583/commits/795af3f69348d0ee403b05676c8f20620f890ca6).

AndresOrtegaGuerrero:
> > I noticed the change in color of the WorkChains with finished[0] from green to gray,
> 
> Hi @AndresOrtegaGuerrero , thanks for pointing out this. it should be fixed in [795af3f](https://github.com/aiidalab/aiidalab-widgets-base/pull/583/commits/795af3f69348d0ee403b05676c8f20620f890ca6).

Indeed now is fixed!, thank is great job

AndresOrtegaGuerrero (approved):
LGTM! , I did different tests in qe app and improve performance immensely!. Great Job.

aiidalab_widgets_base/nodes.py
superstar54: Hi @danielhollas , the already loaded child nodes will be reused when running the `_build_tree`.

PR #585, opened 2024-04-10
Bump version v2.2.0a0 -> v2.2.0a1

danielhollas:
I want to get a fix in #350 in a new alpha version.

Note, I've specifically opened this PR from my fork, to test @unkcpz observation that the tag from fork will not get transferred to origin when the PR is merged (which is most likely what will happen). In that case, I'll create the tag manually after merge.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -67,7 +67,7 @@ docs =
     myst-nb
 
 [bumpver]
-current_version = "v2.2.0a0"
+current_version = "v2.2.0a1"
 version_pattern = "vMAJOR.MINOR.PATCH[PYTAGNUM]"
 commit_message = "Bump version {old_version} -> {new_version}"
 commit = True
--- a/aiidalab_widgets_base/__init__.py
+++ b/aiidalab_widgets_base/__init__.py
@@ -117,4 +117,4 @@ __all__ = [
     "viewer",
 ]
 
-__version__ = "2.2.0a0"
+__version__ = "2.2.0a1"


superstar54 (approved):
LGTM!

danielhollas:
Cheers @superstar54 

Indeed, the tag was not transferred from the forked repo, I've created it manually when making a release. It's nice that you can specify a specific commit for the tag.

image

setup.cfg
```diff
@@ -67,7 +67,7 @@ docs =
     myst-nb
 
 [bumpver]
-current_version = "v2.2.0a0"
+current_version = "v2.2.0a1"
 version_pattern = "vMAJOR.MINOR.PATCH[PYTAGNUM]"
 commit_message = "Bump version {old_version} -> {new_version}"
 commit = True
```
unkcpz, line 73: If the tag from the fork is not sync to the repo, then it is meaningless to tag to the forked repo, we need to remove `tag = True`.

unkcpz:
> It's nice that you can specify a specific commit for the tag.

Now I am confusing again. If the tag has to be create manually, it entirely fallback to the strategy that create the tag to trigger the deployment (upload to PyPI). So in the end, I think we should use create tag to make the release (I think it is also quite standard, no?) What we learned and can added is using this CI review and approve, I don't remember where it is implemented and I am confused why I didn't see it now.

unkcpz:
> pushing the release branch to origin allows you to publish the release on TestPyPI first, where you can verify the package before the actual release.

This is a good point and I totally agree. So we can either always ask to run `bumpver` from a newly cloned folder or add a option to bumpver to push to a certain remote. Let's discuss in person.

PR #587, opened 2024-04-24
Add openbis connection

yakutovicha:


danielhollas:
Eln functionality has been removed from awb so closing

PR #589, opened 2024-04-28
Add pandas as direct dependency

### LINKED ISSUE #588: Add pandas as a dependency
We use pandas to produce HTML tables in DictViewer, but we do not have it declared as a dependency!
This worked in the past since we brought pandas transitively via optimade widget. But now that OPTIMADE is optional, we definitely need to declare it explicitly.

This is a blocker for 2.2 release so I'll do it soon. 

CC @superstar54

danielhollas:
pandas is used in several widgets to generate HTML tables. While the ultimate goal is to replace it with a lighterweight dependency, for now we declare it as direct dependecy. Previously, it was defined only transitively via `optimade_client`, which is now turned into optional `optimade` extras.

Closes #588

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,6 +36,7 @@ install_requires =
     nglview~=3.0
     spglib>=1.14,<3
     vapory~=0.1.2
+    pandas~=2.1
 python_requires = >=3.9
 include_package_data = True
 zip_safe = False


superstar54 (approved):
Thanks for the fix!

PR #592, opened 2024-04-29
Do not add auxiliary cell to uploaded/generated gas phase structures 

danielhollas:
`StructureUpload` and `SmilesWidget` currently automatically add a 10 Angstrom auxiliary cells to gas phase molecules. That's needed for plane wave codes, but is confusing and distracting for Quantum Chemistry codes such as ORCA.

This PR introduces a new parameter (`add_auxiliary_cell`) to these widgets to control this behaviour. The default is kept `True` for backwards compatibility (even though as a chemist (not material scientist) I am not happy about it :-D )

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -375,9 +375,14 @@ class StructureUploadWidget(ipw.VBox):
     )
 
     def __init__(
-        self, title="", description="Upload Structure", allow_trajectories=False
+        self,
+        title="",
+        description="Upload Structure",
+        allow_trajectories=False,
+        add_auxiliary_cell=True,
     ):
         self.title = title
+        self.add_auxiliary_cell = add_auxiliary_cell
         self.file_upload = ipw.FileUpload(
             description=description, multiple=False, layout={"width": "initial"}
         )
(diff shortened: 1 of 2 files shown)


danielhollas:
@yakutovicha @cpignedoli can you have a brief look to see if you're okay with this PR? Since it is a (technically) breaking change, I'd like to get this in for v3 of AWB.

I plan to add some unit tests here but would like to get a green light from you first. Thanks!

cpignedoli:
Hi @danielhollas, I tested it, and it works. I don't have an issue with the principle, but if we leave it as is, it will create problems for all users currently using the QE app and the Empa apps for molecules.

I would suggest triggering a warning for systems without a cell, and possibly adding a button to automatically create a cell using the logic that was enforced before.

What do you think?

danielhollas:
@cpignedoli thanks. In that case, I think a better approach is to introduce an extra parameter in the widget initialized (e.g. `add_auxiliary_cell`) and let the application choose which behaviour it desires. For me personally, it would make sense to have this off by default, but as you said that would mean that EMPA apps and QeApps would have to be modified so I am also fine to keep the current behaviour as default and modify AtmoSpec instead.

Would that work? (please see the updated code).

aiidalab_widgets_base/structures.py
```diff
@@ -375,9 +375,14 @@ class StructureUploadWidget(ipw.VBox):
     )
 
     def __init__(
-        self, title="", description="Upload Structure", allow_trajectories=False
+        self,
+        title="",
+        description="Upload Structure",
+        allow_trajectories=False,
+        add_auxiliary_cell=True,
```
danielhollas, line 382: Not sure about the parameter name, suggestions welcome, especially shorter ones :-)

danielhollas:
I added unit tests and updated the PR description so this is ready for review.

aiidalab_widgets_base/structures.py
yakutovicha: To me, the name is good. It does the job.

yakutovicha (approved):
The code looks good, I've also tested it on the new image with the migrated [surfaces app](https://github.com/nanotech-empa/aiidalab-empa-surfaces/pull/275), and it works without issues!

image

danielhollas:
Perfect, thanks!

PR #593, opened 2024-04-30
Bump version v2.2.0a1 -> v2.2.0

danielhollas:

PR #594, opened 2024-04-30
Run test publish workflow on push to release/ branches

danielhollas:
I noticed when releasing version 2.2.0 that the test publish workflow was not triggered on the PR in #593. This was my mistake: At some point I changed the triggering event from `push` to `pull_request`. However, the `branches` filter that we use so that the test-publish is only run on release branches, has a different semantic meaning for `pull_request` event --- it specifies the **target** branch, not the branch that is being merged.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,6 +24,7 @@ repos:
       rev: 0.2.3
       hooks:
           - id: yamlfmt
+            args: [--preserve-quotes]
 
     - repo: https://github.com/sirosen/check-jsonschema
       rev: 0.28.1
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -24,6 +24,7 @@ repos:
       rev: 0.2.3
       hooks:
           - id: yamlfmt
+            args: [--preserve-quotes]
```
danielhollas, line 27: https://github.com/jumanjihouse/pre-commit-hook-yamlfmt/issues/23

danielhollas:
@yakutovicha can you take a look?

Note that the publish workflow as specified here was there basically as is before I made any changes, the idea that you can inspect a release by first releasing on TestPyPI, where you can then inspect the package. But practically this couldn't really be used with bumpver, since you'd always push to master. (not saying which is better, just providing context).

danielhollas:
Going to merge this, since we need to release a patch version.

Issue #581, opened 2024-04-06
ValueError: invalid port name `_source_structure`: cannot start with an underscore


AndresOrtegaGuerrero:
I am getting a error as a result of this line

https://github.com/aiidalab/aiidalab-widgets-base/blob/86681887f13a2838637d21f3d3293c2f5880c367/aiidalab_widgets_base/structures.py#L224

AttributeError                            Traceback (most recent call last)
/opt/conda/lib/python3.9/site-packages/plumpy/processes.py in spec(cls)
    181         try:
--> 182             return cls.__getattribute__(cls, '_spec')
    183         except AttributeError:

AttributeError: 'ProcessStateMachineMeta' object has no attribute '_spec'

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
/opt/conda/lib/python3.9/site-packages/aiida/engine/processes/ports.py in validate_port_name(port_name)
    232         try:
--> 233             validate_link_label(port_name)
    234         except ValueError as exception:

/opt/conda/lib/python3.9/site-packages/aiida/common/links.py in validate_link_label(link_label)
    117     if link_label.startswith('_'):
--> 118         raise ValueError('cannot start with an underscore')
    119 

ValueError: cannot start with an underscore

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
~/apps/quantum-espresso/src/aiidalab_qe/app/structure/__init__.py in confirm(self, _)
    162 
    163     def confirm(self, _=None):
--> 164         self.manager.store_structure()
    165         self.confirmed_structure = self.structure
    166         self.message_area.value = ""

/opt/conda/lib/python3.9/site-packages/aiidalab_widgets_base/structures.py in store_structure(self, _)
    224                 return self.structure_node
    225 
--> 226             structure_node = user_modifications(self.input_structure)
    227 
    228         else:

/opt/conda/lib/python3.9/site-packages/aiida/engine/processes/functions.py in decorated_function(*args, **kwargs)
    240         def decorated_function(*args, **kwargs):
    241             """This wrapper function is the actual function that is called."""
--> 242             result, _ = run_get_node(*args, **kwargs)
    243             return result
    244 

/opt/conda/lib/python3.9/site-packages/aiida/engine/processes/functions.py in run_get_node(*args, **kwargs)
    182             manager = get_manager()
    183             runner = manager.get_runner()
--> 184             inputs = process_class.create_inputs(*args, **kwargs)
[...]

PR #597, opened 2024-05-03
remove underscore in function call

AndresOrtegaGuerrero:
This address #581

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -221,7 +221,7 @@ class StructureManagerWidget(ipw.VBox):
         ):
             # Make a link between self.input_structure and self.structure_node
             @engine.calcfunction
-            def user_modifications(source_structure): # noqa
+            def user_modifications(source_structure): # noqa F841
                 return self.structure_node
 
             structure_node = user_modifications(self.input_structure)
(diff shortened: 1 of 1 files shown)


superstar54 (approved):
Thanks for the fix, LGTM!

superstar54:
@AndresOrtegaGuerrero, the pre-commit test fails; please fix it before merging.

danielhollas (approved):
@AndresOrtegaGuerrero I guess the precommit fail comes from ruff complaining about unused function argument, so you're going to have to add noqa comment.

It's quite possible I introduced this bug when I was switching linters to ruff, my apologies if that is the case. 
Ideally we should have a test case for this, but that can wait for a follow up PR.

AndresOrtegaGuerrero:
> @AndresOrtegaGuerrero I guess the precommit fail comes from ruff complaining about unused function argument, so you're going to have to add noqa comment.
> 
> It's quite possible I introduced this bug when I was switching linters to ruff, my apologies if that is the case. Ideally we should have a test case for this, but that can wait for a follow up PR.

Thank you @danielhollas ! Have a look

aiidalab_widgets_base/structures.py
```diff
@@ -221,7 +221,7 @@ def store_structure(self, _=None):
         ):
             # Make a link between self.input_structure and self.structure_node
             @engine.calcfunction
-            def user_modifications(_source_structure):
+            def user_modifications(source_structure):  # noqa
```
danielhollas, line 224: It would be better to have a more specific noqa, for the specific linter rule that ruff is complaining about. Please have a look at their docs, I don't remember the notation from the top of my head.

PR #599, opened 2024-05-08
Test with aiida-core=2.5

danielhollas:
- test aiida-core 2.5.1
- update test requirements
- squash some deprecations warnings


tests/conftest.py
```diff
@@ -14,15 +14,7 @@
 
 
 @pytest.fixture
-def fixture_localhost(aiida_localhost):
-    """Return a localhost `Computer`."""
-    localhost = aiida_localhost
-    localhost.set_default_mpiprocs_per_machine(1)
```
danielhollas, line 20: I looked at the source code of the `aiida_localhost` fixture, and it seems to set this up by default, so I am not sure why this was here...

unkcpz: I remember the fixture was directly copied from https://github.com/aiidateam/aiida-quantumespresso/blob/7b6c210a80942aed8ed08fa78f0d123d3227e750/tests/conftest.py#L42
Maybe it is not needed anymore? Then we can also remove it from `aiida-quantumespresso`.

danielhollas: Ah, good to know, thanks. Looking at git blame, that fixture is five years old, so indeed most likely can be removed from there as well.

danielhollas:
Merging this since the changes here are low-risk and we want to start testing with 2.5.

PR #604, opened 2024-05-15
Speedup readthedocs build with uv

danielhollas:
I configured the readthedocs build to use `uv` to install dependencies, according to https://github.com/readthedocs/readthedocs.org/pull/11322

This cuts the build time in half (by full minute!) so imo it's worth doing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
                   python-version: '3.10'
 
             - name: Install uv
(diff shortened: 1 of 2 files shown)


yakutovicha (approved):
Looks good to me, at least it is in sync with https://github.com/readthedocs/readthedocs.org/pull/11322.

@unkcpz do you want to have a look?

danielhollas:
Thanks @yakutovicha 

@unkcpz I've merged this, by you can still review this as part of #607

PR #607, opened 2024-05-16
Hide private methods in API documentation

danielhollas:
I noticed that the current [API documentation](https://aiidalab-widgets-base.readthedocs.io/en/latest/apidoc/aiidalab_widgets_base.html) is extremely messy, because it includes all the private methods. I am not sure what was the original motivation for this, but I'd suggest to remove it.

New version:
https://aiidalab-widgets-base--606.org.readthedocs.build/en/606/apidoc/aiidalab_widgets_base.html

Additional changes: 
 - simplified the sphinx-apidoc invocation
 - basic pins for docs dependencies to make the build more future-proof

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
                   python-version: '3.10'
 
             - name: Install uv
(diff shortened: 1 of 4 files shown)


unkcpz (approved):
@danielhollas thanks! all good.

Issue #595, opened 2024-04-30
Publish AWB on conda-forge


danielhollas:
Just pasting here a handy guide how to create new conda-forge packages.

https://www.pyopensci.org/python-package-guide/tutorials/publish-conda-forge.html

danielhollas:
@yakutovicha thanks. Happy to make a PR, but as I was looking through the code, I was wondering if this needs to be part of AWB at all, especially if we're going to make AWB only a library. 

Admittedly I know nothing about how this is used, but the code in `aiidalab_widgets_base/elns.py` does not use any of the other parts of AWB, and so it's completely independent. Should it be turned into its own library or app? Should it be part of `aiidalab-eln`?

danielhollas:
I had two hours on the airport so I just decided to go for it. :sweat_smile: I published a new pre-release version which allowed me to generate a valid conda-forge recipe with grayskull. I've opened a PR on staged-recipes. https://github.com/conda-forge/staged-recipes/pull/26882

@yakutovicha @unkcpz could you please comment on that PR that you are okay with being added maintainers?

There is still some work left around dependencies, since `pip check` is failing after the conda install. Will continue to work on that.

unkcpz:
> @yakutovicha @unkcpz could you please comment on that PR that you are okay with being added maintainers?

I am fine with it, thanks for adding me.

danielhollas:
To clarify, you have to make a comment on the PR for staged-recipes, it's part of their checklist.

unkcpz:
>  you have to make a comment on the PR for staged-recipes, it's part of their checklist.

So we are good now?

danielhollas:
> So we are good now?

You need to make a comment on the PR :-P Here is the link https://github.com/conda-forge/staged-recipes/pull/26882

unkcpz:
Ah, sorrry, I though this is the PR :smile:

danielhollas:
The recipe has been merged and AWB is not on conda-forge! :tada: 
https://github.com/conda-forge/aiidalab-widgets-base-feedstock

For now @unkcpz  and I are the feedstock maintainers. It would be nice to have at least one more person. (@edan-bainglass LMK in case you have experience with conda-forge :-) )  

I've opened a testing PR for integrating AWB into the docker stack. https://github.com/aiidalab/aiidalab-docker-stack/pull/483

Closing this issue.

PR #608, opened 2024-05-27
Remove eln widgets

danielhollas:
Discussed briefly in #595. We could make eln into an optional dependency, but honestly I don't know if it makes sense for it to be in AWB. It does not directly interact with any other widgets and is completely standalone so it would make more sense to me to move the code elsewhere (perhaps into `aiidalab-eln` package itself) and publish it as a separate app in aiidalab app registry. This is especially true given that we want to make AWB into a library and remove it from the app registry eventually.

@yakutovicha WDYT? Also briefly discussed with @edan-bainglass in last meeting. 

This will unblock publishing AWB on conda-forge and integrating it into docker image.


edan-bainglass:
@danielhollas a bit more context - at the [MADICES](https://madices.github.io/) workshop on open research data interoperability (April 2024), we discussed how to make two platform (say, AiiDAlab and an electronic lab notebook, or ELN) interact without knowing each others internal API. Without going too much into details, one thing that was clear is that to interact with another platform in any way, you must first connect and authenticate. `aiidalab-eln`, as I understand it, is intended to facilitate this part. @yakutovicha had developed the plugin as a means of connecting with the `cheminfo` ELN and was extending it to other platforms at MADICES. With this in mind, the widget could be general, potentially used by apps as part of a larger port widget (which I'll likely be developing soon) to connect and authenticate with other platforms.

edan-bainglass (commented):
@danielhollas a bit more context - at the [MADICES](https://madices.github.io/) workshop on open research data interoperability (April 2024), we discussed how to make two platform (say, AiiDAlab and an electronic lab notebook, or ELN) interact without knowing each others internal API. Without going too much into details, one thing that was clear is that to interact with another platform in any way, you must first connect and authenticate. aiidalab-eln, as I understand it, is intended to facilitate this part. @yakutovicha had developed the plugin as a means of connecting with the cheminfo ELN and was extending it to other platforms at MADICES. With this in mind, the widget could be general, potentially used by apps as part of a larger port widget (which I'll likely be developing soon) to connect and authenticate with other platforms.

danielhollas:
Thank you for the context @edan-bainglass, that's very helpful. So are you saying that these widgets should stay in AWB? I am happy to open an alternative PR which only makes them optional.

edan-bainglass:
Correct, though somewhat hopeful. We'll have to see how it works, but for now, yes, I would say they should stay. I second their optionality.

danielhollas:
Thanks @edan-bainglass, I've created #609 and am closing this one.

yakutovicha:
Thanks, @edan-bainglass for explaining. I agree that the ELN support should stay here for the moment. But I would opt to remove it in 3.0.0 and place it either as a separate app or part of the home app, as proposed by @danielhollas.

PR #609, opened 2024-05-28
Make ELN widgets optional

danielhollas:
As discussed in #608 and #595, this PR makes `aiidalab-eln` dependency optional via new `eln` extras. The ELN widgets will show an error if the dependency is not installed. 

@edan-bainglass @yakutovicha please test, I am not familiar with these widgets. Also please feel free to tweak the error message.

WARNING: This is a breaking change and so it should be released in 0.2.3.


edan-bainglass (approved):
LGTM!

PR #613, opened 2024-06-26
deps: Vendor more_itertools.consecutive_groups

danielhollas:
`more_itertools` package contains useful functions built on top of stdlib itertools, but we only use a single function from the package so let's vendor it and remove the dependency. 

This means app developers can happily use this package without worrying about version clashes with AWB. 

Part of #392

### DIFF
--- a/aiidalab_widgets_base/utils/__init__.py
+++ b/aiidalab_widgets_base/utils/__init__.py
@@ -1,16 +1,17 @@
 """Some utility functions used acrross the repository."""
 
+import itertools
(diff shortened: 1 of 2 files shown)


danielhollas:
I'll merge this to include in an alpha release. But please still feel free to review and pushback against this, happy to incorporate feedback in a followup.

unkcpz:
thanks! It appeared before I join the team, so feel free to optimize it, I have no opinion on these.

PR #616, opened 2024-07-02
Test with aiida 2.6.2

danielhollas:
Just released a new image with aiida-2.6.2 so we should switch our tests (both notebook tests and unit tests) here to this version.


danielhollas:
Blocked on https://github.com/aiidalab/aiidalab-docker-stack/pull/474

unkcpz (approved):
Thanks! @danielhollas 
Tests are all passed so I think it is good to ship.

PR #617, opened 2024-07-04
Adding support for `orm.Containerized` codes

mikibonacci:
This is obtained using the AiidaCodeSetup().code_setup attribute. 

I am not sure is the better way, as we should also provide the possibility to change the `image_name` and the `engine_command` via the GUI. However, I think that for now this is acceptable, as the containerized codes are mainly 
set up by using the information on the `aiida-resource-registry` repo. Actually, it is a really new thing and for now only the `phonopy@merlin-cpu` is defined.

### DIFF
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -1204,7 +1204,7 @@ class AiidaCodeSetup(ipw.VBox):
             kwargs = {key: getattr(self, key).value for key in items_to_configure}
 
             # Check for additional keys needed for orm.ContainerizedCode
-            for container_key in containerized_codes_additional_items:
+            for container_key in containerized_code_additional_items:
                 if container_key in self.code_setup.keys():
                     kwargs[container_key] = self.code_setup[container_key]
 
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -1196,7 +1196,17 @@ class AiidaCodeSetup(ipw.VBox):
                 "append_text",
             ]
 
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/computational_resources.py
```diff
@@ -1196,16 +1196,26 @@ def on_setup_code(self, _=None):
...
+
             # set computer from its widget value the UUID of the computer.
             computer = orm.load_computer(self.computer.value)
 
             # Checking if the code with this name already exists
             qb = orm.QueryBuilder()
             qb.append(orm.Computer, filters={"uuid": computer.uuid}, tag="computer")
             qb.append(
-                orm.InstalledCode,
+                (orm.InstalledCode, orm.ContainerizedCode),
```
unkcpz, line 1218:
```suggestion
                orm.AbstractCode,
```

unkcpz (changes requested):
Thanks @mikibonacci, some minor requests.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -1196,16 +1196,26 @@ def on_setup_code(self, _=None):
                 "append_text",
             ]
 
+            containerized_codes_additional_items = [
```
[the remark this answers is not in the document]
mikibonacci: Yes @unkcpz , you are right, it is indeed better

aiidalab_widgets_base/computational_resources.py
mikibonacci: @unkcpz  thanks also for this, indeed it is better to provide the abstract class here

unkcpz (approved):
Thanks!

PR #621, opened 2024-07-08
Various dependency fixes

danielhollas:
These were uncovered when trying to build a conda recipe for AWB.


aiidalab_widgets_base/nodes.py
```diff
@@ -7,7 +7,6 @@
 import traitlets as tl
 from aiida import common, engine, orm
 from aiida.cmdline.utils.ascii_vis import calc_info
-from aiidalab.app import _AiidaLabApp
```
danielhollas, line 10: We should remove the `aiidalab` dependency from AWB. I'll open a separate issue for that since that requires some discussion. Here I simply move this import inline where it is actually used.

danielhollas:
Going to merge this so I can continue debugging the conda build.

PR #624, opened 2024-07-11
Add `static` folder for static files and pre-loaded stylesheets (using new utility)

edan-bainglass:
This PR adds a `static` folder to the widgets base package. At the moment, the folder holds stylesheets that are loaded (using a new utility function) on any import from the widgets base package.

### DIFF
--- a/aiidalab_widgets_base/__init__.py
+++ b/aiidalab_widgets_base/__init__.py
@@ -1,7 +1,5 @@
 """Reusable widgets for AiiDAlab applications."""
 
-from aiida.manage import get_profile
-
 _WARNING_TEMPLATE = """
 <div style="background-color: #f7f7f7; border: 2px solid #e0e0e0; padding: 20px; border-radius: 5px;">
     <p style="font-size: 16px; font-weight: bold; color: #ff5733;">Warning:</p>
@@ -27,22 +25,31 @@ def is_running_in_jupyter():
         return False
 
 
-# load the default profile if no profile is loaded, and raise a deprecation warning
-# this is a temporary solution to avoid breaking existing notebooks
-# this will be removed in the next major release
-if is_running_in_jupyter() and get_profile() is None:
-    # if no profile is loaded, load the default profile and raise a deprecation warning
-    from aiida import load_profile
+if is_running_in_jupyter():
+    from aiida.manage import get_profile
     from IPython.display import HTML, display
 
-    load_profile()
+    # load the default profile if no profile is loaded, and raise a deprecation warning
+    # this is a temporary solution to avoid breaking existing notebooks
+    # this will be removed in the next major release
+    if get_profile() is None:
+        # if no profile is loaded, load the default profile and raise a deprecation warning
+        from aiida import load_profile
 
-    profile = get_profile()
-    assert profile is not None, "Failed to load the default profile"
+        load_profile()
 
-    # raise a deprecation warning
-    warning = HTML(_WARNING_TEMPLATE.format(profile=profile.name, version="v3.0.0"))
-    display(warning)
+        profile = get_profile()
+        assert profile is not None, "Failed to load the default profile"
+
+        # raise a deprecation warning
+        warning = HTML(_WARNING_TEMPLATE.format(profile=profile.name, version="v3.0.0"))
+        display(warning)
+
+    from .static import styles
+    from .utils.loaders import load_css_stylesheet
+
+    load_css_stylesheet(package=styles)
+
 
 from .computational_resources import (
     ComputationalResourcesWidget,
--- a/aiidalab_widgets_base/__init__.py
+++ b/aiidalab_widgets_base/__init__.py
@@ -11,6 +11,20 @@ load_profile();</pre>
 """
 
 
+def load_default_profile():
+    """Loads the default profile if none loaded and warn of deprecation."""
+    from aiida import load_profile
+
+    load_profile()
+
+    profile = get_profile()
+    assert profile is not None, "Failed to load the default profile"
+
+    # raise a deprecation warning
+    warning = HTML(_WARNING_TEMPLATE.format(profile=profile.name, version="v3.0.0"))
+    display(warning)
+
+
 # We only detect profile and throw a warning if it is on the notebook
 # It is not necessary to do this in the unit tests
 def is_running_in_jupyter():
@@ -33,17 +47,7 @@ if is_running_in_jupyter():
     # this is a temporary solution to avoid breaking existing notebooks
     # this will be removed in the next major release
     if get_profile() is None:
-        # if no profile is loaded, load the default profile and raise a deprecation warning
-        from aiida import load_profile
-
-        load_profile()
-
-        profile = get_profile()
-        assert profile is not None, "Failed to load the default profile"
-
-        # raise a deprecation warning
-        warning = HTML(_WARNING_TEMPLATE.format(profile=profile.name, version="v3.0.0"))
-        display(warning)
+        load_default_profile()
 
     from .static import styles
     from .utils.loaders import load_css_stylesheet
--- a/aiidalab_widgets_base/__init__.py
+++ b/aiidalab_widgets_base/__init__.py
@@ -40,6 +40,8 @@ def is_running_in_jupyter():
 
 
 if is_running_in_jupyter():
+    from pathlib import Path
+
     from aiida.manage import get_profile
     from IPython.display import HTML, display
 
@@ -49,10 +51,9 @@ if is_running_in_jupyter():
     if get_profile() is None:
         load_default_profile()
 
-    from .static import styles
-    from .utils.loaders import load_css_stylesheet
+    from .utils.loaders import load_css
 
-    load_css_stylesheet(package=styles)
+    load_css(css_path=Path(__file__).parent / "static/styles")
 
 
 from .computational_resources import (
--- a/aiidalab_widgets_base/static/styles/README.md
+++ b/aiidalab_widgets_base/static/styles/README.md
@@ -1,0 +1,3 @@
+# Stylesheets for AiiDAlab Widgets Base
+
+This folder contains `.css` stylesheets, which are loaded on any import from the AiiDAlab widgets base package.
--- a/aiidalab_widgets_base/utils/loaders.py
+++ b/aiidalab_widgets_base/utils/loaders.py
@@ -1,0 +1,38 @@
+from importlib.resources import Package, files
+
+from IPython.display import Javascript, display
+
+
+def load_css_stylesheet(package: Package, filename: str = ""):
+    """Load a CSS stylesheet from a package and inject it into the DOM.
+
+    Parameters
+    ----------
+    `package` : `Package`
+        The package where the CSS file is located.
+    `filename` : `str`, optional
+        The name of the CSS file to load.
+        If not provided, all CSS files in the package will be loaded.
+    """
+    root = files(package)
+
+    filenames = []
+    if filename:
+        filenames.append(filename)
+    else:
+        filenames.extend(
+            path.name
+            for path in root.iterdir()
+            if path.is_file() and path.name.endswith(".css")
+        )
+
+    for filename in filenames:
+        stylesheet = (root / filename).read_text()
+        display(
+            Javascript(f"""
+                var style = document.createElement('style');
+                style.type = 'text/css';
+                style.innerHTML = `{stylesheet}`;
+                document.head.appendChild(style);
+            """)
+        )
--- a/aiidalab_widgets_base/utils/loaders.py
+++ b/aiidalab_widgets_base/utils/loaders.py
@@ -1,33 +1,28 @@
-from importlib.resources import Package, files
+from __future__ import annotations
+
+from pathlib import Path
 
 from IPython.display import Javascript, display
 
 
-def load_css_stylesheet(package: Package, filename: str = ""):
-    """Load a CSS stylesheet from a package and inject it into the DOM.
+def load_css(css_path: Path | str) -> None:
+    """Load and inject CSS stylesheets into the DOM.
 
     Parameters
     ----------
-    `package` : `Package`
-        The package where the CSS file is located.
-    `filename` : `str`, optional
-        The name of the CSS file to load.
-        If not provided, all CSS files in the package will be loaded.
+    `css_path` : `Path` | `str`
+        The path to the CSS stylesheet. If the path is a directory,
+        all CSS files in the directory will be loaded.
     """
-    root = files(package)
+    path = Path(css_path)
 
-    filenames = []
-    if filename:
-        filenames.append(filename)
-    else:
-        filenames.extend(
-            path.name
-            for path in root.iterdir()
-            if path.is_file() and path.name.endswith(".css")
-        )
+    if not path.exists():
+        raise FileNotFoundError(f"CSS file or directory not found: {path}")
 
-    for filename in filenames:
-        stylesheet = (root / filename).read_text()
+    filenames = [*path.glob("*.css")] if path.is_dir() else [path]
+
+    for fn in filenames:
+        stylesheet = fn.read_text()
         display(
             Javascript(f"""
                 var style = document.createElement('style');
--- a/tests/test_loaders.py
+++ b/tests/test_loaders.py
@@ -1,0 +1,8 @@
+from aiidalab_widgets_base.utils.loaders import load_css_stylesheet
+
+
+def test_load_css_stylesheet():
+    """Test `load_css_stylesheet` function."""
+    package = "aiidalab_widgets_base.static.styles"
+    load_css_stylesheet(package=package, filename="global.css")
+    load_css_stylesheet(package=package)
--- a/tests/test_loaders.py
+++ b/tests/test_loaders.py
@@ -1,8 +1,10 @@
-from aiidalab_widgets_base.utils.loaders import load_css_stylesheet
+from pathlib import Path
+
+from aiidalab_widgets_base.utils.loaders import load_css
 
 
-def test_load_css_stylesheet():
-    """Test `load_css_stylesheet` function."""
-    package = "aiidalab_widgets_base.static.styles"
-    load_css_stylesheet(package=package, filename="global.css")
-    load_css_stylesheet(package=package)
+def test_load_css():
+    """Test `load_css` utility."""
+    css_dir = Path("aiidalab_widgets_base/static/styles")
+    load_css(css_path=css_dir)
+    load_css(css_path=css_dir / "global.css")
--- a/docs/source/contribute/index.rst
+++ b/docs/source/contribute/index.rst
@@ -8,3 +8,20 @@ Contributions to the AiiDAlab widgets are highly welcome and can take different 
 * `Report bugs <https://github.com/aiidalab/aiidalab-widgets-base/issues>`_.
 * `Feature requests <https://github.com/aiidalab/aiidalab-widgets-base/issues>`_.
 * Help us improve the documentation of widgets.
+
+**************
+Widget styling
+**************
+
+Though ``ipywidgets`` does provide some basic styling options via the ``layout`` and ``style`` attributes, it is often not enough to create a visually appealing widget.
+As such, we recommend the use of `CSS <https://www.w3schools.com/css/>`_ stylesheets to style your widgets.
+These may be packaged under ``aiidalab_widgets_base/static/styles``, which are automatically loaded on import via the ``load_css`` utility.
+
+A ``global.css`` stylesheet is made available for global html-tag styling and ``ipywidgets`` or ``Jupyter`` style overrides.
+For more specific widgets and components, please add a dedicated stylesheet.
+Note that all stylesheets in the ``styles`` directory will be loaded on import.
+
+We recommend using classes to avoid style leaking outside of the target widget.
+We also advise causion when using the `!important <https://www.w3schools.com/css/css_important.asp>`_ flag on CSS properties, as it may interfere with other stylesheets.
+
+If you are unsure about the styling of your widget, feel free to ask for help on the `AiiDAlab Discourse channel <https://aiida.discourse.group/tag/aiidalab>`_.
--- a/setup.cfg
+++ b/setup.cfg
@@ -64,6 +64,10 @@ docs =
     pydata-sphinx-theme~=0.15
     myst-nb~=1.1
 
+
+[options.package_data]
+aiidalab_widgets_base.static.styles = *.css
+
 [bumpver]
 current_version = "v2.3.0a1"
 version_pattern = "vMAJOR.MINOR.PATCH[PYTAGNUM]"
--- a/tests_notebooks/static/styles/test.css
+++ b/tests_notebooks/static/styles/test.css
@@ -1,0 +1,3 @@
+.red-text {
+  color: rgb(255, 0, 0);
+}
--- a/tests_notebooks/test_notebook.ipynb
+++ b/tests_notebooks/test_notebook.ipynb
@@ -1,0 +1,67 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import ipywidgets as ipw"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from aiida import load_profile\n",
+    "\n",
+    "load_profile();"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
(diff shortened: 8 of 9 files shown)


aiidalab_widgets_base/__init__.py
```diff
@@ -27,22 +25,31 @@ def is_running_in_jupyter():
...
+        # raise a deprecation warning
+        warning = HTML(_WARNING_TEMPLATE.format(profile=profile.name, version="v3.0.0"))
+        display(warning)
+
+    from .static import styles
+    from .utils.loaders import load_css_stylesheet
 
-    profile = get_profile()
-    assert profile is not None, "Failed to load the default profile"
+    load_css_stylesheet(package=styles)
```
danielhollas, line 51: To improve readability slightly, I'd move this to the top before the loading of the AiiDA profile.

aiidalab_widgets_base/utils/loaders.py
```diff
@@ -0,0 +1,38 @@
+from importlib.resources import Package, files
+
+from IPython.display import Javascript, display
+
+
+def load_css_stylesheet(package: Package, filename=""):
```
danielhollas, line 6: We currently don't check types so I'd remove it. Or alternatively, type the other parameter and return parameter as well.

aiidalab_widgets_base/static/styles/README.md
```diff
@@ -0,0 +1,15 @@
+# Stylesheets for AiiDAlab Widgets Base
+
+This folder contains the following:
+
+- `scss` - a folder for SCSS-format stylesheets to be compiled
+- `.css` files  - compiled CSS stylesheets
+
+It is recommended to add SCSS stylesheets in the SCSS folder for functionality and flexibility (see https://sass-lang.com/).
+
+Compiling SCSS into CSS can be done using:
```
danielhollas, line 10: Does this mean that compiled CSS files should be checked in the repo? If so, we need a pre-commit hook that will check that the files are up-to-date. Alternatively, we need to create a hook that is called when the package is installed.

aiidalab_widgets_base/static/styles/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+- The [compile-hero](https://marketplace.visualstudio.com/items?itemName=Wscats.eno) extension, if using VS Code
+- The `sass` [CLI](https://sass-lang.com/install/)
```
danielhollas, line 13: I think we should have a simple bash script that will invoke the correct invocation of the compiler. Otherwise it's not immediately clear what needs to be done.  (this could then be used as a pre-commit hook, as noted above).

danielhollas (changes requested):
Hi @edan-bainglass,

thanks so much on working on this, I am really excited to see this land.

My main concern here is about SCSS, and is twofold.
- It introduces additional complexity of extra compilation step (see my other comments)
- I guess that most of us are not so proficient with CSS, and SCSS is yet another cognitive overhead. Especially as we just start using CSS directly, my feeling is that we should start simple.

If you feel strongly about it, could you please open an issue (or topic on Discourse), and explain what are the advantages of SCSS (and what is the additional complexity) so we can discuss it a bit before committing?

aiidalab_widgets_base/__init__.py
edan-bainglass: I think it had to be after. Otherwise, I got a warning message about not importing from the widgets base prior to loading the profile

aiidalab_widgets_base/utils/loaders.py
edan-bainglass: The other parameter is typed via its default (str). The return type is None, so not strictly required. Do you feel differently about this? I'm not against it. In principle, I know packages such as pydantic enforce typing even in cases where static analysis tools can assert it.

aiidalab_widgets_base/static/styles/README.md
edan-bainglass: I gotta be honest, I'm not all that fond of the way I did it. But I couldn't come up with an alternative. It would be great to implement something more flexible, where one can use both CSS and SCSS. Let's chat on Discourse.

aiidalab_widgets_base/static/styles/README.md
edan-bainglass: See my last comment

edan-bainglass:
@danielhollas thanks for your comments. Let's discuss further on [Discourse](https://aiida.discourse.group/t/use-of-external-stylesheets-css-or-even-scss/434).

aiidalab_widgets_base/__init__.py
danielhollas: ah, of course, makes sense.

edan-bainglass:
@danielhollas could you please take a look at why the browser tests are failing? I'm not sure I understand what's happening here.

danielhollas:
@edan-bainglass The same failures are seen on the main branch so it is not related to this PR.

edan-bainglass:
@danielhollas I see. Okay then. Please review 🙏🏻 I have a few PRs built on top of this one both here and in the QE app. I think all is well other than the tests. I've removed the SCSS part.

aiidalab_widgets_base/utils/loaders.py
edan-bainglass: I see. I've added the explicit `str` type :)

aiidalab_widgets_base/utils/loaders.py
```diff
@@ -0,0 +1,38 @@
...
+
+    filenames = []
+    if filename:
+        filenames.append(filename)
+    else:
+        filenames.extend(
+            path.name
+            for path in root.iterdir()
+            if path.is_file() and path.name.endswith(".css")
+        )
```
danielhollas, line 27: Maybe a bit simpler
```suggestion
    if filename:
        filenames = [filename]
    else:
        filenames = [
            path.name
            for path in root.iterdir()
            if path.is_file() and path.name.endswith(".css")
        ]
```

aiidalab_widgets_base/utils/loaders.py
```diff
@@ -0,0 +1,38 @@
...
+
+    for filename in filenames:
+        stylesheet = (root / filename).read_text()
```
danielhollas, line 30: nit: This for loop var shadows the function parameter. Not a huge deal but I am surprised that ruff does not complain about it.

```suggestion
    for f in filenames:
        stylesheet = (root / f).read_text()
```

danielhollas (changes requested):
Just some small nits.

My main thing is that it would be nice to demonstrate this functionality to prove that is working by including some actual CSS.

Perhaps a nice candidate would be this code that defines CSS inline for tables in DictViewer.

https://github.com/aiidalab/aiidalab-widgets-base/blob/de5f0cc23e1e058b46922a2203dd75b292b0f13e/aiidalab_widgets_base/viewers.py#L104-L113

aiidalab_widgets_base/__init__.py
```diff
@@ -27,22 +25,31 @@ def is_running_in_jupyter():
...
+        from aiida import load_profile
+
+        load_profile()
+
+        profile = get_profile()
+        assert profile is not None, "Failed to load the default profile"
+
+        # raise a deprecation warning
+        warning = HTML(_WARNING_TEMPLATE.format(profile=profile.name, version="v3.0.0"))
+        display(warning)
```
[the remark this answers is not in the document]
edan-bainglass: Done!

aiidalab_widgets_base/utils/loaders.py
edan-bainglass: Indeed surprising. Done!

aiidalab_widgets_base/utils/loaders.py
edan-bainglass: Done!

aiidalab_widgets_base/__init__.py
```diff
@@ -13,6 +11,20 @@
 """
 
 
+def load_default_profile():
```
superstar54, line 14: Could you move this to the `utils/__init__.py`?

aiidalab_widgets_base/__init__.py
```diff
@@ -27,22 +39,21 @@ def is_running_in_jupyter():
...
-    # if no profile is loaded, load the default profile and raise a deprecation warning
-    from aiida import load_profile
+if is_running_in_jupyter():
+    from aiida.manage import get_profile
     from IPython.display import HTML, display
 
-    load_profile()
+    # load the default profile if no profile is loaded, and raise a deprecation warning
+    # this is a temporary solution to avoid breaking existing notebooks
+    # this will be removed in the next major release
```
superstar54, line 48: Will the next major release fix this?

aiidalab_widgets_base/utils/loaders.py
```diff
@@ -0,0 +1,37 @@
+from importlib.resources import Package, files
+
+from IPython.display import Javascript, display
+
+
+def load_css_stylesheet(package: Package, filename: str = ""):
```
superstar54, line 6: I suggest changing `package` to `module`, as it accepts a module.

tests/test_loaders.py
```diff
@@ -0,0 +1,8 @@
+from aiidalab_widgets_base.utils.loaders import load_css_stylesheet
+
+
+def test_load_css_stylesheet():
+    """Test `load_css_stylesheet` function."""
+    package = "aiidalab_widgets_base.static.styles"
+    load_css_stylesheet(package=package, filename="global.css")
+    load_css_stylesheet(package=package)
```
superstar54, line 8: The test here will not work because it is not running inside a notebook. To test whether the CSS works, you need to add a test in `test_notebooks/test_notebooks.py.`

superstar54 (changes requested):
Hi @edan-bainglass , thansk for the work.

- It would be great that you add this in the docs to tell the developer how to use it in their app.
- You need add your test in the `test_notebooks`.

aiidalab_widgets_base/__init__.py
danielhollas: This needs to stay here, this code was here before and was just moved to a function.

aiidalab_widgets_base/__init__.py
danielhollas: This is unrelated to this PR. Indeed we plan to remove this in next major release most likely.

aiidalab_widgets_base/__init__.py
superstar54: Ah, you are right!

aiidalab_widgets_base/__init__.py
superstar54: Ok.

aiidalab_widgets_base/utils/loaders.py
danielhollas: I think the name is correct, it accepts a package (i.e. a folder), not a module (which would be an individual python file).

tests/test_loaders.py
edan-bainglass: Are you saying to add a test to test_notebooks.py, or to move this test to test_notebooks.py? Also, could you explain why exactly it doesn't work?

aiidalab_widgets_base/utils/loaders.py
superstar54: `package` is more like a Python pacage, like the `aiidalab_widgets_base`. From the example, 

```python
    from .static import styles
    from .utils.loaders import load_css_stylesheet

    load_css_stylesheet(package=styles)
```
It passes the module `styles` into it.  'aiidalab_widgets_base.static.styles' is a module, not a package.

tests/test_loaders.py
superstar54: Yes, adding a new test in the test_notebooks.py, and check if the CSS is indeed loaded in the selenium_driver.

> Also, could you explain why exactly it doesn't work?

the `test.py` file is not running in the notebook, thus the `display(Javascript...` code will not run. You can add some wrong code inside it, and the test will not throw error.

edan-bainglass:
> Just some small nits.
> 
> My main thing is that it would be nice to demonstrate this functionality to prove that is working by including some actual CSS.
> 
> Perhaps a nice candidate would be this code that defines CSS inline for tables in DictViewer.
> 
> https://github.com/aiidalab/aiidalab-widgets-base/blob/de5f0cc23e1e058b46922a2203dd75b292b0f13e/aiidalab_widgets_base/viewers.py#L104-L113
> 
> EDIT: I know that you have follow-up PRs that do this so feel free to ignore this.

@danielhollas yeah, maybe an example is good just for this commit. Since you've suggested one from the code base, it's a great test case. Matter of fact, I could use a test case for a notebook test, as recommended by @superstar54. So to clarify, the bit of code you mentioned (`df` styles) can be removed once I add it via css, correct?

aiidalab_widgets_base/utils/loaders.py
danielhollas: I don't think that's correct, citing e.g. from [this glossary](https://the-hitchhikers-guide-to-packaging.readthedocs.io/en/latest/glossary.html)

> **Module**
>     A python source code file (ex. mymodule.py), most often found in a package (mypackage/mymodule.py). You can import a module: import mymodule, or import mypackage.mymodule if it lives in mypackage.
>
> **Package** 
>     A directory containing an __init__.py file (ex. mypackage/__init__.py), and also usually containing modules (possibly along with other packages). You can import a package: import mypackage
>     A package should not be confused with a compressed archive file used to install Python code.

`aiidalab_widgets_base` is a package, containing `static` subpackage which itself contains `styles` subpackage. After all, the type of this variable *is* called `Package`.

tests/test_loaders.py
superstar54: For the test in selenium, for example, you create a vbox, and attach a class, then in your testing CSS  file, you change its color, and in selenium_driver, you try to find the vbox, and verify the color.

danielhollas: The test here is fine, I think but I agree it would be nice to have *an additional` test in `test_notebooks` that runs an actual browser via selenium that would prove that the CSS is indeed loaded and applied to a given element (I don't actually know if selenium allows that but I would expect so).

edan-bainglass: Will add. Thanks

edan-bainglass:
@danielhollas proceeding with a made-up class for testing purposes only, as to not interfere with the code base for now.

tests/test_loaders.py
edan-bainglass: @superstar54 I'm not familiar with selenium. I see that you have to set a driver which wraps on an existing notebook. I then do `driver.find_element`, but does that only check THAT notebook? If so, which notebook do I use?

superstar54: I think you can create a new notebook (e.g., `test_css.ipynb`), and in one cell you add
```python
label = ipw.Label("Testing")
label.add_class("for-testing-purposes")
display(label)
```

in your test
```python
def test_load_css_stylesheet(selenium_driver):
    """Test `load_css_stylesheet` function."""
    driver = selenium_driver("notebooks/test_css.ipynb")
    driver.find_element(By.CLASS_NAME, "for-testing-purposes")
```
you may investigate how to check the style of the element, I don't know the how to do this exactly.

edan-bainglass: Sure, if you guys don't mind another notebook just for testing. Though maybe just call it `test.ipynb`? You never know what other tests might need a dedicated notebook🤷🏻‍♂️

danielhollas: I feel like there is a bit of misunderstanding here. The purpose of the notebook test is not to test the load_css_stylesheet on its own. It should be mainly test that the CSS style is actually applied. 

If you added the CSS code for the DictViewer, you could just extend the existing DictViewer test. As @superstar54 said, you'll need to investigate if selenium allows you to actually test CSS properties. If not, then adding a notebook test does not give us much.

The test itself shouldn't be using `load_css_stylesheet`, it should be loaded implicitly when awb is imported.

aiidalab_widgets_base/utils/loaders.py
superstar54: Thanks for the clarification! But it is still very weird to call, e.g., `aiida.orm`, a package.

tests/test_loaders.py
edan-bainglass: Take a look at the notebook test I pushed. It does exactly that - it checks that a property of a class is present on a widget. However, it does use `load_css_stylesheet` in the test. This would be redundant if awb is imported, but is it imported anywhere? I only imported it in the test to use `load_css_stylesheet`. If I remove the call, where would awb be called?

superstar54: Yes, in the test notebook, you need at least 
```python
import aiidalab_widgets_base
```

edan-bainglass: Why?

superstar54: I mean you only need to import `aiidalab_widgets_base`, since it will call `load_css_stylesheet` in the `__init__.py` file.

edan-bainglass: I've loaded `aiidalab_widgets_base` in the `test.ipynb` notebook with a comment regarding its use, as it is not actually used explicitly.

edan-bainglass: Okay, notebook test passes 👍

edan-bainglass:
I think this is good now? Failed tests unrelated I believe.

superstar54 (approved):
LGTM! Thanks for the nice work!

aiidalab_widgets_base/static/styles/test.css
```diff
@@ -0,0 +1,3 @@
+.for-testing-purposes {
```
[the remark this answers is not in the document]
edan-bainglass: Sure, but then I'd have to explicitly load it in the test notebook.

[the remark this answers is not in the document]
edan-bainglass: So the idea is to test that the loader works. But test_notebooks is not a package, so how would I load `test.css` from there?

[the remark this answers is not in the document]
edan-bainglass: I rather not mess with an internal component, lest I break it 😬 The style of `DictViewer` is assigned to `self.value`. `value` is an observed property, so not touching it, nor do I have time to understand what is happening in this code 🤷🏻‍♂️. But you're welcome to contribute to this PR with the proposed change 🙂

For now, I've modified the CSS loader utility and used it in the test notebook. I think this is acceptable, but again, if you feel strongly about it, see above comment.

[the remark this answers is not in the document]
danielhollas: Okay, fair enough, sorry for the back-and-forth. However, I really don't like the changes to the loader function. Feels like the complexity greatly increase only for the purposes of testing. :-(

aiidalab_widgets_base/utils/loaders.py
```diff
@@ -0,0 +1,52 @@
...
+from pathlib import Path
+
+from IPython.display import Javascript, display
+
+
+def load_css_stylesheet(
+    package: Package | None = None,
+    css_path: str | Path = "",
+    filename: str = "",
+):
```
danielhollas, line 13: Having both `package` and `css_path` makes this interface suuuuper messy.

How about we instead only have a single parameter of type `Path` that would be a directory with the css files. (maybe `css_dir` would be a better name, since `css_path` does not make it clear that it is not a path to a single CSS filename. 

It would then be the callers responsibility to convert Package to Path if needed.

@edan-bainglass does that make sense?

edan-bainglass: Sure

edan-bainglass: Hmm, on second thought, the only reason I introduced another use case for the API is due to the relocation of the `test.css` stylesheet outside of the package. So, as stated in the comment regarding `DictViewer`, you are welcome to modify the test to simply look at custom styles for the `DictViewer`, at which point, most of the complexity can be removed.

danielhollas: I will not have time to dig into into myself at this point, sorry. I think my suggestion makes sense regardless of the motivation imo. To me it makes more sense if the function accepts a Path to CSS dir instead of Package, it makes it more versatile, and it's users are free to decide where they want to put the css files (i.e. not necessarily  inside the module).

edan-bainglass: Understood. I will take a look at implementing it. Thanks 🙂

danielhollas: Thanks for bearing with me. Once we merge this I'll release a new alpha version. We can still iterate on it until the final `2.3.0` release.

edan-bainglass:
I'll fix the tests in the morning 😮‍💨

aiidalab_widgets_base/utils/loaders.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+from IPython.display import Javascript, display
+
+
+def load_css_stylesheet(css_dir: Path, filename: str = ""):
```
danielhollas, line 8: Thank you, this looks better! 

...now that we're passing a path, I am wondering, it would seem more natural to have a single parameter, that can be either a dir or a file. The split to two parameters for dir and file doesn't seem intuitive, and I don't see any advantage of doing it that way. 

(Sorry, I will shut up eventually : 😅) But perhaps good to iterate on this interface now rather to make it as intuitive as possible).

edan-bainglass: I'd prefer to keep the dir and filename separate. The reason for it is clear in the logic of the function.

danielhollas: Not sure if the function logic should define its interface. From a users perspective, passing a single path seems such simpler to think about (also it's not great that the two parameters have different type, another thing that can trip the user).

@superstar WDYT? 

(otherwise LGTM, thanks!)

danielhollas: Sorry, I guess I was replying to the older version of your comment. 

Yes, I would expect the the function can internally use Path.is_dir() to determine what to do.

Otherwise I am happy with the changes.

edan-bainglass: Your comments regarding the API are valid. In general, the user comes first (reasonably). I'll adjust the code in the morning. Thanks.

edan-bainglass: API now takes `css_path` of type `Path` or `str` (the latter converted to `Path`).

danielhollas: Looks great, thank you!

danielhollas (approved):
LGTM. :rocket: Thanks for bearing with me, I am super happy with the end result.

edan-bainglass:
@danielhollas what now? You merge? Also, docs build failed?

danielhollas:
@edan-bainglass if you have permissions to merge go ahead. If not let me know and I'll merge.

The doc failure is `

```
checking consistency... /home/docs/checkouts/readthedocs.org/user_builds/aiidalab-widgets-base/checkouts/624/docs/source/widget-list/notebooks/test_misc.ipynb: WARNING: document isn't included in any toctree
```

Please investigate how get rid of this warning or ignore it. I suspect the easiest solution is to move the notebook into the `notebook_tests` directory since it is not supposed to be in the docs.

edan-bainglass:
@danielhollas @superstar54 committed a bit of documentation. Please give it a looksee 🙏 Also, @danielhollas, the notebook tests are still failing 😔

danielhollas:
The tests are passing on another PR.

edan-bainglass:
Found some oversights in my code. Fixed it. Still not passing the tests. But its not just my test. @danielhollas @superstar54 could you please have a look?

danielhollas:
@edan-bainglass have you tried running the tests locally? That should make it clearer why they are failing.

edan-bainglass:
I couldn't figure out how to run selenium tests locally. I'll give it another go.

danielhollas:
@edan-bainglass congrats on merging this. :tada: I'll make a new alpha release that you can then use on PRs on QEApp. 

Could you please write here what you did to fix the tests and if / how you were able to test notebooks locally? I'll then add it to README for contributors. Thanks!

edan-bainglass:
@danielhollas thanks 🙏 Though I'm really not sure what actually fixed it. I was unable to run selenium tests locally in the container. I just applied the ol' step-by-step approach, walking back my commits, pushing chunks at a time and observing the tests. The CSS loading on import using the new utility passed the tests. Adding the notebook test in the notebooks folder with the test CSS as part of the module passed the tests. Then finally migrating the test notebook and CSS to the notebook tests folder... also passed the tests 🤷‍♂️ So in the end, I actually changed nothing from the state at which tests were failing. This was very confusing in general, as it was tests unrelated to my own that were failing before. Something to do with the selenium driver itself not finding the elements stated by the tests. According to [this article](https://reflect.run/articles/everything-you-need-to-know-about-nosuchelementexception-in-selenium/), "[NoSuchElementException](https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/NoSuchElementException.html) is one of the most common exceptions in Selenium WebDriver..."

Of course, happy to get another set of eyes on it. Maybe I missed something.

PR #630, opened 2024-08-23
Fix notebook tests

danielhollas:
The last successfully run of notebook tests is from July 8. There were no code changes that should impact the tests, we need to investigate what changed since then.

EDIT: It looks like just increasing the wait time tolerance fixes the issue. Perhaps the github runners just got slower? Not sure.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,12 +44,12 @@ dev =
     bumpver>=2023.1129
     pgtest~=1.3
     pre-commit>=3.5
-    pytest~=8.2.0
+    pytest~=8.3.0
     pytest-cov~=5.0
-    pytest-docker~=3.0
-    pytest-selenium~=4.1
-    pytest-timeout~=2.2
-    selenium==4.20.0
+    pytest-docker~=3.1.0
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -44,12 +44,12 @@ dev =
...
-    pytest~=8.2.0
+    pytest~=8.3.0
     pytest-cov~=5.0
-    pytest-docker~=3.0
-    pytest-selenium~=4.1
-    pytest-timeout~=2.2
-    selenium==4.20.0
+    pytest-docker~=3.1.0
+    pytest-selenium~=4.1.0
+    pytest-timeout~=2.3.0
```
danielhollas, line 51: To be on the safe side and to reduce potential churn, let's pin all test dependencies to minor versions.

danielhollas:
@edan-bainglass please take a look. Ideally we merge this first so other PRs can have tests passing.

danielhollas:
Going to land this now as I will be AFK for the next two hours.

edan-bainglass (approved):
Hopefully increasing the time is sufficient. LGTM!

PR #632, opened 2024-09-02
Discard module file from static styles folder

edan-bainglass:
Replaced module loading with using direct path in #624


edan-bainglass:
Interesting. So the `__init__.py` file is required in static/styles even though it is no longer used as a module. This is because without it, the styles directory does not transfer to site-packages on install, and hence is not found by the loader. Thanks @superstar54 for assisting in debugging this.

PR #637, opened 2024-10-08
Avoid black sphere in structure viewer

### LINKED ISSUE #633: black sphere for a primitive cell with single atom
Choose gold structure, and then convert to primitive cell → the atom turns black. It’s just a visualization issue, would be good to check if we can easily avoid it.

superstar54:
Fix #633 .
In the case of a structure with only one atom, the structure viewer shows a black sphere. As shown below:
Screenshot from 2024-10-08 13-35-31

This could be because of the overlap between the atoms and the control of the viewer. This PR implements a quick solution by shifting the center of the control in z direction by 1. Now the viewer shows:
Screenshot from 2024-10-08 13-32-03


danielhollas:
Just wondering, is this a bug that should be reported upstream? I.e. is this an nglview problem?

superstar54:
> Just wondering, is this a bug that should be reported upstream? I.e. is this an nglview problem?

I am not sure if it is a bug or not. The problem is that we call the `center` method after drawing the atoms and bonds. If we use the default setting, the nglview shows the correct atoms, but the view is not centered. 
Anyway, I opened an [issue](https://github.com/nglviewer/nglview/issues/1122) and waited for other's options.

danielhollas (approved):
Not a particularly pretty workaround, but if it works let's ship it :-)

PR #638, opened 2024-10-08
Allow user to hide the header of the Wizard App

superstar54:
We plan to remove the header in the QE App, as suggested by @giovannipizzi and @cpignedoli in https://github.com/aiidalab/aiidalab-qe/issues/813.

This PR allows the user to hide the header of the Wizard App.

### DIFF
--- a/aiidalab_widgets_base/wizard.py
+++ b/aiidalab_widgets_base/wizard.py
@@ -88,7 +88,7 @@ class WizardAppWidget(ipw.VBox):
 
     selected_index = tl.Int(allow_none=True)
 
-    def __init__(self, steps, hide_header=False, **kwargs):
+    def __init__(self, steps, show_header=True, **kwargs):
         # The number of steps must be greater than one
         # for this app's logic to make sense.
         if len(steps) < 2:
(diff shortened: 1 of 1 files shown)


aiidalab_widgets_base/wizard.py
```diff
@@ -142,11 +142,13 @@ def __init__(self, steps, **kwargs):
...
 
-        header = ipw.HBox(
+        self.header = ipw.HBox(
             children=[self.back_button, self.reset_button, self.next_button]
         )
 
-        super().__init__(children=[header, self.accordion], **kwargs)
+        if hide_header:
+            self.header.layout.display = "none"
+        super().__init__(children=[self.header, self.accordion], **kwargs)
```
AndresOrtegaGuerrero, line 151: Should we add a method to display or hide it as well ?

superstar54: Thanks for the suggestion. I updated the code, now the user can set
```python
# switch on the header
app.show_header=True
# switch off 
app.show_header=False
```

AndresOrtegaGuerrero:
this test is failing , tests/test_databases.py::test_cod_query_widget

superstar54:
> this test is failing , tests/test_databases.py::test_cod_query_widget

The failed test is related to the query of the Crystallography Open Database, Maybe some internet problem of the database. thus it is not related to this PR.

AndresOrtegaGuerrero:
Should the timeout be increased ? at least to 60 sec ?

AndresOrtegaGuerrero (approved):
LGTM!

danielhollas:
> It seems the timeout is not a problem, but the cod service is not reachable for the moment. http://www.crystallography.net/

@superstar54  could you open an issue to remove http requests in our test suite? I think we should mock them somehow so that we don't run into these transient issues.

superstar54:
Hi @danielhollas , the HTTP request is inside an AiiDA function. Do we really need to test it here?

danielhollas:
> Hi @danielhollas , the HTTP request is inside an AiiDA function. Do we really need to test it here?

I don't know, probably not, as long as we test the widget functionality in some way.

PR #644, opened 2024-11-22
Migrate to ipywidgets 8.x

### LINKED ISSUE #473: Support ipywidget 8.x
The ipywidget 8.x has been released in Aug 2022 and it provides a better user experience with more useful widgets. I don't know how many widgets in AWB will start malfunctioning after migration. As I know, some of our widget, the OPTIMADE client and bandsplot-widget is not working well with ipywidgets.

yakutovicha:
fixes #473

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -423,7 +423,6 @@ class StructureUploadWidget(ipw.VBox):
 
     def _on_file_upload(self, change=None):
         """When file upload button is pressed."""
-
         assert len(change["new"]) == 1, "Only single file upload is supported."
         file = change["new"][0]
         self.structure = self._read_structure(file["name"], file["content"])
--- a/tests/test_structures.py
+++ b/tests/test_structures.py
@@ -146,15 +146,16 @@ def test_structure_upload_widget():
     # Simulate the structure upload.
     widget._on_file_upload(
         change={
-            "new": {
-                "test.xyz": {
+            "new": (
+                {
+                    "name": "test.xyz",
                     "content": b"""2
 
                 Si 0.0 0.0 0.0
                 Si 0.5 0.5 0.5
                 """,
-                }
-            }
+                },
+            )
         }
     )
     assert isinstance(widget.structure, ase.Atoms)


danielhollas (commented):
Whoops, I had some comments during the hackathon that I forgot to submit.

yakutovicha:
I can conclude that `uv` is drunk today:
image

aiidalab_widgets_base/structures.py
```diff
@@ -118,7 +118,8 @@ def __init__(
             children=self._structure_importers(importers),
             selected_index=0 if importers else None,
         )
-        select_panel.set_title(0, "Select structure")
+        if importers:  # Otherwise ipywidgets 8.x throws an error when setting a title for non-existing tab.
+            select_panel.set_title(0, "Select structure")
```
edan-bainglass: Sure. But also https://github.com/MADICES/MADICES-2025/discussions/27. The API has changed. Unclear yet (need to test) if `titles` can be provided directly on `Tab` definition, but in any case, we can open a separate issue to update to the newer API of `.titles`.

aiidalab_widgets_base/wizard.py
```diff
@@ -148,6 +147,8 @@ def __init__(self, steps, show_header=True, **kwargs):
         self.show_header = show_header
 
         super().__init__(children=[self.header, self.accordion], **kwargs)
+        ipw.link((self.accordion, "selected_index"), (self, "selected_index"))
+        self.selected_index = 0
```
edan-bainglass: Seems unrelated. What was this about?

danielhollas:
Thanks a ton for working on this @yakutovicha. Before merging I'd like to test this on my app once I am back from holidays next week. (and obviously somebody should test on QeApp)

setup.cfg
```diff
@@ -28,8 +28,7 @@ install_requires =
     humanfriendly~=10.0
     ipytree~=0.2
     traitlets~=5.4
-    ipywidgets~=7.7
-    widgetsnbextension<3.6.3
```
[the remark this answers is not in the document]
yakutovicha: > See this issue [aiidalab/aiidalab-docker-stack#512](https://github.com/aiidalab/aiidalab-docker-stack/issues/512) Definitely will need testing.

I think this particular issue is related to the unpredictable behaviour of the `ipw.Output` widget. I replaced it with `ipw.VBox`, which is stable: https://github.com/aiidalab/aiidalab-qe/pull/1425

danielhollas (changes requested):
@yakutovicha thanks so much on working on this! I left a couple of minor comments.
I'll do some manual testing tomorrow (or later in the week depending on time).

aiidalab_widgets_base/wizard.py
yakutovicha: > Let the user choose if they wish to have it open.

done in https://github.com/aiidalab/aiidalab-widgets-base/pull/644/commits/8c7426d45cfefd8666b7e7efc5f9ea06f202c535 and https://github.com/aiidalab/aiidalab-widgets-base/pull/644/commits/c55e4def46d7a8efce130eb99fa290ff1f0e4a3d.

yakutovicha:
@danielhollas, @edan-bainglass, @cpignedoli, please be informed that this change is somewhat "contagious".

If your current version of ipywidgets is `7.7.x`, it will be updated to `7.8.x`. Unfortunately, there seems to be no way back from 7.8 to 7.7 (at least I didn't find). After you downgrade, the widgets won't appear any longer. 

So it seems like only `7.8` can be swapped with `8.x` back and forth without too much burden.

Bottom line is - don't test this change in a production environment.

aiidalab_widgets_base/wizard.py
```diff
@@ -101,8 +101,8 @@ def __init__(self, steps, show_header=True, **kwargs):
 
         # Initialize the accordion with the widgets ...
         self.accordion = ipw.Accordion(children=widgets)
+        print("After link")
```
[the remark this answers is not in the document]
yakutovicha: sorry

yakutovicha:
> @yakutovicha thanks, makes sense. Should this migration also be tied to the switch from Python 3.9 to 3.11 and to nbclassic? In terms of managing the docker images that might be the cleanest option, since users can then pin their images to `python-3.9.13` tag to ensure their production environments won't break once we release everything.

As far as I understand, they are not strictly dependent on each other - ipywidgets 8 can work in Python 3.9 without problems, so I am not entirely sure it is a good idea, as it will add another waiting factor. 

At the same time, AiiDA is dropping Python 3.9 in the next minor release (2.8.x), and we have the [aiida tag](https://github.com/aiidalab/aiidalab-docker-stack/pkgs/container/full-stack/478312909?tag=aiida-2.7.1) in our aiidalab-docker-containers containers. We can bind aiida and aiidalab Python migration. Shall we?

aiidalab_widgets_base/structures.py
yakutovicha: > The API has changed. Unclear yet (need to test) if `titles` can be provided directly on `Tab` definition, but in any case, we can open a separate issue to update to the newer API of `.titles`.

Since we are dropping ipywidgets 7.x, I implemented your request: 3b2856cfdca230b206fe33a31a08a7e051b065d6

edan-bainglass (commented):
Thanks @yakutovicha. Have a look at my comment regarding the file uploader.

PR #646, opened 2024-12-08
Make the setup new codes widget part of the code selection widget optional

edan-bainglass:
This PR provides a backwards compatible path to make the new code setup widget part of the code selection widget optional.

The PR also adds a flag (true by default) to control if the codes should be pre-fetched. This is useful, for example, in the QE app, where the MVC redesign (#802) provides these codes via models, removing the responsibility from the UI widgets.


superstar54 (changes requested):
Thanks for the work! 
Could you also add a test on the new optional parameters?

danielhollas (approved):
I've had only a very brief look and don't see anything unreasonable, thanks!

superstar54 (approved):
Thanks for adding the tests!

PR #647, opened 2024-12-11
Silence warning

edan-bainglass:
Recent developments in aiidalab-home of an external notebook to manage codes run into an issue with the AWB computational resources widgets when used without a default calcjob plugin. This PR silences a suspected false warning for now, to be properly investigated in a future PR.

Handling issue #648

### DIFF
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -1302,6 +1302,13 @@ class AiidaCodeSetup(ipw.VBox):
         for key, value in self.code_setup.items():
             if hasattr(self, key):
                 if key == "default_calc_job_plugin":
+                    if "None" in value:
+                        # NOTE: Using this widget through the `_ResourceSetupBaseWidget`
+                        # without an explicit `default_calc_job_plugin` causes `value`
+                        # to be "<plugin>.None", which is not a valid plugin name.
+                        # HACK to avoid the warning message
+                        # TODO see https://github.com/aiidalab/aiidalab-widgets-base/issues/648
+                        return
                     try:
                         self.default_calc_job_plugin.value = value
                     except tl.TraitError:


superstar54 (approved):
Could open an issue so that we don't forget to investigate it in the future?

aiidalab_widgets_base/computational_resources.py
```diff
@@ -1302,6 +1302,10 @@ def _observe_code_setup(self, _=None):
         for key, value in self.code_setup.items():
             if hasattr(self, key):
                 if key == "default_calc_job_plugin":
+                    if "None" in value:
```
danielhollas, line 1305: What is the type of `value`? Somewhat surprising if it would be list for tuple?

danielhollas (commented):
Just one question. Agreed about opening an issue and link to it in the TODO comment. 

Also can you please expand the PR desription with how the problem manifests exaxtly? Is there a warning in the UI? Does the code crash?

aiidalab_widgets_base/computational_resources.py
danielhollas: Ah, okay, that definitely needs an explanatory comment in the code, because `"None" in value` looks suuuper weird :smile:

edan-bainglass:
> Just one question. Agreed about opening an issue and link to it in the TODO comment.
> 
> Also can you please expand the PR desription with how the problem manifests exaxtly? Is there a warning in the UI? Does the code crash?

Referenced the issue, where the matter is described in detail. Added a reference in the TODO comment.

edan-bainglass:
@danielhollas Wow! I'm at a loss as to how changing the comment now causes a test to fail 🤦🏻‍♂️

danielhollas (approved):
Sure, go ahead! I am a bit concerned this hacky fix will bite us in the future unless we investigate it properly, but it's fine for now. :-) 

> @danielhollas Wow! I'm at a loss as to how changing the comment now causes a test to fail 🤦🏻‍♂️

Looks like a test flake, tests are passing after I rerun them. Could you open an issue about a flaky `test_cod_widget`, including the errors from the CI? That test is accessing an external database, we should mock that out.

Thanks!

edan-bainglass:
> Sure, go ahead! I am a bit concerned this hacky fix will bite us in the future unless we investigate it properly, but it's fine for now. :-)
> 
> > @danielhollas Wow! I'm at a loss as to how changing the comment now causes a test to fail 🤦🏻‍♂️
> 
> Looks like a test flake, tests are passing after I rerun them. Could you open an issue about a flaky `test_cod_widget`, including the errors from the CI? That test is accessing an external database, we should mock that out.
> 
> Thanks!

Indeed, very flaky, as I **also** re-ran the tests before making my previous comment, but they still failed 😢 Okay, opening an issue.

PR #650, opened 2024-12-22
Replace check of empty string in unfilled template widgets to `None`

### LINKED ISSUE #648: Entry point warning raised when using `_ResourceSetupBaseWidget` without a default calcjob plugin
Running the following in a notebook

```python
from aiidalab_widgets_base.computational_resources import _ResourceSetupBaseWidget

widget = _resourceSetupBaseWidget()
widget
```

and selecting a domain and a computer in the widget, then selecting the Quantum ESPRESSO code in the codes selector results in an entry point `quantumespresso.None`, which is clearly not installed, causing the widget to raise a warning.

Not sure at the moment how we end up with `None`.

This is the resultant value of of the `code_setup` trait in the `AiidaCodeSetup` widget (which is linked with the `filled_templates` trait of the `TemplateVariableWidget`, both widgets contained in the `_ResourceSetupBaseWidget`:
```python
{
    "label": "None-7.1",
    "description": "The code None of Quantum ESPRESSO compiled for daint-mc",
    "default_calc_job_plugin": "quantumespresso.None",
    "filepath_executable": "/apps/dom/UES/jenkins/7.0.UP03/21.09/dom-mc/software/QuantumESPRESSO/7.1-CrayIntel-21.09/bin/None.x",
    "prepend_text": "module load daint-mc\nmodule load QuantumESPRESSO\n",
    "append_text": "",
}
```
This suggests the guard should perhaps be moved up to the template widget, or higher. Pinging @unkcpz due to #511

### LINKED ISSUE #651: `None` value of unfilled template variables in code setup widget
Opening this issue as a reminder to investigate further what is currently being handled by #647, which provides a temporary hacky solution to #648.

edan-bainglass:
This PR corrects a guard for unfilled template variables which checks if the value of the corresponding widget is an empty string. Empty widgets may also be represented by `None`, especially dropdown selectors. This PR adds `None` to the check by comparing the widget's value to both - replaces `== ""` with `in ("", None)`.

Resolves #648
Resolves #651

@superstar54 @danielhollas have a look 🙏
@unkcpz tagging you due to git blame (suggests you implemented these sections)

### DIFF
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -1302,13 +1302,6 @@ class AiidaCodeSetup(ipw.VBox):
         for key, value in self.code_setup.items():
             if hasattr(self, key):
                 if key == "default_calc_job_plugin":
-                    if "None" in value:
-                        # NOTE: Using this widget through the `_ResourceSetupBaseWidget`
-                        # without an explicit `default_calc_job_plugin` causes `value`
-                        # to be "<plugin>.None", which is not a valid plugin name.
-                        # HACK to avoid the warning message
-                        # TODO see https://github.com/aiidalab/aiidalab-widgets-base/issues/648
-                        return
                     try:
                         self.default_calc_job_plugin.value = value
                     except tl.TraitError:
@@ -1597,7 +1590,7 @@ class TemplateVariablesWidget(ipw.VBox):
                 # if the widget is not filled, use the original template var string e.g. {{ var }}
                 inp_dict = {}
                 for _var in line.vars:
-                    if self._template_variables[_var].widget.value == "":
+                    if self._template_variables[_var].widget.value in ("", None):
                         variable_key = self._template_variables[_var].widget.description
                         self.unfilled_variables.append(variable_key.strip(":"))
                         inp_dict[_var] = "{{ " + _var + " }}"


danielhollas (commented):
LGTM but I am not familiar with the template code at all so this better be reviewed by somebody that does :-)

danielhollas:
Also thanks for following up on this @edan-bainglass! :love_you_gesture:

edan-bainglass:
> LGTM but I am not familiar with the template code at all so this better be reviewed by somebody that does :-)

100% why I tagged @unkcpz 😅

edan-bainglass:
@unkcpz can you have a look at this?

unkcpz (approved):
Thanks @edan-bainglass, I think it is a correct change, `None` was what I didn't considered when initialize the widget value. This will for sure directly solve https://github.com/aiidalab/aiidalab-widgets-base/issues/648 

Not sure it is easy to understand logic of testing, but when I created this widget, I tried to catch most of conditions with tests in https://github.com/aiidalab/aiidalab-widgets-base/blob/7a93035d6a25990628d0c5477984fa4c044be8e2/tests/test_computational_resources.py#L318-L372. Would you mind to give it a look to add a test for this `None` situation? I guess you can add https://github.com/aiidalab/aiidalab-widgets-base/issues/648 example as test and use the change of this PR to test it.

PR #653, opened 2024-12-23
Add volume information in the cell tab

### LINKED ISSUE #652: Add volume in Å^3 in the cell tab
Add the information in the last line of the widget.

Note: it should depending on the periodicity!
xyz -> Cell volume (ang^3)
xy -> Cell area (ang^2)
x -> Cell length (and)
(for 0D, show nothing!)

Note that StructureData already has a function to return both the dimensionality and the correct value (with the correct units): https://github.com/aiidateam/aiida-core/blob/779cc29d8a47eddabdf9b274d7fa711220ee1aa9/src/aiida/orm/nodes/data/structure.py#L772

Image

superstar54:
Fix #652 . 

Add volume information in the cell tab, the text depends on the periodic boundary condtion.
- xyz -> Cell volume (ang^3)
- xy -> Cell area (ang^2)
- x -> Cell length (and)

As suggested by @cpignedoli , we use the [function](https://github.com/aiidateam/aiida-core/blob/779cc29d8a47eddabdf9b274d7fa711220ee1aa9/src/aiida/orm/nodes/data/structure.py#L2412) from `orm.StructureData` directly. Note, the cell tab is inside class `_StructureDataBaseViewer`, which uses `ase.Atoms` to store the structure, so we can not use `StructureData.get_dimensionality` directly, but use the private function `_get_dimensionality`. Even though it's a private function, I decided to use it so we get updated in the future automatically and do not re-write duplicate code.

### DIFF
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -16,6 +16,7 @@ import spglib
 import traitlets as tl
 import vapory
 from aiida import cmdline, orm, tools
+from aiida.orm.nodes.data.structure import _get_dimensionality
 from ase.data import colors
 from IPython.display import clear_output, display
 from matplotlib.colors import to_rgb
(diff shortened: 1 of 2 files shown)


aiidalab_widgets_base/viewers.py
```diff
@@ -710,6 +711,20 @@ def _observe_cell(self, _=None):
...
             )
+            # Calculate the volume of the cell using the function from orm.StructureData
+            dimension_data = _get_dimensionality(self.structure.pbc, self.cell)
+            # Determine the label and unit based on dimensionality
+            cell_labels = {
+                1: ["length", "Å"],
+                2: ["area", "Å²"],
+                3: ["volume", "Å³"],
+            }
+            cell_label = cell_labels.get(dimension_data["dim"])
```
edan-bainglass, line 722: Consider defining `CELL_LABELS` outside to avoid redefining them

edan-bainglass (approved):
LGTM! Minor comment.

PR #660, opened 2025-01-08
Update uv in RTD config

danielhollas:
See also discussion in aiidateam/aiida-core#6685.
Includes a fix from aiidateam/aiida-core#6700

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -20,7 +20,7 @@ build:
         create_environment:
             - uv venv $READTHEDOCS_VIRTUALENV_PATH
         install:
-            - uv pip install .[docs] --prefix $READTHEDOCS_VIRTUALENV_PATH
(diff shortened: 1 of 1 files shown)


.readthedocs.yml
```diff
@@ -10,10 +10,17 @@ build:
...
-        - uv pip install .[docs]
-        - .venv/bin/python -m sphinx -W --keep-going -d _build/doctrees -D language=en -b html docs/source $READTHEDOCS_OUTPUT/html
+    jobs:
+        pre_create_environment:
+            - asdf plugin add uv
+            - asdf install uv 0.5.17
+            - asdf global uv 0.5.17
+        create_environment:
+            - uv venv $READTHEDOCS_VIRTUALENV_PATH
+        install:
```
danielhollas, line 20: Unlike with `uv sync` which we use in aiida-core, here we can manually specify the target virtual environment. This in turn let's us use the `sphinx` section below instead of specifying the build command manually.

PR #662, opened 2025-01-11
Tag structure manager and viewer with CSS classes for easy styling

edan-bainglass:


superstar54:
The class is added but not used. So, is it necessary to add the class here? The developer can add their class when using the widgets, right?

edan-bainglass:
> The class is added but not used. So, is it necessary to add the class (hard-coded) here? The developer can add their class when using the widgets, right?

Sure. But this way, they don't have to. We provide the widget with the class, so developers can just add styles in their stylesheets. Otherwise, let's say they use the structure manager in multiple places (as we do in aiidalab-qe), they'll have to tag each instances. Not ideal. Hence this PR.

superstar54 (approved):
Thanks for the explanation. LGTM!

PR #671, opened 2025-02-05
Remove sklearn dependency.

### LINKED ISSUE #590: Remove scikit-learn dependency
scikit-learn is only used for PCA to align molecular axes in SmilesWidget. Pulling such a heavy dependency just for this is an overkill. I am not sure if it is even still needed, and I am sure there must be less heavy-weight to do it.

https://stats.stackexchange.com/questions/235882/pca-in-numpy-and-sklearn-produces-different-results

Past PRs for context:
https://github.com/aiidalab/aiidalab-widgets-base/pull/160
https://github.com/aiidalab/aiidalab-widgets-base/pull/164

yakutovicha:
fixes #590 

Sklearn was only used to align SMILES-generated molecules with the Z-axis. The usefulness of such functionality is questionable, and it doesn't justify depending on such a heavy library.

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -751,11 +751,6 @@ class SmilesWidget(ipw.VBox):
 
     def _make_ase(self, species, positions, smiles):
         """Create ase Atoms object."""
-        from sklearn.decomposition import PCA
-
-        # Get the principal axes and realign the molecule along z-axis.
-        if len(species) > 2:
-            positions = PCA(n_components=3).fit_transform(positions)
         atoms = ase.Atoms(species, positions=positions, pbc=False)
         atoms.cell = np.ptp(atoms.positions, axis=0) + 10
         atoms.center()
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,7 +57,6 @@ eln =
     requests-cache~=1.0
 smiles =
     rdkit>=2021.09.2
-    scikit-learn~=1.1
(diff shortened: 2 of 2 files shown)


yakutovicha:
We've also discussed this with @cpignedoli (who originally introduced that functionality) and he agrees.

danielhollas (changes requested):
Looks good, thanks so much for doing this!

I think we should have a unit test that is sensitive the the generated coordinates from the smiles code. Could you maybe extend the existing tests/test_structure.py::test_smiles_widget() to assert the coordinates of some of the molecules? (perhaps the N2 is simplest). I would use the `np.allclose()` of comparisons to reference so that we're not sensitive to minor changes in the rdkit library.

danielhollas:
The Firefox UI tests failures are related to #672, no need to worry about that here.

yakutovicha:
> I think we should have a unit test that is sensitive the the generated coordinates from the smiles code.

Hey @danielhollas, I've been thinking about this. Are we sure this is needed?

Suppose the structure that comes out of rdkit is rotated - is it a problem? The tests will fail, but the only thing we will do to fix them is to change the tests.

It is not a problem to add such a check, I don't necessary see the benefits of doing that.

danielhollas:
Yeah, I don't feel strongly about it. It just feels weird that the change in this PR did not break any tests. Feel free to merge, we can revisit later.

danielhollas (approved):

PR #627, opened 2024-07-24
Unpin ASE, fix test

danielhollas:
fixes #612

PR #675, opened 2025-02-06
Use tmpdir for povray rendering

### LINKED ISSUE #676: StructureViewer: Race issue in povray rendering
I am pretty sure that the current povray rendering code does not support concurrent rendering of several structures (e.g. if user has multiple open windows) because it creates two non-unique temporary files in the current working directory so two running process can stomp on each other.

danielhollas:
~~Stacked on top of #627 which needs to be merged first.~~

Fixes #676

### DIFF
--- a/tests/test_viewers.py
+++ b/tests/test_viewers.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 import ase
 import pytest
 import traitlets as tl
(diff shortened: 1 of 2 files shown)


tests/test_viewers.py
```diff
@@ -107,6 +109,10 @@ def test_structure_data_viewer_storage(structure_data_object):
     # fmt: on
     v._render_structure()
 
+    # Make sure we don't polute current working dir with tempfiles
```
superstar54, line 112: Thanks for the fix, but is this necessary for the test to pass? If it pollutes the working dir, will the test fail?

danielhollas:
@superstar54 I've rebased this branch to get rid of the unrelated changes, please take a look.

superstar54 (approved):
LGTM!

PR #681, opened 2025-02-11
Fix issue with `StructureManager` when no editor is provided

yakutovicha:
fixes #680

PR #666, opened 2025-01-30
Clarify structure operations

edan-bainglass:
This PR aims to emphasize the distinction between visualizing the super cell and actually transforming it. It does so by (i) wrapping the viewer in its own accordion titled "View structure", and (ii) adding a comment below the super cell control in the appearance tab of the viewer clarifying usage.

Closes #667

PR #683, opened 2025-02-12
Wrap structure importers in accordion panel

edan-bainglass:
Following #666, I make structure selection consistent with viewing and editing, as well as guard against no importers following #681.

@giovannipizzi by request, I had wrapped the viewer in an accordion panel (opened by default) in #666. Here I'm extending it to the selectors. Question, should we keep this "Select structure" accordion panel open on start?

---

image

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -114,7 +114,19 @@ class StructureManagerWidget(ipw.VBox):
                 f"Unknown data format '{node_class}'. Options: {list(self.SUPPORTED_DATA_FORMATS.keys())}"
             )
 
-        viewer_panel = ipw.Accordion(
+        structure_importers = self._structure_importers(importers)
+
+        select_panel = ipw.Accordion(
+            children=[
+                *structure_importers,
+            ]
+            if structure_importers
+            else [],
+            selected_index=0 if structure_importers else None,
+        )
+        select_panel.set_title(0, "Select structure")
+
+        view_panel = ipw.Accordion(
             children=[
                 ipw.VBox(
                     children=[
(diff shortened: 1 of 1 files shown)


giovannipizzi:
Yes, I'd keep the select structure open by default, no need to keep it closed if the first thing you have to do is anyways clicking on it

edan-bainglass:
> Yes, I'd keep the select structure open by default, no need to keep it closed if the first thing you have to do is anyways clicking on it

Okay, great. Thanks. This then gives the user the ability to hide it, but otherwise does not change the natural flow of the app 👍

danielhollas:
@edan-bainglass can you write a bit more about the motivation behind the split? To me at least it visually looks a bit worse than before (too many borders, perhaps we could get rid of some of them?). At least in my app I don't see this as useful but maybe I am missing something.

edan-bainglass:
Less about how it looks, more about allowing users to collapse completed sections.

aiidalab_widgets_base/structures.py
```diff
@@ -114,7 +114,19 @@ def __init__(
...
+        structure_importers = self._structure_importers(importers)
+
+        select_panel = ipw.Accordion(
+            children=[
+                *structure_importers,
+            ]
+            if structure_importers
+            else [],
+            selected_index=0 if structure_importers else None,
+        )
```
yakutovicha, line 126: I think the logic here has already been taken care of by the `_structure_importers()`, so one can just do:
 ```suggestion
        select_panel = ipw.Accordion(
            children=self._structure_importers(importers)
        )
```

**Review by yakutovicha — CHANGES_REQUESTED** [review]
Hey @edan-bainglass, thanks, I like that it now becomes more uniform!

See my suggestions below.

File: aiidalab_widgets_base/structures.py [hunk 2]
```diff
@@ -170,11 +181,14 @@ def _structure_importers(self, importers):
...
 
+        if not importers:
+            return []
+
         # If there is only one importer - no need to make tabs.
         if len(importers) == 1:
             # Assigning a function which will be called when importer provides a structure.
             tl.dlink((importers[0], "structure"), (self, "input_structure"))
-            return importers[0]
+            return [importers[0]]
```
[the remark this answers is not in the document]
yakutovicha: I refer to the whole `if...` section:

```python
        # If there is only one importer - no need to make tabs.
        if len(importers) == 1:
            # Assigning a function which will be called when importer provides a structure.
            tl.dlink((importers[0], "structure"), (self, "input_structure"))
            return [importers[0]]
```

aiidalab_widgets_base/structures.py
edan-bainglass: Well, if no importers, you get back a Tab with no children, which is still a child, no? I mean, I guess this is okay. Technically if developers don't provide any children, then what are they even doing?

yakutovicha: if no importers, you get an empty list as far as I can see:

```python
        if not importers:
            return []
```

edan-bainglass: My point is that this lifts the logic up to the accordion, so no importers now mean an empty accordion, not an accordion with an empty Tab. Which do we prefer? Again, either case is nonesense from a development point of view.

edan-bainglass: Oh right, my bad. Missed that if statement. Okay then 👍

edan-bainglass: Keeping the `selected_index` conditional though, yeah?
```python
        select_panel = ipw.Accordion(
            children=self._structure_importers(importers),
            selected_index=0 if importers else None,
        )
        select_panel.set_title(0, "Select structure")
```

yakutovicha: You can, but this is not necessary. You are literally replicating the default behaviour.

yakutovicha (approved):
Grazie 🙏 !

PR #685, opened 2025-02-14
Implement `LoadingWidget`

edan-bainglass:
This was in the QE app, but is of general use. Moving to AWB.

PR #686, opened 2025-02-14
Implement caching in `AiidaNodeViewWidget`

edan-bainglass:
Moving node view caching mechanism implemented by @superstar54 in the QE app (https://github.com/aiidalab/aiidalab-qe/pull/921) into `AiidaNodeViewWidget`

### DIFF
--- a/aiidalab_widgets_base/static/styles/global.css
+++ b/aiidalab_widgets_base/static/styles/global.css
@@ -1,0 +1,7 @@
+.aiida-node-view-widget {
+  border: var(--jp-widgets-border-width) solid var(--jp-border-color1);
+  padding: 12px;
+  margin: 2px;
+  height: auto;
+  width: auto;
+}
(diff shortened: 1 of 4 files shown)


edan-bainglass:
@AndresOrtegaGuerrero this should resolve your issue

danielhollas:
> Moving caching mechanism implemented by @superstar54 in the QE app into the AiiDANodeViewWidget

Can you point to the original PR in QeApp. It's not super clear what is meant by caching in this context.

superstar54 (approved):
Nice!

superstar54:
Hi @danielhollas , It here: https://github.com/aiidalab/aiidalab-qe/pull/921

superstar54:
Hi @edan-bainglass , as suggested by @yakutovicha , could you add a test for the caching?

edan-bainglass:
Yep. I'll add a test for this and the loader from #685

danielhollas (commented):
I am a bit wary of the changes in this PR since the cache introduced here is unbounded. So for example for a large workchain, a user clicking around the node tree can blow up the memory of the Python process (I don't know if that is a real concern or not, but I assume some viewers may consume nontrivial amount of memory.

At the very least we need to provide a way to clear the cache, probably in the `reset` method. Or better yet, use one of pythons stdlib lru_cache decorators and wrap the logic in  a cached function call somehow.

PR #690, opened 2025-03-20
Fix private key handling logic

### LINKED ISSUE #689: Private key file possibly not handled correctly
A user was trying to setup SSH connection to a local cluster and ran into an issue when using the custom private key verification option:
Image

This is handled in the snippet below. The comments are a bit unclear. If the file exists (user already uploaded it to container?), we append a string to the filename and override current name? But the filename is then different so how is this overriding? Also, what if it doesn't exist? This is likely what triggered the error for the user, as in that case, there is no `private_key_fpath` variable to be used in the upload.

I think two things need to happen here:
1. Clean up code/comments, clarifying precisely what is the intended behavior
2. If the user is expected to first load the private key into the container (confusing, as we also do an upload here), then we need to improve documentation of this process. If this is not the expected of the user, then we need to handle the case where they don't, which presently leads to the above error.

```python
        # For not password ssh auth (such as using private_key or 2FA), key pair is not needed (2FA)
        # or the key pair is ready.
        # There are other mechanism to set up the ssh connection.
        # But we still need to write the ssh config to the ssh config file for such as
        # proxy jump.

        private_key_fname = None
        if self._verification_mode.value == "private_key":
            # Write private key in ~/.ssh/ and use the name of upload file,
            # if exist, generate random string and append to filename then override current name.

            # unwrap private key file and setting temporary private_key content
            private_key_fname, private_key_content = self._private_key
            if private_key_fname is None:  # check private key file
                message = "Please upload your private key file."
                self.message = wrap_message(message, MessageLevel.ERROR)
                return

            filename = Path(private_key_fname).name

            # if the private key filename is exist, generate random string and append to filename subfix
            # then override current name.
            if filename in [str(p.name) for p in Path(self._ssh_folder).iterdir()]:
                private_key_fpath = self._ssh_folder / f"{filename}-{shortuuid.uuid()}"

            self._add_private_key(private_key_fpath, private_key_content)

        # TODO(danielhollas): I am not sure this is correct. What if the user wants
        # to overwrite the private key? Or any other config? The configuration would never be written.
        # And the user is not notified that we did not write anything.
        # https://github.com/aiidalab/aiidalab-widgets-base/issues/516
        if not self._is_in_config():
            self._write_ssh_config(private_key_abs_fname=private_key_fname)
```

I purposely included the remaining code after the problem line as a reminder to give #516 another look.

@unkcpz appears to have touched this snippet last (#511). Maybe he can comment on this logic first. In the meantime, I'll check with the user to see if they pre-uploaded the key to the container, or maybe created it in the container, on any other information. Will post here shortly.

edan-bainglass:
This PR corrects the logic of handling custom private keys during SSH connection setup.

The issue was that `private_key_fpath` was only set if the condition below was true, not setting it for cases where the key does not exist. Surely I'm missing something, because that logic is counterintuitive to the expected use case (uploading a private key for the first time, i.e., key does not already exist). The code below ensures that both cases are handled.

```python
# if the private key filename is exist, generate random string and append to filename subfix
# then override current name.
if filename in [str(p.name) for p in Path(self._ssh_folder).iterdir()]:
    filename = f"{filename}-{shortuuid.uuid()}"

private_key_fpath = self._ssh_folder / filename

self._add_private_key(private_key_fpath, private_key_content)
```

Fixes #689

However, I suspect this is not the full story. Below this code section, there is a reference to #511

```python
# TODO(danielhollas): I am not sure this is correct. What if the user wants
# to overwrite the private key? Or any other config? The configuration would never be written.
# And the user is not notified that we did not write anything.
# https://github.com/aiidalab/aiidalab-widgets-base/issues/516
if not self._is_in_config():
    self._write_ssh_config(private_key_abs_fname=private_key_fname)
```

With the above fix, loading a private key for the first time works as expected. However, loading the same key again (not sure if this is a valid practice to support) does not pass the `is_in_config` check. This leads to the addition of the key, but not to the addition of the corresponding config section, effectively rendering the key useless w.r.t SSH connections. But again, I don't see why we would support this. I would opt for notifying the user that the key already exists.

But maybe @danielhollas can comment on his TODO comment 🙏
And maybe @unkcpz can say more about the original implementation and the intended use cases to support 🙏

### DIFF
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -20,6 +20,7 @@ from aiida.transports.plugins import ssh as aiida_ssh_plugin
 from humanfriendly import InvalidSize, parse_size
 from IPython.display import clear_output, display
 from jinja2 import meta as jinja2_meta
+from paramiko import SSHConfig
 
 from .databases import ComputationalResourcesDatabaseWidget
 from .utils import MessageLevel, StatusHTML, wrap_message
@@ -441,42 +442,70 @@ class SshComputerSetup(ipw.VBox):
         )
         return ret == 0
 
-    def _is_in_config(self):
-        """Check if the SSH config file contains host information."""
-        config_path = self._ssh_folder / "config"
-        if not config_path.exists():
-            return False
-        sshcfg = aiida_ssh_plugin.parse_sshconfig(self.hostname.value)
-        # NOTE: parse_sshconfig returns a dict with a hostname
-        # even if it is not in the config file.
-        # We require at least the user to be specified.
-        if "user" not in sshcfg:
-            return False
-        return True
+    def _write_ssh_config(self, private_key_name=None):
+        """Write the ssh config file."""
 
-    def _write_ssh_config(self, private_key_abs_fname=None):
-        """Put host information into the config file."""
-        config_path = self._ssh_folder / "config"
-
-        self.message = wrap_message(
-            f"Adding {self.hostname.value} section to {config_path}",
-            MessageLevel.SUCCESS,
-        )
-        with open(config_path, "a") as file:
-            file.write(f"Host {self.hostname.value}\n")
-            file.write(f"  User {self.username.value}\n")
-            file.write(f"  Port {self.port.value}\n")
-            if self.proxy_jump.value != "":
+        def write_entry_to_config_file(file, entry):
+            file.write(f"Host {entry['hostname']}\n")
+            file.write(f"  User {entry['user']}\n")
+            file.write(f"  Port {entry['port']}\n")
+            if entry.get("proxyjump", "") != "":
                 file.write(
-                    f"  ProxyJump {self.proxy_jump.value.format(username=self.username.value)}\n"
+                    f"  ProxyJump {entry['proxyjump'].format(username=entry['user'])}\n"
                 )
-            if self.proxy_command.value != "":
+            if entry.get("proxycommand", "") != "":
                 file.write(
-                    f"  ProxyCommand {self.proxy_command.value.format(username=self.username.value)}\n"
+                    f"  ProxyCommand {entry['proxycommand'].format(username=entry['user'])}\n"
                 )
-            if private_key_abs_fname:
-                file.write(f"  IdentityFile {private_key_abs_fname}\n")
+            if private_key_name:
+                file.write(f"  IdentityFile '{private_key_name}'\n")
             file.write("  ServerAliveInterval 5\n")
+
+        def add_host_to_ssh_config():
+            with open(config_path, "a") as file:
+                write_entry_to_config_file(file, entry)
+
+        def update_ssh_config():
+            with open(config_path, "w") as file:
+                for host in config.get_hostnames():
+                    if host == "*":
+                        continue
+                    write_entry_to_config_file(
+                        file,
+                        entry if host == self.hostname.value else config.lookup(host),
+                    )
+
+        config_path = self._ssh_folder / "config"
+        config = SSHConfig.from_path(config_path)
+
+        entry = {
+            "hostname": self.hostname.value,
+            "user": self.username.value,
+            "port": str(self.port.value),
+            "proxyjump": self.proxy_jump.value,
+            "proxycommand": self.proxy_command.value,
+            "identityfile": private_key_name,
+            "serveraliveinterval": "5",
+        }
+
+        # Check if the exact entry already exists
+        existing = config.lookup(self.hostname.value)
+
+        if all(existing.get(k) == v for k, v in entry.items() if v):
+            message = f"Identical {self.hostname.value} entry already exists"
+            self.message = wrap_message(message, MessageLevel.INFO)
+            return False
+
+        if self.hostname.value not in config.get_hostnames():
+            add_host_to_ssh_config()
+            message = f"Adding {self.hostname.value} section to {config_path}"
+        else:
+            update_ssh_config()
+            message = f"Updating {self.hostname.value} section in {config_path}"
+
+        self.message = wrap_message(message, MessageLevel.SUCCESS)
+
+        return True
 
     def key_pair_prepare(self):
         """Prepare key pair for the ssh connection."""
@@ -501,28 +530,8 @@ class SshComputerSetup(ipw.VBox):
 
     def _on_setup_ssh_button_pressed(self, _=None):
         """Setup ssh connection."""
-        if self._verification_mode.value == "password":
-            try:
-                self.key_pair_prepare()
-
-            except ValueError as exc:
-                self.message = wrap_message(str(exc), MessageLevel.ERROR)
-                return
-
-            self.thread_ssh_copy_id()
-
-        # For not password ssh auth (such as using private_key or 2FA), key pair is not needed (2FA)
-        # or the key pair is ready.
-        # There are other mechanism to set up the ssh connection.
-        # But we still need to write the ssh config to the ssh config file for such as
-        # proxy jump.
-
-        private_key_fname = None
+        private_key_fpath = None
         if self._verification_mode.value == "private_key":
-            # Write private key in ~/.ssh/ and use the name of upload file,
-            # if exist, generate random string and append to filename then override current name.
-
-            # unwrap private key file and setting temporary private_key content
             private_key_fname, private_key_content = self._private_key
             if private_key_fname is None:  # check private key file
                 message = "Please upload your private key file."
@@ -531,8 +540,7 @@ class SshComputerSetup(ipw.VBox):
 
             filename = Path(private_key_fname).name
 
-            # if the private key filename is exist, generate random string and append to filename subfix
-            # then override current name.
+            # If the private key exists, we append a random string to the filename
             if filename in [str(p.name) for p in Path(self._ssh_folder).iterdir()]:
                 filename = f"{filename}-{shortuuid.uuid()}"
 
@@ -540,12 +548,20 @@ class SshComputerSetup(ipw.VBox):
 
             self._add_private_key(private_key_fpath, private_key_content)
 
-        # TODO(danielhollas): I am not sure this is correct. What if the user wants
-        # to overwrite the private key? Or any other config? The configuration would never be written.
-        # And the user is not notified that we did not write anything.
-        # https://github.com/aiidalab/aiidalab-widgets-base/issues/516
-        if not self._is_in_config():
-            self._write_ssh_config(private_key_abs_fname=private_key_fname)
+        private_key_name = private_key_fpath.name if private_key_fpath else None
+        written = self._write_ssh_config(private_key_name)
+
+        # If using the password option and a new ssh config entry has been written,
+        # we prepare the ssh key pair and copy to remote computer.
+        if written and self._verification_mode.value == "password":
+            try:
+                self.key_pair_prepare()
+
+            except ValueError as exc:
+                self.message = wrap_message(str(exc), MessageLevel.ERROR)
+                return
+
+            self.thread_ssh_copy_id()
 
     def _ssh_copy_id(self):
         """Run the ssh-copy-id command and follow it until it is completed."""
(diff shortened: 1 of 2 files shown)


edan-bainglass:
Failed test seems unrelated

danielhollas:
> But maybe @danielhollas can comment on his TODO comment 🙏

I don't have any more context then what is written in the TODO and the issue, I was just passing by the code when I wrote that.

I do have a big concern over this feature though in general, it feels extremely dangerous to support this, private keys should really not be going over the network right? 

Shouldn't we instead support some kind of workflow that sets-up the public-private key locally in the container (if we don't already?)

edan-bainglass:
> > But maybe @danielhollas can comment on his TODO comment 🙏
> 
> I don't have any more context then what is written in the TODO and the issue, I was just passing by the code when I wrote that.
> 
> I do have a big concern over this feature though in general, it feels extremely dangerous to support this, private keys should really not be going over the network right?
> 
> Shouldn't we instead support some kind of workflow that sets-up the public-private key locally in the container (if we don't already?)

I did find this feature an odd choice and was unclear as to the reason for it. Intuitively, my first choice would be an automated key generation in the container. I'm not sure of the use case for a user to want to create their own key and upload it. @yakutovicha @unkcpz can you please comment on this widget design and the original thoughts behind the implementation? 🙏

edan-bainglass:
@unkcpz we could really use your comments on this part 🙏 just to make sure we're not undoing some prior reasoning.

edan-bainglass:
@danielhollas @yakutovicha I spoke to @unkcpz regarding this matter. He agrees that this custom private key upload feature is maybe not the best thing to support, though he admits that many do use it (though they probably should not). The container already has a key pair (id_rsa) available for copy (one of the options in widget) and deployment on the desired machine.

So, should we still support this? The fix in this PR does resolve the issue that was raised by our users, but on a more philosophical level w.r.t security and sensibility of using private keys instead of public, is this really something we should make available? Or do we discard it?

Please comment 🙏

yakutovicha:
@edan-bainglass and @danielhollas, I think we have a use case for that.

Although I agree with you that the general idea that the private key never leaves the place where it was created, some recent developments at CSCS made this case valid. Essentially, they offer keys of 1- or 7-day validity. The user is supposed to download the private key and place it on their machine (see [here](https://confluence.cscs.ch/spaces/KB/pages/794296407/Multi-Factor+Authentication+MFA#MultiFactorAuthentication(MFA)-UsingMFAtoSSHintosystems)).

I don't have enough security expertise to comment on the choice of CSCS, but from my experience, they handle security pretty well and I would trust their choices.

yakutovicha:
> I think the problem is that this is perhaps a specific case that only(?) applies to CSCS users.

Well, CSCS provides computational resources for the whole of Switzerland, so we will have enough use cases here. I wouldn't be surprised that other European providers come with similar solution, as they are talking to each other.


> I think it is best if this widget is accompanied by some instructions/recommendations?

I agree. We should mention something like this: "Use this option when you know what you are doing, as it poses additional security threats". We can also ask the question to CSCS so they can explain their choices - happy to do that.

Finally, please add the test for this behaviour so we don't need to fix it again in the future.

edan-bainglass:
> > I think it is best if this widget is accompanied by some instructions/recommendations?
> 
> I agree. We should mention something like this: "Use this option when you know what you are doing, as it poses additional security threats". We can also ask the question to CSCS so they can explain their choices - happy to do that.

Please do, so our instructions are clear and accurate.
 
> Finally, please add the test for this behaviour so we don't need to fix it again in the future.

Will do. Also, should we recommend a particular option, say, the "download public key" option? I'm assuming this option is to copy this key (that's already on the container) and deploy it on the desired machine.

danielhollas:
@yakutovicha the CSCS approach allows for a command-line access. I think instead of offering to upload SSH keys via GUI, we should just tell users to open the AiiDAlab terminal and go from there, no?

yakutovicha:
> @yakutovicha the CSCS approach allows for a command-line access. I think instead of offering to upload SSH keys via GUI, we should just tell users to open the AiiDAlab terminal and go from there, no?

@danielhollas, are you implying that the command line option is safer or is it a question of simplifying the setup?

In terms of security, both choices necessitate the keys to pass through the network, which leads me to believe that neither option is more secure than the other.

In my experience, users tend to avoid the command line whenever possible; that's why we do the setup in the GUI in the first place.

danielhollas:
> @danielhollas, are you implying that the command line option is safer or is it a question of simplifying the setup?

Yes, rather then downloading locally and then uploading via GUI where we don't offer any protection, I'd say that generating it locally in the container sounds better to me. In the GUI, we don't even have control whether we're connecting via HTTPS or plain HTTP, in which case we would be sending it completely unencrypted.

> In my experience, users tend to avoid the command line whenever possible; that's why we do the setup in the GUI in the first place.

For sure, but this feels like enough of a power user case that perhaps should necessitate this interaction?

yakutovicha:
> I'd say that generating it locally in the container sounds better to me.

The command-line tool does not generate the key pair; it only downloads them.

Regarding security, I suggest discussing this during today's AiiDAlab meeting to reach a consensus. Others may have insights to help determine whether this feature is worth the effort.

danielhollas:
> Regarding security, I suggest discussing this during today's AiiDAlab meeting to reach a consensus. Others may have insights to help determine whether this feature is worth the effort.

I don't have anything in my calendar, when is the meeting? (I am teaching between 14-16 so might not be able to join)

edan-bainglass:
> > Regarding security, I suggest discussing this during today's AiiDAlab meeting to reach a consensus. Others may have insights to help determine whether this feature is worth the effort.
> 
> I don't have anything in my calendar, when is the meeting? (I am teaching between 14-16 so might not be able to join)

Meeting's at 14 🙂

edan-bainglass:
@danielhollas @yakutovicha @superstar54 can we approve this PR? My argument is that this feature exists. This PR simply fixes a bug in it. As to the matter of removing the feature, or adjusting it somehow, that is to be raised in another issue/PR.

danielhollas:
> This PR simply fixes a bug in it. As to the matter of removing the feature, or adjusting it somehow, that is to be raised in another issue/PR.

Can you please create an issue so we don't forget?

> With the above fix, loading a private key for the first time works as expected. However, loading the same key again (not sure if this is a valid practice to support) does not pass the is_in_config check. This leads to the addition of the key, but not to the addition of the corresponding config section, effectively rendering the key useless w.r.t SSH connections. But again, I don't see why we would support this. I would opt for notifying the user that the key already exists.

I feel like this should be addressed somehow in this PR, otherwise this is gonna be extremely hard to debug for users.

aiidalab_widgets_base/computational_resources.py
```diff
@@ -502,7 +502,9 @@ def _on_setup_ssh_button_pressed(self, _=None):
             # if the private key filename is exist, generate random string and append to filename subfix
             # then override current name.
```
danielhollas, line 503:
```suggestion
            # If the private key filename already exists, generate a random string and append it as a suffix
```

edan-bainglass: I think this old documentation provided by @unkcpz when originally implemented needs to be reconsidered anyhow.

edan-bainglass:
> > This PR simply fixes a bug in it. As to the matter of removing the feature, or adjusting it somehow, that is to be raised in another issue/PR.
> 
> Can you please create an issue so we don't forget?

For sure!

> > With the above fix, loading a private key for the first time works as expected. However, loading the same key again (not sure if this is a valid practice to support) does not pass the is_in_config check. This leads to the addition of the key, but not to the addition of the corresponding config section, effectively rendering the key useless w.r.t SSH connections. But again, I don't see why we would support this. I would opt for notifying the user that the key already exists.
> 
> I feel like this should be addressed somehow in this PR, otherwise this is gonna be extremely hard to debug for users.

Thanks for the reminder. I will address this. in the PR.

yakutovicha:
> @danielhollas @yakutovicha @superstar54 can we approve this PR? My argument is that this feature exists. This PR simply fixes a bug in it. As to the matter of removing the feature, or adjusting it somehow, that is to be raised in another issue/PR.

No blocking from my side.


On a side note, I wrote a question to CSCS to better understand their security considerations. Here is the full text:


Dear CSCS Support,

We have been discussing the security implications of transferring private keys over the network [1]. It is generally advised that private keys remain on the system where they were created for security reasons (this is widely discussed, see e.g., [2]).

In the case of CSCS, the MFA setup requires users to download the private key. I assume this approach was carefully considered from a security perspective. Given that this method was chosen, I would be interested in understanding its rationale. Is this deemed acceptable due to the short-lived nature of the keys? Are other supercomputing centers adopting similar practices?

I would greatly appreciate any insights you can share, including any input directly on the discussion at [1].

Thank you for your time.

Best regards,
Sasha

[1] https://github.com/aiidalab/aiidalab-widgets-base/pull/690#issuecomment-2757360139

[2] https://www.digicert.com/blog/why-private-keys-should-not-be-shared

edan-bainglass:
@danielhollas @yakutovicha this PR has evolved quite a bit beyond the initial plan! I did some refactoring of the SSH config writing part. Have a look. The previous issues have been resolved. Bonus - you can now edit existing entries. The matter of the private key requires more discussion. Later!

danielhollas:
@edan-bainglass thanks for your work on this! Just a heads up that I am currently on a conference and will be on short holidays next week so will not be able to review this closely at this time.

Just a quick note that given the amount of change this should add some tests, at the very least a regression test for the bug that this is fixing, and for the editing the existing entries. 

(btw: the existing tests are currently failing)

yakutovicha (changes requested):
Thanks, @edan-bainglass.

The code seems to function flawlessly, though I have to admit I didn't go deep into detail. At the same time, I've noticed two things:

1. The private key alone doesn't work. It needs to be accompanied by a public one (it seems that we've opened a can of worms here 😅).
2. The private key filename is better to put in single quotes. I had downloaded the key a second time, so it got auto-renamed to `cscs-key (2)`. After uploading, SSH didn't like the name without the quotes.

Finally, CSCS guys haven't yet replied.

yakutovicha:
Got a reply from CSCS, they indeed carefully considered the security implications of transferring the private keys through the network. To mitigate the risks, they have implemented several security measures:

* User Authentication: CSCS ensure robust user authentication, including Multi-Factor Authentication (MFA).
* Activity Monitoring: Activities related to the key retrieval process are monitored and logged.
* Data Encryption: CSCS guarantee data encryption in transit (between the user and CSCS) and recommends their users do it at rest (under the user’s responsibility, as the key will be stored at the user's device).

danielhollas:
> CSCS guarantee data encryption in transit (between the user and CSCS) 

Worth noting that we don't guarantee this with AiiDAlab (i.e. this depends on the details of each deployment, right?)

unkcpz:
my two cents.
The CSCS problem is orthogonal to the problem relate to this PR. I believe AiiDAlab should never ask for to upload private key. 
CSCS's reply means nothing to me it is more like they find excuse for their unprofessional, but whatever.
AiiDAlab run in container, it can generate key pairs these is unavoidable truth. We should not promise and guarantee anything except the container is open source and use standard base image. 

So, I think we should remove upload private key functionality. We provide CSCS MFA app for user to interact with CSCS MFA.

yakutovicha:
> my two cents. The CSCS problem is orthogonal to the problem relate to this PR. I believe AiiDAlab should never ask for to upload private key. CSCS's reply means nothing to me it is more like they find excuse for their unprofessional, but whatever. AiiDAlab run in container, it runs a os, it can generate key pairs, these is unavoidable truth. We should not promise and guarantee anything else except the container build file is open source and use standard base image (from jupyter community).
> 
> So, I think we should remove upload private key functionality. We provide CSCS MFA app for user to interact with CSCS MFA.

yakutovicha:
Sorry, clicked the wrong button 😅 and mistakenly closed the pull request. Reopened.

edan-bainglass:
@yakutovicha I tend to agree with @unkcpz on this matter. I understand that CSCS made a choice regarding its operations, but I'm not sure it is in our best interest to accommodate it. I am unfamiliar with the MFA app, but if it indeed provides a path to authenticate and connect to CSCS machines, do we need the other approach. Even if it is simpler, perhaps it is best we stick to proper standards and avoid any perceived hacky means of transport.

@danielhollas opinions?

danielhollas:
I'd vote to remove this. The widgets in this repo are meant to be for general usage, and uploading private keys is ill-advised in general, and we can't provide the guarantees that the CSCS mentions.

Users can anyway upload files via the File Manager interface if they need that. Ideally we'd provide a separate app (similar to the MFA app that @yakutovicha made) if we need to provide nicer UI.

yakutovicha:
> Apologies for the delay on this. @yakutovicha comments? Would you be okay with removing it from AWB?

Yes, go ahead. It seems most people are in favour of removing this option. While I would personally prefer to keep it, I understand the advantages of dropping it: less code (and thus fewer potential issues) and a clear signal against questionable key-handling practices.

edan-bainglass:
@unkcpz @danielhollas @yakutovicha per the discussion, the private key feature has been removed. Please review 🙏

PR #692, opened 2025-04-02
Fix password input

### LINKED ISSUE #691: Can not setup computer using `password` mode
Here is the example, I chose the `Passwrod` mode, and clicked the `setup ssh`, but it did not ask me to input the password, and in the next step, the computer test failed.

Image

superstar54:
Fix #691 .
As a user, I want to set up my own HPC, since the HPC is not in the domain, so I clicked the detailed step, but I can not input the password. Here is the screenshot of the bug:

image

Here is a quick fix:

image

### DIFF
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -1858,7 +1858,7 @@ class ResourceSetupBaseWidget(ipw.VBox):
         if change["new"]:
             self.detailed_setup_widget.layout.display = "block"
             self.quick_setup_button.disabled = True
-            #reset and trigger the verification mode to update the password box.
+            # reset and trigger the verification mode to update the password box.
             self.ssh_computer_setup._verification_mode.value = "private_key"
             self.ssh_computer_setup._verification_mode.value = "password"
             # fill the template variables with the default values or the filled values.
--- a/aiidalab_widgets_base/computational_resources.py
+++ b/aiidalab_widgets_base/computational_resources.py
@@ -1792,6 +1792,12 @@ class ResourceSetupBaseWidget(ipw.VBox):
                 detailed_setup,
             ]
         )
+        # hide this conatainer if the detailed setup is enabled.
+        self.password_box_container = ipw.VBox(
+            children=[
+                self.ssh_computer_setup.password_box,
+            ]
+        )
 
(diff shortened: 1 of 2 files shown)


superstar54:
Also ping @unkcpz , in case you are interested in looking at this PR and have any comments.

edan-bainglass (approved):
Appreciate the emojis! I trust you tested this. LGTM!

superstar54:
Thanks for opening the issue for the failed tests.

> I think this should be fixed, seems really confusing for users to have that field there.

I fixed, as shown below:

Screenshot from 2025-04-03 11-29-48

superstar54:
Hi @danielhollas , would you like to have a look at this PR again? Or I can merge it?

aiidalab_widgets_base/computational_resources.py
```diff
@@ -1787,6 +1792,12 @@ def __init__(
                 detailed_setup,
             ]
         )
+        # hide this conatainer if the detailed setup is enabled.
```
danielhollas, line 1795: I don't understand how this comment relates to the code below, perhaps it should be moved somewhere else?
```suggestion
        # hide this container if the detailed setup is enabled.
```

danielhollas (approved):
Looks good, just one question about a comment. 
Thanks for expanding the tests!

aiidalab_widgets_base/computational_resources.py
superstar54: Thanks for pointing out this. I updated it to
```
# This container is used to control the appearance of the password box
```

edan-bainglass:
@superstar54 I noticed the following comment from Jason that may be related to this PR:

```python
        self._ssh_password = ipw.Password(
            description="password:",
            disabled=False,
            layout=LAYOUT,
            style=STYLE,
        )
        # Don't show the continue button until it ask for password the
        # second time, which happened when the proxy jump is set. The
        # first time it ask for password is for the jump host.
        self._continue_with_password_button = ipw.Button(
            description="Continue",
            layout={"width": "100px", "display": "none"},
        )
```

Seems like the button was hidden for a reason?

superstar54:
The bug is not the `continue_with_password_button`, but the box to input the password, as shown in the screenshot.

PR #695, opened 2025-04-04
Document the SSH setup widget

edan-bainglass:
In this PR, I provide some documentation for the SSH setup widget. I also disable the `Setup ssh` button unless both hostname and username are provided.


superstar54 (approved):
For me, it's already very good to have this simple guide for the user. Of course, in the future, we need to write more detailed instruction, may be in the online docs.

PR #698, opened 2025-05-06
Test aiida 2.7

danielhollas:
Both notebook tests and unit tests are updated to use aiida-core 2.7.1.

### DIFF
--- a/tests/test_computational_resources.py
+++ b/tests/test_computational_resources.py
@@ -572,9 +572,9 @@ def test_resource_setup_widget_for_password_configure(monkeypatch, tmp_path):
 
     # Test select a new resource setup will update the output interface (e.g. ssh_config, computer_setup, code_setup)
     # and the computer/code setup widget will be updated accordingly.
-    w.comp_resources_database.domain_selector.value = "merlin.psi.ch"
+    w.comp_resources_database.domain_selector.value = "merlin7.psi.ch"
     w.comp_resources_database.computer_selector.value = "cpu"
-    w.comp_resources_database.code_selector.value = "QuantumESPRESSO-7.0"
+    w.comp_resources_database.code_selector.value = "QuantumESPRESSO-7.4"
 
     # Fill in the computer name and trigger the setup button again, the message should be updated.
     for (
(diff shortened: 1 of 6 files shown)


agoscinski:
Some of the errors I think are related to the usage `tempfile.mkdtemp`, I at least remember that I also had issues with this. That it was already removed when I created it, so I switched to the pytest fixture or `tempfile.TemporaryDirectory`. It might be related due to an update in the dependencies of aiida-core.

danielhollas:
I think it has something to do with the profile management. The actual error from the `export_aiida_subgraph` is comes from a failing `verdi archive create` command, which complains that a profile is not set. 
```
Command '['verdi', 'archive', 'create', '/tmp/pytest-of-runner/pytest-0/test_export_button_widget0/export.aiida', '-N', '5']' returned non-zero exit status 1.
b"Critical: no default profile defined: None\n{'CONFIG_VERSION': {'CURRENT': 9, 'OLDEST_COMPATIBLE': 9}, 'profiles': {'909900f38064e37d3f242f6336098653': {'storage': {'backend': 'core.sqlite_dos', 'config': {'filepath': '/tmp/pytest-of-runner/pytest-0/test_sqlite_dos_storage0'}}, 'process_control': {'backend': 'core.rabbitmq', 'config': {'broker_protocol': 'amqp', 'broker_username': 'guest', 'broker_password': 'guest', 'broker_host': '127.0.0.1', 'broker_port': 5672, 'broker_virtual_host': ''}}, 'test_profile': True, 'PROFILE_UUID': '53e42cace0c14a31816a52dbd0a11632', 'options': {}, 'default_user_email': 'test@localhost'}}}\n"

```

tests/test_computational_resources.py
```diff
@@ -572,9 +572,9 @@ def test_resource_setup_widget_for_password_configure(monkeypatch, tmp_path):
 
     # Test select a new resource setup will update the output interface (e.g. ssh_config, computer_setup, code_setup)
     # and the computer/code setup widget will be updated accordingly.
-    w.comp_resources_database.domain_selector.value = "merlin.psi.ch"
```
danielhollas, line 575: This was super-painful to fix (these tests now fail on master as well). Really these tests should be rewritten to use mocked-out data instead of real data that can change at any moment.

danielhollas: See also this note. 

https://github.com/aiidalab/aiidalab-widgets-base/pull/698/files#diff-71fe007c958acd11b69e1011aaaa2116a3a7f62a2e5ef8ec3743a8d191dcec88R49

superstar54 (approved):
LGTM!

PR #710, opened 2025-10-08
Update import from ELN

yakutovicha:
* Remove node viewer from eln_import notebook
* Add Empa surfaces app for the geometry optimization.
* Display apps where the structure is imported only when "The object is ready" button is pressed.
* ElnImportWidget - implement "ready" trait.


danielhollas:
Eln functionality has been removed from AWB so closing

PR #712, opened 2025-11-05
added PK structure uploader

cpignedoli:
When a user would like to start a simulation from a previous one, having the possibility to enter directly the PK of the input structure is a convenient shortcut compared to searching in a long list of AiiDA nodes


danielhollas:
@cpignedoli thank you, I've often needed this. Would it make sense to integrate this into the existing `StructureBrowserWidget`? That way this improvement is readily available to every app that uses it. I am not sure if it makes sense for this to be a fully standalone widget.

cpignedoli:
Hi @danielhollas, this would perfectly fine from my side if nobody is against. Also I was kind of in a hurry and I excluded the case the node is a cif we could have your suggestion also including what i left behind. Would you have time to complete the PR as you suggest?

cpignedoli:
Thanks for the reviews and sorry for reviving this old PR a bit noisily.

I updated the branch against current `master` and cleaned up the implementation. The direct-PK loader now:

- accepts a structure PK directly from the widget,
- keeps the dropdown selection and `structure` trait in sync,
- respects the configured query types (`StructureData`, `CifData`, etc.),
- clears the current selection on invalid/missing/unsupported PKs,
- has focused test coverage.

A couple of CI fixes were needed because current `master` now tests Python 3.12 and lowest-direct dependencies. The final non-feature fix is only test isolation: `test_open_aiida_node_in_app_widget` was hitting the live AiiDAlab app registry, so I mocked the app icon there. No runtime dependency workaround was kept for that.

All CI checks are now green, and the PR remains compatible with Python 3.9 (`python_requires >=3.9`). Could one of you please have a quick look and approve if this looks reasonable?

aiidalab_widgets_base/structures.py
```diff
@@ -569,6 +569,16 @@ def disable_drop_label(change):
         )
         self.end_date_widget = ipw.Text(value="", description="To: ")
 
+        self.info = ipw.HTML("")
+        self.pk_input = ipw.Text(
+            value="",
+            placeholder="1617",
```
danielhollas: Putting a random number here seems confusing. I'd prefer to leave this empty I think, the description should make it clear.

danielhollas:
@cpignedoli thank you for fixing CI, I've pulled the fixes out to a separate PR #743. Can you please merge in master to this branch and revert any remaining dependency changes? They should no longer be needed.

> The final non-feature fix is only test isolation: test_open_aiida_node_in_app_widget was hitting the live AiiDAlab app registry, so I mocked the app icon there. No runtime dependency workaround was kept for that.

Thanks! This is great.

aiidalab_widgets_base/structures.py
```diff
@@ -678,18 +696,64 @@ def search(self, _=None):
...
         options = [(f"Select a Structure ({len(matches)} found)", False)]
         for mch in matches:
-            label = f"PK: {mch.pk}"
-            label += " | " + mch.ctime.strftime("%Y-%m-%d %H:%M")
-            label += " | " + mch.base.extras.get("formula", "")
-            label += " | " + mch.node_type.split(".")[-2]
-            label += " | " + mch.label
-            label += " | " + mch.description
-            options.append((label, mch))
+            options.append((self._format_node_label(mch), mch))
```
danielhollas, line 699: Thanks for this refactor!

danielhollas (commented):
Thanks @cpignedoli for picking this up. Just had a very brief look and haven't seen anything concerning from my side, would prefer if @yakutovicha could do some review and testing. 

@cpignedoli can you include some screenshot of the current UI in the PR description?

aiidalab_widgets_base/structures.py
cpignedoli: Hi @danielhollas here is how it would look like:
- Nothing selected:
image
- Selection from the list of structures triggers population of the PK widget
image
- Selection fro the PK number triggers same selection in the structure list:
image
- Selection of wrong PK triggers error message, here we have a refresh problem in the visualizer that does not depend on this PR and I will open a new PR for that
image

cpignedoli: @danielhollas do we merge?

cpignedoli: @danielhollas I forgot a coment of yours above, I check it then I would merge within today

cpignedoli: @danielhollas, everything is in place. @yakutovicha is in vacation, if you want to merge please go ahead; otherwise, please approve and I will merge
Next, I will fix the bug of bonds that are not deleted

aiidalab_widgets_base/structures.py
```diff
@@ -678,18 +696,64 @@ def search(self, _=None):
...
+        try:
+            pk_value = int(self.pk_input.value)
+            if pk_value <= 0:
+                self._clear_structure_selection()
+                self.info.value = "Invalid PK: please enter a positive integer."
+                pk_value = None
+        except (ValueError, TypeError):
+            self._clear_structure_selection()
+            self.info.value = "Invalid PK: please enter a positive integer."
+            pk_value = None
```
danielhollas: I'd simplify this a bit and use the early-return pattern so that we can reduce the indendation in the following code.

```suggestion
        try:
            if pk_value := int(self.pk_input.value) <= 0:
                raise ValueError
        except ValueError:
            self._clear_structure_selection()
            self.info.value = "Invalid PK: please enter a positive integer."
            return
```

aiidalab_widgets_base/structures.py
```diff
@@ -678,18 +696,64 @@ def search(self, _=None):
...
+                    self.structure = node
+
+                else:
+                    self._clear_structure_selection()
+                    self.info.value = (
+                        f"The PK does not correspond to {self._query_type_names()}."
+                    )
+            except common.NotExistent:
+                self._clear_structure_selection()
+                self.info.value = f"No AiiDA node found for PK={pk_value}."
```
danielhollas: Might be better to use the else clause here
```suggestion
        try:
            node = orm.load_node(pk_value)
        except common.NotExistent:
            self._clear_structure_selection()
            self.info.value = f"No AiiDA node found for PK={pk_value}."
        else:
            if isinstance(node, self.query_structure_type):
                self.results.options = [(self._format_node_label(node), node)]
                self.results.value = node
                self.structure = node
            else:
                self._clear_structure_selection()
                self.info.value = (
                    f"The PK does not correspond to {self._query_type_names()}."
                )
```

danielhollas (approved):
Thanks @cpignedoli, I have some suggestion for simplification but after that feel free to merge.

This widget could use a UI refresh I think (cc @edan-bainglass) but that can be done separately.

PR #714, opened 2025-11-25
CI: Use uv to test with lowest supported dependency versions

danielhollas:
Instead of specifying aiida-core version explicitly when running unit tests, we utilize the uv's `--resolution` parameter to tell it whether we want to test we highest or lowest versions of dependencies. This will be helpful for #644 so we can test both ipywidgets 7.x and 8.x

See more details in the corresponding change in aiida-core: https://github.com/aiidateam/aiida-core/pull/6940

To make this work I needed to bump minimum versions of some packages, most notably aiida-core from 2.1 to 2.2 (see comment below why this was necessary). We could do a different workaround, but I think it is fair to drop support for 2.1 at this point. :-) The corresponding `full-stack:aiida-2.1.2` image has been published 3 years ago and hopefully nobody is using it.


setup.cfg
```diff
@@ -20,18 +20,18 @@ classifiers =
...
-    bokeh~=2.0
+    bokeh~=2.4
     humanfriendly~=10.0
     ipytree~=0.2
     traitlets~=5.4
     ipywidgets~=7.7
     widgetsnbextension<3.6.3
     pymysql~=0.9
-    nglview~=3.0
+    nglview~=3.0,>=3.0.8
```
danielhollas: Earlier patch versions failed to build on python 3.9: https://github.com/aiidalab/aiidalab-widgets-base/actions/runs/19681981549/job/56377980039

setup.cfg
```diff
@@ -44,7 +44,7 @@ zip_safe = False
 dev =
     bumpver>=2023.1129
     pgtest~=1.3
-    pre-commit>=3.5
+    pre-commit>=4.0
```
danielhollas: I think this version bump in the end was not needed, but probably doesn't hurt. Happy to revert.
(I still maintain we should not be installing pre-commit into every venv, it should instead be installed globally via `pipx` or `uv tool install`)

yakutovicha (approved):
I like the change a lot, and in general, it is cool that such a "resolution" feature is available.

PR #720, opened 2025-12-04
Remove process notebooks

### LINKED ISSUE #719: Remove process notebooks from AWB
Since the 3.0 release, the package won't be an aiidalab application anymore. That is why notebooks are no longer needed here.

yakutovicha:
fixes #719

In this PR, I remove process-related notebooks and the accompanying widgets that are exclusively used in those notebooks. The list of removed widgets and corresponding tests is below:

* `ProcessCallStackWidget`
* `ProcessInputsWidget`
* `ProcessListWidget`
* `ProcessOutputsWidget`
* `ProcessReportWidget`
* `ProgressBarWidget`
* `RunningCalcJobOutputWidget`

related to https://github.com/aiidalab/aiidalab-home/pull/211


danielhollas (approved):
Thanks, code looks good, but please don't merge yet before we release 2.6 and before we have a chance to discuss a bit more about the coupling between AWB and home app (which I think we should avoid if possible, let's discuss over at aiidalab-home repo)

danielhollas (commented):
Some more thoughts about the links to the notebooks moved to home app. 
This can also be deferred for future PR, but we should definitely think about this a bit more before the final AWB 3.0 release (e.g. do we want to perhaps get rid of some of the widgets if they are gonna live in the home app?).

edan-bainglass (changes requested):
Thanks @yakutovicha . LGTM! Regarding the links, let's allow the user to inject them via widget constructors as to avoid hard wiring them, to home app or any other place. Happy to discuss 🙂

edan-bainglass (approved):
@yakutovicha LGTM! Though I must admit, it is hard to verify that all is well and sane with these sort of moves. Good to have a document somewhere to describe the overall goal. If you have such a doc/issue, please link here 🙂🙏

danielhollas (changes requested):
Just a quick question / note. I think we should verify we actually want to remove all of these widgets, are they really not used in any apps? (I'll check atmospec, not sure about QeApp or others). Maybe we should just deprecate these instead of removing them?

danielhollas (approved):
Thanks @yakutovicha, let's try it out. Is there any more planned work on AWB side regarding python 3.12 migration? If not, I'd suggest we make an alpha release of 3.0 for easier testing.

PR #730, opened 2026-03-02
CI: Test python 3.12, bump minimum aiida-core version

danielhollas:
- Run unit tests with python 3.12, since that's the version that we want to upgrade to.
- Bump minimum aiida-core version. The next release of AWB is going to be 3.0 and that is never going to run in an image with aiida-core <2.7. So we can safely bump the minimum version here.
- Bump minimum version of pandas and ansi2html to resolve test failures.

I needed to disable Optimade tests due to #731.


.github/workflows/ci.yml
```diff
@@ -71,7 +71,7 @@ jobs:
               if: matrix.browser == 'Firefox'
 
             - name: Run pytest
-              run: pytest -v --driver ${{ matrix.browser }} tests_notebooks
+              run: pytest --tb=line -v --driver ${{ matrix.browser }} tests_notebooks
```
danielhollas: Selenium tracebacks are worse than useless, `--tb=line` results in much smaller output when a test fails

danielhollas:
@edan-bainglass if you're okay with this PR let's merge this so we can continue testing in #732

edan-bainglass (approved):
@danielhollas I approve to keep things going, but let's please continue to discuss, so I can get a clearer picture 🙂

PR #737, opened 2026-03-16
Get rid of pandas dependency

### LINKED ISSUE #591: Remove pandas dependency
We should use a more lighter-weight library to generate HTML tables. (we don't do anything else with pandas.

yakutovicha:
fixes #591

### DIFF
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -4,10 +4,13 @@ from __future__ import annotations
 
 import base64
 import copy
+import csv
+import io
 import os
 import pathlib
 import re
 import warnings
+from html import escape
 
 import ase
 import ipywidgets as ipw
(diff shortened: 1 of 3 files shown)


danielhollas (commented):
Thanks for doing this.

I am really not stoked about the hand-written HTML and CSS. Can we try to use the [tabulate](https://github.com/astanin/python-tabulate) package? This is a dependency that is used heavily in aiida-core and supports html output so should hopefully be easy to try out.

yakutovicha:
> @yakutovicha could you post in the description an example of a table before/after this PR? 🙏

A small fix was required in [e6e7b1b](https://github.com/aiidalab/aiidalab-widgets-base/pull/737/commits/e6e7b1b0c9360049c5973344e0d158defa69bdf8)

Here we go:

before:
image

after:

image

edan-bainglass (approved):
LGTM! Thanks @yakutovicha.

Unrelated note for the future, regardless of which tech stack we use, great if we can avoid 1000+ line files 😵 Maybe better if we follow:
```
viewers/
    __init__.py
    utils.py  # helper functions
    this_viewer.py
    that_viewer.py
    ...
```
Just a thought 🙂

danielhollas (approved):
Perfect, thanks @yakutovicha!

PR #739, opened 2026-04-08
Don't test with Python 3.11

danielhollas:
We're jumping all the way to 3.12 so there's no point in running unit tests with Python 3.11


edan-bainglass (approved):
LGTM!

PR #740, opened 2026-04-08
Pin numpy to 1.x due to bokeh incompatibility

danielhollas:
The currently pinned bokeh version is incompatible with numpy 2.x. Since aiida-core 2.8 started supporting numpy 2.x, widget tests using bokeh started failing.

We will also need to pin numpy in the image based on aiida-core 2.8 in order not to break older apps.

See #733 for more details

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,6 +25,7 @@ install_requires =
     ansi2html~=1.7
     ase~=3.18
     bokeh~=2.4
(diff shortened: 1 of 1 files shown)


edan-bainglass (approved):
Fine for now. See comments on #733. Hopefully the referenced team members can have a look.

PR #741, opened 2026-04-09
Bump version v2.5.0 -> v3.0.0a0

danielhollas:
Let's release a 3 alpha version for an easier testing with the nbclassic image

(needs to be merged with a merged commit!)

### DIFF
--- a/aiidalab_widgets_base/__init__.py
+++ b/aiidalab_widgets_base/__init__.py
@@ -117,4 +117,4 @@ __all__ = [
     "viewer",
 ]
 
-__version__ = "2.5.0"
+__version__ = "3.0.0a0"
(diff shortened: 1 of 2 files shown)


edan-bainglass (approved):
👍 but why merge commit?

danielhollas:
@edan-bainglass good question, tl;dr is that the commit in this PR is already tagged, so squashing or rebasing would create a new commit and we would lose the tag (or rather the tag would stay on github but would point to a commit that's not on a main branch)

I ran bumpver as 

```
bumpver update  --major --tag alpha
```

The current release process is described on the wiki: https://github.com/aiidalab/aiidalab/wiki/Development-guide#releasing-a-new-version

We should make it less complicated, happy to discuss at some point when you have time.

danielhollas:
New release is now available :rocket: 

https://github.com/aiidalab/aiidalab-widgets-base/releases/tag/v3.0.0a0

PR #742, opened 2026-04-20
Bump nglview version

danielhollas:
Let's see if nglview 4.0 is working

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ install_requires =
     traitlets~=5.4
     ipywidgets~=8.1
     pymysql~=0.9
-    nglview~=3.0,>=3.0.8
+    nglview>=3.0.8,<5
     spglib>=1.14,<3
     vapory~=0.1.2
     pandas~=2.2


danielhollas:
@cpignedoli since you're now doing a lot of work on the viewer, it would be super helpful if you could test that nglview v4 works. FWIW I've looked at the changelog and it doesn't seem to contain any breaking changes, but I might have missed something (and I haven't yet tested this at all)

cpignedoli:
@danielhollas sure, I will, in case it does not work in my present docker I will wait for Sasha for an update one. let me try!!

danielhollas:
Yeah, in your existing container, running `pip install -U --user nglview` should do the trick I think.

danielhollas:
@cpignedoli in case you want to help test the new python 3.12 image (which would be very helpful!) you can pull in this image: `ghcr.io/aiidalab/full-stack:pr-479`. It's ready for testing, barring some minor issues, see my last comment:
https://github.com/aiidalab/aiidalab-docker-stack/pull/479#issuecomment-4359743191

cpignedoli:
@danielhollas yes!! this is exactly what I was nearby to ask you, super now I aiidalab-launch with this image and test.
Coudl be I am done for today but will be back on this tomorrow!!

cpignedoli:
image
my beautiful axes and large view work with 4.0.1 The pan will need to be adjusted but I can take care, and switch back my PRs for axes and viewr controls to draft

cpignedoli:
@danielhollas I could not test this with the QE app, but I expect the behavior to be the same I observed in the CP2K apps.

From my side, this PR should be merged.

I still have some unrelated issues when upgrading packages. For example, I had to revert to:

```text
requests==2.32.5
requests-cache==1.3.1
cattrs==25.2.0```

cpignedoli (approved):
I tested it on my local AiiDAlab instance on a Mac arm64, ghcr.io/aiidalab/full-stack:pr-479, nglview 4.0.1, Empa surfaces app. To me, everything works

danielhollas:
@cpignedoli thanks so much for testing! It's great that you've been able to test this with the new image. I haven't even realized that nglview v4 requires ipywidgets8 so it can't be tested in the old image. Going to merge this and hopefully we'll get more testing soon, once AtmoSpec and QeApp works in the new image.

PR #743, opened 2026-05-16
Fix CI: Pin RabbitMQ version and update requests-cache

danielhollas:
Two fixes here, originally identified by @cpignedoli in #712

1.  It looks like RMQ v4.3 broke AiiDA so pin to v4.2 for now. I'll open an issue on aiida-core.

2. `requests=2.34` release broke `requests-cache` package. The fix was release in v1.3.2, so we set it as minimum version. See https://github.com/requests-cache/requests-cache/issues/1151 for details 
(`requests-cache` is only needed by the `eln` extras, which should hopefully be removed soon, before we release v3.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,7 +96,7 @@ jobs:
         timeout-minutes: 30
         services:
             rabbitmq:
-                image: rabbitmq:latest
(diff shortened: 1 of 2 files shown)


edan-bainglass (approved):
👍👍

PR #746, opened 2026-05-19
Clear stale bond shapes from structure viewer

cpignedoli:
## Summary

Fix stale bond cylinders remaining in `StructureDataViewer` after the selected structure is cleared.

## Root cause

Bonds are drawn with `nglview._add_shape(...)`, which creates a separate NGL component in addition to the structure component. The previous cleanup removed only `component_0`, so clearing the viewer removed the atoms but left the bond shape component behind.

## Changes

- Update `remove_viewer_components()` to remove every NGL component from the viewer.
- Add a regression test that renders a structure with a bond shape component, clears the structure, and verifies all NGL components are removed.

## Validation

- Direct component check: rendering creates `['Structure', 'nglview.shape.Shape']`; clearing leaves zero components.
- `python -m compileall aiidalab_widgets_base/viewers.py tests/test_viewers.py`
- `python -m ruff check aiidalab_widgets_base/viewers.py tests/test_viewers.py`

A focused local `pytest` run is blocked in this container before reaching these tests because the AiiDA PostgreSQL fixture expects the system `locate` command to find `pg_ctl`.

### DIFF
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -1050,9 +1050,10 @@ class _StructureDataBaseViewer(ipw.VBox):
 
     def remove_viewer_components(self, c=None):
         """Remove all components from the viewer except the one specified."""
-        if hasattr(self._viewer, "component_0"):
-            self._viewer.component_0.clear_representations()
-            cid = self._viewer.component_0.id
+        keep_id = getattr(c, "id", c)
+        for cid in list(self._viewer._ngl_component_ids):
+            if cid == keep_id:
+                continue
             self._viewer.remove_component(cid)
 
     @tl.default("supercell")
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -1048,12 +1048,9 @@ class _StructureDataBaseViewer(ipw.VBox):
                     kwargs=params["params"],
                 )
 
-    def remove_viewer_components(self, c=None):
-        """Remove all components from the viewer except the one specified."""
-        keep_id = getattr(c, "id", c)
+    def remove_viewer_components(self):
+        """Remove all components from the viewer."""
         for cid in list(self._viewer._ngl_component_ids):
-            if cid == keep_id:
-                continue
             self._viewer.remove_component(cid)
 
     @tl.default("supercell")
--- a/tests/test_viewers.py
+++ b/tests/test_viewers.py
@@ -292,7 +292,6 @@ def test_structure_data_viewer_clears_bond_shape_components():
 
     viewer.structure = water
     assert len(viewer._viewer._ngl_component_ids) == 2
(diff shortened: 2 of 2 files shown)


danielhollas:
@cpignedoli can you please provide steps how to reproduce this problem, perhaps with a screenshot how it looks? I'd like to test locally but am unclear about how.

cpignedoli:
here we go @danielhollas 
https://github.com/user-attachments/assets/14df1c97-7dbd-459f-9724-2b44b4d0dfe3

tests/test_viewers.py
```diff
@@ -279,6 +279,28 @@ def test_structure_data_viewer_representation(structure_data_object):
...
+            (0.0, 0.0, 0.119262),
+            (0.0, 0.763239, -0.477047),
+            (0.0, -0.763239, -0.477047),
+        ],
+    )
+    viewer = viewers.StructureDataViewer()
+
+    viewer.structure = water
+    assert len(viewer._viewer._ngl_component_ids) == 2
+    assert "nglview.shape.Shape" in viewer._viewer._ngl_component_names
```
danielhollas: I am not sure what this assert is testing, but it seems like it's way to implementation-specific, I'd remove it.

```suggestion
```

aiidalab_widgets_base/viewers.py
```diff
@@ -1050,9 +1050,10 @@ def highlight_atoms(
 
     def remove_viewer_components(self, c=None):
         """Remove all components from the viewer except the one specified."""
-        if hasattr(self._viewer, "component_0"):
-            self._viewer.component_0.clear_representations()
-            cid = self._viewer.component_0.id
+        keep_id = getattr(c, "id", c)
```
danielhollas: I am not sure what this line is doing, but the only caller of this function doesn't pass any `c` argument, so let's remove it, unless you already have a PR that needs it. If you do, at least please rename the argument to something more description, it's not clear what its type is supposed to be.

danielhollas (changes requested):
Thanks for the screencast, very helpful!

danielhollas (approved):
Thanks, LGTM! (I haven't done any actual manual testing but I trust that you did :-) )

aiidalab_widgets_base/viewers.py
cpignedoli: Good point, thanks. The optional `c` argument was a defensive/generalized version of the cleanup, but we do not currently need to preserve a component here and no caller passes it. I simplified `remove_viewer_components()` to remove all components and dropped the unused argument.

tests/test_viewers.py
cpignedoli: Agreed, that assertion was too tied to nglview internals. The test still verifies the behavior we care about: rendering creates multiple components, and clearing the structure removes all of them. I removed the component-name assertion.

cpignedoli:
seems to work, I merge
https://github.com/user-attachments/assets/1ddf6107-ef7b-443d-85b5-0503891f95f1

PR #748, opened 2026-05-19
tests: use sqlite-based test fixtures from aiida-core

danielhollas:
Switch from deprecated `aiida.manage.tests.pytest_fixtures` pytest plugin to `aiida.tools.pytest_fixtures` which are based on SQLite DB backend so we no longer need Postgresql to run the tests. Currently, we still need the RabbitMQ service running though.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,7 +10,28 @@ import numpy as np
 import pytest
 from aiida import engine, orm, plugins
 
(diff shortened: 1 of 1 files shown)


danielhollas:
@cpignedoli this should make running tests locally simpler.

edan-bainglass (approved):
LGTM!

PR #750, opened 2026-05-21
Bump version v3.0.0a0 -> v3.0.0a1

danielhollas:
Let's make another alpha version, there's been a few commits since the last one, but most notably I want the nglview v4 bump, which should resolve the integration test failures in aiidalab/aiidalab-docker-stack#479 (don't ask why, the reason is too annoying to write down here :anguished: )


edan-bainglass (approved):
LGTM!

PR #751, opened 2026-05-21
Remove requests-cache hack in elns.py

danielhollas:
The issue that the hack was solving has long been fixed in aiidalab/aiidalab#447 so we can remove it now and get rid of the `requests-cache` dependency.


danielhollas:
Huh, interesting that the tests are failing (though not surprising), we need to bump up `requests-cache` version in `aiidalab` package and release a new version.

EDIT: Solved by bumping minimum `aiidalab` package version

cpignedoli:
Hi @danielhollas, to be sure on this, I will countercheck with @fabioacl

cpignedoli (approved):
I tested this PR together with @fabioacl using the aiidalab-openbis app in a real AiiDAlab instance.

Environment:
- AiiDAlab: 26.5.2
- Python: 3.12.11
- aiida-core: 2.8.0
- ipywidgets: 8.1.8
- aiidalab-widgets-base: PR #751 editable checkout
- aiidalab-openbis: editable checkout

The widgets-base PR was installed from the PR branch and confirmed to resolve from the checkout path. The aiidalab-openbis app imports and runs with this setup. We also verified that `pip check` reports no broken requirements.

The remaining issues encountered during testing were app-side compatibility/setup issues in aiidalab-openbis, not caused by this PR, and are being addressed separately.

I approve this PR.

danielhollas:
@fabioacl @cpignedoli thanks so much for detailed testing!

Just so you are aware, I believe after the latest discussion with @yakutovicha the plan was to remove the ELN code from AWB to somewhere else (possibly `aiidalab-eln`?), since the ELN functionality is not general enough for AWB, and removing it is in line with the plan of turning AWB into just a library and removing it from AiiDAlab registry (which will coincide with the v3 release and python 3.12 image). 

Please let me know if you have any concerns about this plan.

cpignedoli:
Hi @danielhollas I see no problems , we just have to keep in mind that this will affect a app used heavily by our experimentalists so we just have to be informed in advance before any transition

PR #752, opened 2026-05-21
tests: Simplify generate_calc_job_node fixture

danielhollas:
This fixture has clearly been copied over from
aiida-quantumespresso plugin and even has some qe-specific and imports which we definitely don't need!


edan-bainglass (approved):
LGTM!

PR #754, opened 2026-05-31
Fix AiidaNodeViewer not resetting when assigned None

danielhollas:
I noticed this issue when trying to upgrade AtmoSpec to AWB 2.4.0.

Turns out #686 subtly broke the `AiidaNodeViewWidget` is it was not resetting when the `node` traitlet was assigned None. See the added assertion in the unit test which is failing on current master.

### DIFF
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -76,8 +76,15 @@ class AiidaNodeViewWidget(ipw.VBox):
 
     @tl.observe("node")
     def _observe_node(self, change):
-        if not ((node := change["new"]) and node != change["old"]):
+        node = change["new"]
+        if node == change["old"]:
             return
+        if not node:
+            with self._output:
+                clear_output()
+            self.children = []
+            return
+
         if node.uuid in self.node_views:
             self.children = [self.node_views[node.uuid]]
             return
--- a/tests/test_viewers.py
+++ b/tests/test_viewers.py
@@ -361,6 +361,9 @@ def test_node_view_caching():
     node = orm.Int(1)
     node_view.node = node
     viewer = node_view.children[0]
+
     node_view.node = None
+    assert not node_view.children
+
     node_view.node = node
     assert node_view.children[0] is viewer
--- a/tests/test_viewers.py
+++ b/tests/test_viewers.py
@@ -358,12 +358,19 @@ def test_node_view_for_non_widget_viewer():
 def test_node_view_caching():
     """Test that providing a given node a second time returns the cached viewer."""
     node_view = viewers.AiidaNodeViewWidget()
-    node = orm.Int(1)
+    node = orm.Dict()
     node_view.node = node
     viewer = node_view.children[0]
+    assert len(node_view.node_views) == 1
+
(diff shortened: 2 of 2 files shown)


aiidalab_widgets_base/viewers.py
```diff
@@ -76,8 +76,15 @@ def __init__(self, **kwargs):
 
     @tl.observe("node")
     def _observe_node(self, change):
-        if not ((node := change["new"]) and node != change["old"]):
+        node = change["new"]
+        if node == change["old"]:
             return
+        if not node:
+            with self._output:
```
danielhollas, line 83: It might be also a good idea in this case to clear out the `self.node_views` cache, since currently it just grows indefinitely.

danielhollas:
Hmm, based on code coverage I don't think the pre-existing unit test actually tests the viewer caching.

image


EDIT: I think I fixed the test

aiidalab_widgets_base/viewers.py
edan-bainglass: So if I understand you correctly, you would only keep the cache alive while moving between nodes? The moment `None` comes through (maybe better to be explicit that it is `None`?), you're proposing to reset along with the view also the cache? If so, that would defeat a good chunk of the cache benefit, no? Are you experiencing a memory issue here?

I'm also not clear on how `None` arrives here? Can you clarify?

edan-bainglass (commented):
I left a comment for clarification. Regarding the test, looks like the main change is to test `Dict` for real viewer caching, as `Int` does not have a dedicated viewer. But then how did this pass before? And can you clarify what you're doing with the `Int` part? What is going to `stdout`?

danielhollas:
> Regarding the test, looks like the main change is to test Dict for real viewer caching, as Int does not have a dedicated viewer. 

Yep!

> But then how did this pass before? And can you clarify what you're doing with the Int part?

It was passing tautologically, since Nodes without their viewer are displayed directly in a dedicated `ipw.Output` object which is always the same. 

> What is going to stdout?

That's the `__str__` (or `__repr__`, not sure) representation of the node. See comment in the `test_node_view_for_non_widget_viewer` test.

aiidalab_widgets_base/viewers.py
danielhollas: > I'm also not clear on how None arrives here? Can you clarify?

This is perhaps related to my app, where I have a button that hides the tree (and the tree is not displayed by default). Not sure how things are handled in QeApp. 

>  Are you experiencing a memory issue here?

I haven't done any memory benchmarking, but as a point of principle, unbounded caches are bad. :-) If the user clicks around enough and keeps the tab open, the widgets will be kept in the cache, and essential you'll have memory leak. (you can imagine e.g. user loading different workflows in QeApp, and the cache would still keep the nodes from a previous workflow).

Anyway, let's not block this bugfix PR which is about something else, I'll open a dedicated issue where I can explain the use case in my app in more detail.

edan-bainglass (approved):
Thanks @danielhollas. LGTM! Let's continue discussion w.r.t the cache on a separate issue.

PR #756, opened 2026-06-01
Make StructureUploadWidget resilient towards invalid files

### LINKED ISSUE #755: StructureUploadWidget is not resilient towards invalid files
I've found at least three different ways to crash the app by trying to upload an invalid file.

The stacktraces are available here:
ispg-group/aiidalab-ispg#276
ispg-group/aiidalab-ispg#277
ispg-group/aiidalab-ispg#278
ispg-group/aiidalab-ispg#279

I'll make a PR shortly.

The previous attempt to fix this was #339

danielhollas:
Fixes #755

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -448,12 +448,7 @@ class StructureUploadWidget(ipw.VBox):
                     structures = get_ase_from_file(temp_file.name)
             except ValueError as e:
                 self._status_message.message = f"""
-                    <div class="alert alert-danger">ERROR: {e}</div>
-                    """
-                return None
-            except KeyError:
-                self._status_message.message = f"""
-                    <div class="alert alert-danger">ERROR: Could not parse file {fname}</div>
+                    <div class="alert alert-danger">ERROR: Could not parse file '{fname}': {e}</div>
                     """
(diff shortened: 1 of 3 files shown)


aiidalab_widgets_base/structures.py
```diff
@@ -448,12 +448,7 @@ def _read_structure(self, fname, content):
                     structures = get_ase_from_file(temp_file.name)
             except ValueError as e:
```
yakutovicha, line 449: Shouldn't it be, or, in other words, why "KeyError" was excluded?
```suggestion
            except (KeyError, ValueError) as e:
```

yakutovicha (commented):
little question

aiidalab_widgets_base/structures.py
danielhollas: Good Q. I think originally KeyError here wass catching some obscure edge condition in `ase.io.read.`. But now we're catching all exceptions in `get_ase_from_file` and converting them to `ValueError` we don't have to catch `KeyError` here anymore.

yakutovicha: Ah, right, ok then!

yakutovicha (approved):

PR #761, opened 2026-06-03
Remove OpenAiidaNodeInAppWidget

### LINKED ISSUE #760: Remove OpenAiidaNodeInAppWidget
This widget represent a weird circular dependency between AWB and a hardcoded list of apps (QeApp, LSMO app). It also couples AWB to `aiidalab` up. I think v3 is a perfect opportunity to remove it.

danielhollas:
Closes #760.

Hopefully the diff itself makes it clear why we want to do this. :sweat_smile:


setup.cfg
```diff
@@ -35,6 +34,7 @@ install_requires =
     spglib>=1.14,<3
     vapory~=0.1.2
     ipython>=7.33,<9.0
+    pip>=25
```
edan-bainglass, line 37: Why?

edan-bainglass (approved):
Thanks @danielhollas. LGTM! Can you quickly say what the widget was used for? What does it mean to "open an AiiDA node in app"?

setup.cfg
danielhollas: There is a pip invocation in the bug report widget.

danielhollas:
That's a question for @yakutovicha , I've never used it

yakutovicha:
@edan-bainglass, this widget was used in conjunction with the ELN widget. When an object is imported from an ELN, it first goes into an intermediate notebook (`notebooks/eln_import.ipynb`). The `OpenAiidaNodeInAppWidget` then handles opening that item in its respective app.

I agree that the current implementation is bad. We chose to hardcode the paths and URL parameters because there was (and still is) no way to retrieve that information dynamically from the apps.

yakutovicha (approved):

PR #767, opened 2026-06-24
Center SMILES molecules after adding auxiliary cell

cpignedoli:
## Summary

This PR restores the intended SMILES import behavior around auxiliary cells:

- when `add_auxiliary_cell=True` (the default), the auxiliary cell is added first and the generated molecule is centered inside it;
- when `add_auxiliary_cell=False`, no auxiliary cell is added and the RDKit coordinates are left untouched.

## Root cause

PR #592 moved the auxiliary-cell creation out of `_make_ase()`, but `_make_ase()` still called `atoms.center()` before any real cell was available. As a result, the molecule was centered in ASE's zero/default cell and then the auxiliary cell was added afterward, leaving the molecule near the cell corner.

## Validation

- `python -m pytest tests/test_structures.py::test_smiles_widget tests/test_structures.py::test_smiles_widget_without_cell tests/test_structures.py::test_smiles_widget_make_ase_preserves_positions`
- `python -m pytest tests/test_structures.py`

Implemented with Codex.

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -796,7 +796,6 @@ class SmilesWidget(ipw.VBox):
     def _make_ase(self, species, positions, smiles):
         """Create ase Atoms object."""
         atoms = ase.Atoms(species, positions=positions, pbc=False)
-        atoms.center()
         # We're attaching this info so that it
         # can be later stored as an extra on AiiDA Structure node.
         atoms.info["smiles"] = smiles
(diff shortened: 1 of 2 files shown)


cpignedoli:
Hi @danielhollas std for me when I create a molecule with cell is to have the molecule centered. Can you check this PR?

danielhollas (approved):
Thanks, this makes sense, sorry about the omission.

PR #769, opened 2026-06-25
Handle viewer representations through extras

cpignedoli:
## What

PR-R2 in the representation split. This is stacked on top of #768 (`feature/encoded-representation-style-ids`).

This PR teaches the structure viewer/manager to read viewer representation masks from AiiDA node extras and to clean stale representation rows when switching between structures.

## Why

Encoded representation arrays already survive file upload/download, but stored AiiDA nodes need an extras-based path because `get_ase()` does not preserve custom viewer state by itself.

This PR intentionally only reads/restores that state. Writing it is kept for the next PR in the chain.

## Chain

- PR-R1: #768, encoded representation style IDs
- PR-R2: this PR, load representation masks from extras and drop stale rows
- PR-R3: persist representation masks back to extras

Implemented with Codex.

## Validation

```bash
python -m pytest tests/test_viewers.py tests/test_structures.py
```

Result: 35 passed, 20 warnings.


yakutovicha:
Failing pre-commit is fixed in #814.

yakutovicha (approved):
Tested, reviewed, improved - happy 🥳 .

PR #770, opened 2026-06-25
PR-R3: Persist viewer representations in extras

cpignedoli:
## What

PR-R3 in the representation split. This is stacked on top of #769.

This PR adds the persistence side of viewer representations:

- convert viewer representation arrays into JSON-safe masks;
- store those masks in `StructureData`/`CifData` extras;
- add a `Store representations` button for already-stored structures;
- make `Store in AiiDA` persist representation masks when a structure is first stored;
- keep standalone viewer `Apply representations` in sync with extras when the viewer has a stored node.

## Why

Users can now change visualization styles without creating a new structure node. The geometry remains the same AiiDA node, while the mutable viewer state lives in extras.

## Small defaulting fix

This PR also makes the `StructureManagerWidget` storage format selector explicitly default to `StructureData` when no `node_class` is passed.

Previously the `RadioButtons` widget relied on the first entry in `SUPPORTED_DATA_FORMATS`. That is implicit and can silently select the wrong storage type if the option order changes. The new logic keeps explicit caller choices unchanged, but falls back to `StructureData` for the default structure-manager path. This matters here because representation persistence is stored on the AiiDA node extras, so the node type must be the one the widget is actually intended to create.

## Chain

- PR-R1: #768, encoded representation style IDs
- PR-R2: #769, load representation masks from extras and drop stale rows
- PR-R3: this PR, persist representation masks back to extras

Implemented with Codex.

## Validation

```bash
python -m pytest tests/test_viewers.py tests/test_structures.py
```

Result: 37 passed, 24 warnings.

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -19,6 +19,7 @@ from .utils import StatusHTML, exceptions, get_ase_from_file, get_formula
 from .viewers import (
     StructureDataViewer,
     restore_viewer_representations_from_extras,
+    store_viewer_representations_in_extras,
 )
 
 CifData = plugins.DataFactory("core.cif")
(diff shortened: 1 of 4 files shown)


yakutovicha:
superseded by #769

PR #773, opened 2026-07-24
Cleanup test_aiida_code_setup

danielhollas:
Changes in this PR were originally needed when I tested the aiida-core v2.9.0rc1 prerelease due to some breaking changes to `aiida_localhost` fixture. These changes were ultimately reverted in aiida-core and are not present in final v2.9.0 version, but I think this is still a somewhat worthwhile cleanup.


edan-bainglass (approved):
Yeah, less hardcoded. I like 👍

yakutovicha (approved):

PR #775, opened 2026-07-27
Migrate lowest Python version to 3.10

yakutovicha:


aiidalab_widgets_base/wizard.py
```diff
@@ -97,7 +97,7 @@ def __init__(self, steps, show_header=True, **kwargs):
         self.steps = steps
 
         # Unzip the steps to titles and widgets.
-        self.titles, widgets = zip(*steps)
+        self.titles, widgets = zip(*steps, strict=True)
```
edan-bainglass, line 100: What's this about?

danielhollas:
I don't think this is worth it, we'll never run on Python 3.10. I opened an alternative just now to remove the failing job #776. It's probably worth it to keep testing Python 3.9 a while longer in case we need to backport any fixes for 2.x release.

yakutovicha:
> I don't think this is worth it, we'll never run on Python 3.10.

In that case, let's set a minimal version of Python to 3.11 or 3.12. The only reason I opened this PR is because 3.9 is end of life, 3.10 is still not.

> I opened an alternative just now to remove the failing job #776. It's probably worth it to keep testing Python 3.9 a while longer in case we need to backport any fixes for 2.x release.

There is a branch for that, exactly for this kind of fixes: https://github.com/aiidalab/aiidalab-widgets-base/tree/support/2.5.x. We can keep running 3.9 there.

aiidalab_widgets_base/wizard.py
yakutovicha: ruff complained: https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/

danielhollas:
> There is a branch for that, exactly for this kind of fixes: https://github.com/aiidalab/aiidalab-widgets-base/tree/support/2.5.x. We can keep running 3.9 there.

I know, but typically you do the fix on master first and then backport. Backporting is much easier if you can simply cherry-pick the commit and it works as intended, without having to to worry about a compatibility with the Python version.

(to be clear I am for making the minimum version 3.12, just a bit later, I don't see a rush to do that)

yakutovicha:
Alright, fine, I will bring it up in a meeting so we can collectively decide whether we want to move away from 3.9. To not block the development, I am closing this PR and approving yours.

PR #776, opened 2026-07-27
Remove failing CI job

danielhollas:
The Python 3.9 with resolution `lowest-direct` started failing for some unknown reason. Given the iminent migration to 3.12 we simply drop this particular job. I haven't removed testing on Python 3.9 entirely, since it might be useful in case we need to backport any fixes on 2.x release.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
       hooks:
           - id: ruff-format
             exclude: ^docs/.*
-          - id: ruff
+          - id: ruff-check
             args: [--fix, --exit-non-zero-on-fix, --show-fixes]
 
     - repo: https://github.com/jumanjihouse/pre-commit-hook-yamlfmt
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -17,7 +17,7 @@ repos:
       hooks:
           - id: ruff-format
             exclude: ^docs/.*
-          - id: ruff
+          - id: ruff-check
```
danielhollas, line 20: Drive-by fix, the `ruff` pre-commit id is deprecated

edan-bainglass (approved):
LGTM!

yakutovicha (changes requested):
I really don't think this is a good way forward. 3.9 is EOL and we don't plan to use it moving forward. The [2.5.x support branch](https://github.com/aiidalab/aiidalab-widgets-base/tree/support/2.5.x) exists too.

yakutovicha (approved):
As stated in https://github.com/aiidalab/aiidalab-widgets-base/pull/775#issuecomment-5092049277.

PR #777, opened 2026-07-31
Update ruff to 0.16.1 and adopt its new default ruleset

danielhollas:
Same as aiidalab/aiidalab-home#237


edan-bainglass (approved):
As @yakutovicha said elsewhere, less code, cleaner code 💪 LGTM!

PR #781, opened 2026-08-09
Add ty type-checker

danielhollas:
See aiidalab/aiidalab-home#236 for motivation. I also successfully deployed it to AtmoSpec in ispg-group/aiidalab-ispg#290


.pre-commit-config.yaml
```diff
@@ -35,3 +35,16 @@ repos:
...
           - id: nbstripout
+
+    - repo: local
+
+      hooks:
+
+          - id: ty
+            name: ty
+            entry: ty check
+            args: [aiidalab_widgets_base/, notebooks/]
```
danielhollas: I did not add type checking of tests in `tests/` and `notebook_tests` since there are a lot of errors, I'll do that in a follow-up PR.

notebooks/wizard_apps.ipynb
```diff
@@ -55,7 +55,7 @@
     "    \"\"\"Helper class to keep track of our pizza order.\"\"\"\n",
     "\n",
     "    configuration = traitlets.Dict()  # the pizza configuration that was ordered\n",
-    "    status = traitlets.Enum(OrderStatus)\n",
+    "    status = traitlets.UseEnum(OrderStatus)\n",
```
danielhollas, line 58: `traitlets.Enum` accepts a Sequence of valid values, whereas `UseEnum` accepts a `enum.Enum` class. I think this worked by accident since Enum classes are iterable

```python
>>> from enum import Enum
>>> class A(Enum):
...     a = 1
...     b = 2
...     
>>> for a in A:
...     print(a)
...     
A.a
A.b
``` 

See https://traitlets.readthedocs.io/en/stable/trait_types.html#traitlets.UseEnum

pyproject.toml
```diff
@@ -43,3 +43,13 @@ target-version = "py39"
 
 [tool.ruff.lint]
 ignore = ["RUF012"]
+
+# Temporarily disable ty type checking on this file
+[[tool.ty.overrides]]
+include = [
+  "aiidalab_widgets_base/viewers.py",
```
danielhollas, line 50: There's over 70 errors in these files, I'll fix them in a follow-up PR

notebooks/wizard_apps.ipynb
edan-bainglass: Very interesting! I should check on our end (QE app) also (don't believe we use any though).

aiidalab_widgets_base/computational_resources.py
```diff
@@ -1335,7 +1350,9 @@ def __init__(self, description="Select computer:", **kwargs):
 
         description (str): Text to display before dropdown.
         """
-        self._default_user_email = orm.User.collection.get_default().email
+        default_user = orm.User.collection.get_default()
+        assert default_user is not None
+        self._default_user_email = default_user.email
```
edan-bainglass, line 1355: Worth maybe making this an AWB utility, as it is at least common enough to occur also in the QE app

aiidalab_widgets_base/nodes.py
```diff
@@ -127,7 +127,7 @@ def _observe_nodes(self, change):
                 old_nodes=self._tree.nodes, new_nodes=change["new"]
             ),
             key=lambda node: node.pk,
-        )
+        )  # ty: ignore[invalid-assignment]
```
edan-bainglass: How so?

notebooks/viewers.ipynb
```diff
@@ -273,12 +273,13 @@
     "from aiidalab_widgets_base.viewers import ProcessNodeViewerWidget\n",
     "\n",
     "result, workfunction = engine.run_get_node(\n",
-    "    add_multiply, {\"x\": orm.Int(3), \"y\": orm.Int(4), \"z\": orm.Int(5)}\n",
+    "    add_multiply,  # ty: ignore[invalid-argument-type]\n",
```
edan-bainglass, line 276: This is odd and seems like it would occur a lot? 😅

setup.cfg
```diff
@@ -40,13 +40,14 @@ zip_safe = False
...
     bumpver>=2023.1129
-    pgtest~=1.3
     pre-commit>=4.0
+    ty==0.0.69
     pytest~=8.3.0
     pytest-cov~=5.0
     pytest-docker~=3.1.0
     pytest-selenium~=4.1.0
     pytest-timeout~=2.3.0
+    pgtest~=1.3
```
edan-bainglass: I guess just to appease `ty`, but then still removed in #783, yeah? Just checking 🙂

edan-bainglass (approved):
In principle LGTM. Left a few comments. Good to go once tests pass.

aiidalab_widgets_base/computational_resources.py
danielhollas: Maybe. It also feels like perhaps that is an inconvenient API in aiida-core, would probably make sense to raise rather then return None? (perhaps consider for AiiDA v3, but maybe not worth the churn)

notebooks/viewers.ipynb
danielhollas: We had the same issue in aiidalab-home, `ty` doesn't understand the typing of `run_get_node` or `submit` methods which accepts arguments based on the WorkChain.

aiidalab_widgets_base/computational_resources.py
edan-bainglass: No, that's a good opportunity to revisit this. I'll open an issue 🙂

notebooks/viewers.ipynb
edan-bainglass:
```python
TYPE_RUN_PROCESS = Process | type[Process] | ProcessBuilder
...
def run_get_node(
    process: TYPE_RUN_PROCESS,
    inputs: dict[str, t.Any] | None = None,
    **kwargs: t.Any,
) -> tuple[dict[str, t.Any], ProcessNode]:
```
What is the issue? Do we need to be more explicit with the alias, i.e, `TYPE_RUN_PROCESS: TypeAlias = ...` for it to work?

danielhollas: Hmm, you're right, this might be a ty limitation around protocols, the error is

```
error[invalid-argument-type]: Argument to function `run_get_node` is incorrect
   --> notebooks/viewers.ipynb:cell 19:7:5
    |
  7 |     add_multiply,
    |     ^^^^^^^^^^^^ Expected `Process | type[Process] | ProcessBuilder`, found `ProcessFunctionType[(x, y, z), Unknown, WorkFunctionNode]`
```

`ProcessFunctionType` is a Protocol defined in `src/aiida/engine/processes/functions.py` for calcfunction and workfunction decorators.

We can revisit this once `ty` releases a stable version.

setup.cfg
danielhollas: Yep, thanks for catching this, removed

aiidalab_widgets_base/nodes.py
danielhollas: `self._tree.nodes` is a `traitlets.Tuple`. I'll convert to tuple instead of adding the ignore.

```
error[invalid-assignment]: Invalid assignment to data descriptor attribute `nodes` on type `Tree`
     --> aiidalab_widgets_base/nodes.py:125:28
      |
  125 |           self._tree.nodes = sorted(
      |  ____________________________^
  126 | |             self._convert_to_tree_nodes(
  127 | |                 old_nodes=self._tree.nodes, new_nodes=change["new"]
  128 | |             ),
  129 | |             key=lambda node: node.pk,
  130 | |         )
      | |_________^ Expected `tuple[Any, ...]`, found `list[UnspecializedTypeVar]`
  info: Argument to function `TraitType.__set__` is incorrect
  info: This assignment implicitly calls `__set__` on a descriptor of type `Tuple`
  info: Function defined here
     --> .venv/lib/python3.12/site-packages/traitlets/traitlets.py:727:9
      |
  727 |     def __set__(self, obj: HasTraits, value: S) -> None:
      |         ^^^^^^^                       -------- Parameter declared here

  Found 1 diagnostic

aiidalab_widgets_base/computational_resources.py
edan-bainglass: aiidateam/aiida-core#7537

PR #784, opened 2026-08-10
Pin setuptools version to fix notebook tests

danielhollas:
Setuptools 84.0 broke the AWB installation inside the Docker container in notebook tests. As a hotfix let's pin the version. I've opened #785 as a follow-up.

(also contains some unrelated conftest.py cleanup)


tests_notebooks/conftest.py
edan-bainglass, line 1: Can you briefly explain? 🙏

edan-bainglass (approved):
LGTM! Just a call for some explaination regarding the unrelated cleanup 🙏

tests_notebooks/conftest.py
danielhollas: The aiida-core version is now pinned in the AiiDAlab  docker image , so it is pointless to test that its version did not change. Does that make sense?

PR #786, opened 2026-08-10
Simplify RTD config

danielhollas:
https://docs.readthedocs.com/platform/stable/config-file/v2.html#uv


danielhollas:
Hmm, this doesn't work

image

PR #790, opened 2026-08-13
Fix gap on the right in global.css

### LINKED ISSUE #639: Gap on the right
Jupyter defines in the `output_subarea` CSS class a `max-width` of `100% - 14ex`. As far as I can tell, this is the cause of the white space on the right of AiiDAlab apps. One can bypass this by adding in the app's notebooks the following lines:

```
from IPython.display import display, HTML
display(HTML("<style>.output_subarea { max-width:100% !important; }</style>"))
```

This is a bit annoying though. Anyone aware of another solution?

danielhollas:
This brings in the fix that's in QeApp (and others) to AWB.

Closes #639

@edan-bainglass any concerns in bringing this to AWB?

### DIFF
--- a/aiidalab_widgets_base/static/styles/global.css
+++ b/aiidalab_widgets_base/static/styles/global.css
@@ -5,3 +5,11 @@
   height: auto;
   width: auto;
 }
+
+/*
+  Override Jupyter width limitation to
+  make apps take full notebook width
+*/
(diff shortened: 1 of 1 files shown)


edan-bainglass (approved):
I approve this excellent solution 👍

PR #792, opened 2026-08-13
`WizardAppWidget`: add `open_first_step` parameter, True by default

### LINKED ISSUE #793: WizardAppWidget starts fully closed in ipywidgets=8
WizardAppWidgets uses `ipywidgets.Accordion` to control which step is opened. 
`Accordion` (and other similar widgets) [use `selected_index` trait](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#accordion-tab-and-stack-use-selected-index-not-value) to determine which step is opened. Value of None closes all steps, value 0 opens the first step.

The default `selected_index` value changed in ipywidgets 8 from 0 to None so now all steps are closed when the app is first loaded. (Note that QeApp already had this even before, which is why this was not noticed).

I propose that the we should add a parameter that determines this behaviour, e.g. `open_first_step`.
To preserve backwards compatibility, it should be True by default, but if people feel strongly that this default would be wrong, we can also change it.

One point in favour of not changing it is that this issue was causing a bug in AtmoSpec (which is how I noticed), which the nglview viewer didn't work, because apparently the viewer is not properly initialized if the containing ipywidgets is not rendered (opened). QeApp has a horrible workaround for this issue

https://github.com/aiidalab/aiidalab-qe/blob/a4e4280c37aaad1dba52e3fa23c9602cf7c341a3/src/aiidalab_qe/app/structure/step.py#L191

```python
    def _post_render(self):
        # After rendering the widget, nglview needs to be resized
        # to properly display the structure
        self.manager.viewer._viewer._set_size("100%", "300px")
```

I haven't yet figured out how to use this workaround in my app since I don't have the MVC architecture. 
In any case, if we don't restore the previous default behaviour, other application may run into this issue as well if they have a Structure widget in the first step.

danielhollas:
In ipywidgets v8, the default `selected_index` changed from 0 to None,
which means that the WizardAppWidget now starts fully closed.

We want the app developers to have control over this so we add a new parameter `open_first_step` and we have it True by default to be backwards compatible.

In a second commit I added a bit of typing to `wizard.py` so best reviewed commit by commit.

Closes #793. See more details in this issue. 

(happy to bikeshed the parameter name)

### DIFF
--- a/aiidalab_widgets_base/wizard.py
+++ b/aiidalab_widgets_base/wizard.py
@@ -95,6 +95,7 @@ class WizardAppWidget(ipw.VBox):
         self,
         steps: Sequence[tuple[str, WizardAppWidgetStep]],
         show_header: bool = True,
+        selected_index: int | None = 0,
         **kwargs,
     ):
         # The number of steps must be greater than one
@@ -106,7 +107,7 @@ class WizardAppWidget(ipw.VBox):
 
         self.titles, widgets = zip(*steps)
 
-        self.accordion = ipw.Accordion(children=widgets)
+        self.accordion = ipw.Accordion(children=widgets, selected_index=selected_index)
         self._update_titles()
 
         # Watch for changes to each step's state
--- a/aiidalab_widgets_base/wizard.py
+++ b/aiidalab_widgets_base/wizard.py
@@ -95,7 +95,7 @@ class WizardAppWidget(ipw.VBox):
         self,
         steps: Sequence[tuple[str, WizardAppWidgetStep]],
         show_header: bool = True,
-        selected_index: int | None = 0,
+        open_first_step: bool = True,
         **kwargs,
     ):
         # The number of steps must be greater than one
@@ -106,6 +106,8 @@ class WizardAppWidget(ipw.VBox):
         self.steps = steps
 
         self.titles, widgets = zip(*steps)
+
+        selected_index = 0 if open_first_step else None
 
         self.accordion = ipw.Accordion(children=widgets, selected_index=selected_index)
         self._update_titles()
--- a/tests/test_wizard.py
+++ b/tests/test_wizard.py
@@ -141,9 +141,13 @@ def test_closed_first_step():
         selected_index=None,
     )
 
-    # All steps should be closed initially
-    # First step should be opened by default
+    # All steps should be closed
     assert widget.accordion.selected_index is None
+
+    # With no step selected, all header buttons should be disabled.
+    assert widget.back_button.disabled is True
+    assert widget.next_button.disabled is True
+    assert widget.reset_button.disabled is True
 
 
 def test_at_least_two_steps():
--- a/tests/test_wizard.py
+++ b/tests/test_wizard.py
@@ -150,6 +150,24 @@ def test_closed_first_step():
     assert widget.reset_button.disabled is True
 
 
+def test_selected_index_at_construction():
+    widget = WizardAppWidget(
+        steps=[
+            ("Setup", Step1()),
+            ("View results", Step2()),
+        ],
+        selected_index=1,
+    )
+
+    # The widget should open directly on the requested step
+    assert widget.accordion.selected_index == 1
+    assert widget.selected_index == 1
+    # Buttons should reflect that step's state immediately,
+    # without requiring any further interaction.
+    assert widget.back_button.disabled is False  # this is NOT the case atm
+    assert widget.next_button.disabled is True  # s2 is still in State.INIT
+
+
 def test_at_least_two_steps():
     with pytest.raises(AtLeastTwoStepsWizardError):
         WizardAppWidget(
--- a/tests/test_wizard.py
+++ b/tests/test_wizard.py
@@ -53,52 +53,6 @@ class Step2(ipw.HBox, WizardAppWidgetStep):
 
 
 def test_wizard_app_widget():
-    class Step1(ipw.HBox, WizardAppWidgetStep):
-        config = tl.Bool()
-
-        def __init__(self, **kwargs):
-            self.order_button = ipw.Button(description="Submit order", disabled=False)
-            self.order_button.on_click(self.submit_order)
-            super().__init__(children=[self.order_button], **kwargs)
-
-        def submit_order(self, _=None):
-            self.config = True
-
-        @tl.default("config")
-        def _default_config(self):
-            return False
-
-        def reset(self):
-            self.config = False
-
-        @tl.observe("config")
-        def _observe_config(self, _=None):
-            self.state = self.State.SUCCESS if self.config else self.State.INIT
-
-    class Step2(ipw.HBox, WizardAppWidgetStep):
-        config = tl.Bool()
-
-        def __init__(self, **kwargs):
-            self.results = ipw.HTML("Results")
-            super().__init__(children=[self.results], **kwargs)
-
-        def submit_order(self, _=None):
-            pass
-
-        def reset(self):
-            pass
-
-        @tl.default("config")
-        def _default_config(self):
-            return False
-
-        @tl.observe("config")
-        def _observe_config(self, change):
-            if self.config:
-                self.state = self.State.READY
-            else:
-                self.state = self.State.INIT
-
     s1 = Step1(auto_advance=True)
     s2 = Step2(auto_advance=True)
     tl.dlink((s1, "config"), (s2, "config"))
--- a/tests/test_wizard.py
+++ b/tests/test_wizard.py
@@ -1,7 +1,9 @@
 import ipywidgets as ipw
+import pytest
 import traitlets as tl
 
 from aiidalab_widgets_base import WizardAppWidget, WizardAppWidgetStep
+from aiidalab_widgets_base.wizard import AtLeastTwoStepsWizardError
 
 
 class Step1(ipw.HBox, WizardAppWidgetStep):
@@ -142,3 +144,33 @@ def test_closed_first_step():
     # All steps should be closed initially
     # First step should be opened by default
     assert widget.accordion.selected_index is None
+
+
+def test_at_least_two_steps():
+    with pytest.raises(AtLeastTwoStepsWizardError):
+        WizardAppWidget(
+            steps=[
+                ("Setup", Step1()),
+            ],
+        )
+
+
+def test_invalid_selected_index():
+    with pytest.raises(tl.TraitError, match="Invalid selection: index out of bounds"):
+        WizardAppWidget(
+            steps=[
+                ("Setup", Step1()),
+                ("View results", Step2()),
+            ],
+            selected_index=-1,
+        )
+
+    # Test that index greater than number of steps raises
+    with pytest.raises(tl.TraitError, match="Invalid selection: index out of bounds"):
+        WizardAppWidget(
+            steps=[
+                ("Setup", Step1()),
+                ("View results", Step2()),
+            ],
+            selected_index=2,
+        )
--- a/tests/test_wizard.py
+++ b/tests/test_wizard.py
@@ -138,7 +138,7 @@ def test_closed_first_step():
             ("Setup", s1),
             ("View results", s2),
         ],
-        selected_index=None,
+        open_first_step=False,
     )
 
     # All steps should be closed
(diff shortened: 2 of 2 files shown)


edan-bainglass (approved):
LGTM! TBD if we want to replace this with the MVC-based one in QE, i.e., move the aiidalab_qe/common/wizard package to AWB.

edan-bainglass:
@danielhollas I approved this rather quickly, as it is fine in principle, though can't one just set the `selected_index` to their chosen default step? Setting it to 0 is effectively what `open_first_step` does, no?

danielhollas:
> though can't one just set the selected_index to their chosen default step? Setting it to 0 is effectively what open_first_step does, no?

Sure, but then you'd be accessing implementation details of the class (via self.accordion)

edan-bainglass:
> Sure, but then you'd be accessing implementation details of the class (via self.accordion)

```python
class Wizard(ipw.Accordion):
    def __init__(
        self,
        steps: dict[str, ipw.VBox],
        selected_index: int | None = None,  # or 0, TBD
    ):
        super().__init__(children=list(steps.values()))

        for i, title in enumerate(steps.keys()):
            self.set_title(i, title)

        self.selected_index = selected_index


STEPS = {
    "1. Input": ipw.VBox([ipw.HTML("<h3>Step 1: Input</h3>")]),
    "2. Workflow": ipw.VBox([ipw.HTML("<h3>Step 2: Workflow</h3>")]),
    "3. Output": ipw.VBox([ipw.HTML("<h3>Step 3: Output</h3>")]),
}

wizard = Wizard(steps=STEPS, selected_index=None)
```
We can call our `selected_index` whatever, say, `default_step`. My point is to keep it general, in case a developer wishes to open their wizard in step 2 by default (for some reason). With `open_first_step`, we hardcode "first" 🙂

danielhollas:
Oh, I see that make sense, I'll update the PR.

edan-bainglass (approved):
LGTM! Thanks @danielhollas. I'm only wondering if `default_step` is more user friendly, though one may argue that `selected_index` is more consistent 🤷🏻‍♂️ @yakutovicha thoughts?

yakutovicha:
> LGTM! Thanks @danielhollas. I'm only wondering if `default_step` is more user friendly, though one may argue that `selected_index` is more consistent 🤷🏻‍♂️ @yakutovicha thoughts?

If unclear, I would choose whatever ipywidgets have chosen, at least to be in sync. In that case, `selected_index`.

tests/test_wizard.py
```diff
@@ -4,6 +4,54 @@
 from aiidalab_widgets_base import WizardAppWidget, WizardAppWidgetStep
 
 
+class Step1(ipw.HBox, WizardAppWidgetStep):
```
yakutovicha, line 9: It is literally a copy of the classes defined in the test below - maybe you should delete the ones in the `test_wizard_app_widget()`?

tests/test_wizard.py
```diff
@@ -118,3 +171,20 @@ def _observe_config(self, change):
...
+    widget = WizardAppWidget(
+        steps=[
+            ("Setup", s1),
+            ("View results", s2),
+        ],
+        selected_index=None,
+    )
+
+    # All steps should be closed initially
+    # First step should be opened by default
```
yakutovicha: leftover?
```suggestion
```

tests/test_wizard.py
```diff
@@ -118,3 +171,20 @@ def _observe_config(self, change):
...
+    assert widget.accordion.selected_index is None
```
yakutovicha, line 145: Please also test that the Navigation buttons are correctly disabled when the selected index is `None`.

```suggestion
    assert widget.accordion.selected_index is None
    # With no step selected, all header buttons should be disabled.
    assert widget.back_button.disabled is True
    assert widget.next_button.disabled is True
    assert widget.reset_button.disabled is True
```

tests/test_wizard.py
```diff
@@ -118,3 +171,20 @@ def _observe_config(self, change):
     assert widget.accordion.selected_index == 1
     assert widget.next_button.disabled is True
     assert widget.back_button.disabled is False
+
+
+def test_closed_first_step():
```
yakutovicha, line 132: Since this is a new possibility, please also test when the index at initialization is 1:

```python
def test_selected_index_at_construction():
    s1 = Step1()
    s2 = Step2()

    widget = WizardAppWidget(
        steps=[
            ("Setup", s1),
            ("View results", s2),
        ],
        selected_index=1,
    )

    # The widget should open directly on the requested step ...
    assert widget.accordion.selected_index == 1
    assert widget.selected_index == 1
    # ... and buttons should reflect that step's state immediately,
    # without requiring any further interaction.
    assert widget.back_button.disabled is False # this is NOT the case atm
    assert widget.next_button.disabled is True  # s2 is still in State.INIT
```

And, more generally, what do we mean when the index is set to 1? How `back_button`/`next_button` should be displayed? Especially when step 0 isn't successful, should `back_button` be enabled to invite the user to go back? If that is the case, then the current implementation is not doing that currently.

aiidalab_widgets_base/wizard.py
```diff
@@ -88,19 +91,23 @@ class WizardAppWidget(ipw.VBox):
 
     selected_index = tl.Int(allow_none=True)
 
-    def __init__(self, steps, show_header=True, **kwargs):
+    def __init__(
+        self,
+        steps: Sequence[tuple[str, WizardAppWidgetStep]],
+        show_header: bool = True,
+        selected_index: int | None = 0,
```
yakutovicha: Please add validation when the `selected_index` is out of range. Also, make sure to have a test for that.
Something like:
```python
        if selected_index is not None and not 0 <= selected_index < len(steps):
            raise ValueError(
                f"selected_index must be None or in range [0, {len(steps) - 1}], "
                f"got {selected_index}."
            )
```

aiidalab_widgets_base/wizard.py
```diff
@@ -5,7 +5,10 @@
     * Carl Simon Adorf <simon.adorf@epfl.ch>
 """
 
+from __future__ import annotations
```
yakutovicha, line 8: Just flagging that this is a drive-by fix and, strictly speaking, not part of the issue.

yakutovicha (changes requested):
Thanks very much @danielhollas, I have a few minor requests.

Also, a side remark about type annotations. I understand the desire to clean up other things while doing the work, but let's stay disciplined about keeping that to separate PRs. I see two main reasons for doing that:

- It keeps PRs focused and goal-oriented, so later it is easier to find what was done by looking at the title/description.
- If a change needs to be reverted, it reverts other unexpected things (type annotations in that case).

This is not a blocker, but if you have some time, please split them up. Otherwise, go ahead, but remember for the future 🙏 .

aiidalab_widgets_base/wizard.py
danielhollas: I thought about adding explicit validation at first, but we would essentially be duplicating the validation that Accordion already does. 
So I don't think it's worth it. 

I should certainly add a test (if I hadn't already).

tests/test_wizard.py
danielhollas: Oh, this was certainly my intention, I just forgot to delete them, thanks for catching this!

danielhollas:
> Also, a side remark about type annotations. I understand the desire to clean up other things while doing the work, but let's stay disciplined about keeping that to separate PRs.

Agreed, I've opened #796 and rebased this one on top of it. Marking as draft while I work on the review, thanks!

aiidalab_widgets_base/wizard.py
danielhollas: Added tests in 72d4184

tests/test_wizard.py
danielhollas: Thanks, added in 8c7c34f

tests/test_wizard.py
danielhollas: Yep

tests/test_wizard.py
danielhollas: I added your proposed test in 9731edd and it passes so I think the buttons are behaving correctly here, because the existing logic already handles the case where Step 2 is opened while Step 1 is not finished (you can do that manually by clicking, this is not new in this PR).

> And, more generally, what do we mean when the index is set to 1?

Well, this is why I originally only had a boolean `opened_first_step`, because I can't imagine the case where setting `selected_index` to anything other than 0 or None is useful. WDYT? We might want to revisit this.

danielhollas:
https://github.com/aiidalab/aiidalab-widgets-base/pull/792#issuecomment-5289938045

> We can call our selected_index whatever, say, default_step. My point is to keep it general, in case a developer wishes to open their wizard in step 2 by default (for some reason). With open_first_step, we hardcode "first" 🙂

I am still not sure about this. "Keeping it general" feels like it results in a somewhat confusing API, since I can't imagine how the fully generality is useful at all (i.e. why would you *ever* want your app to be opened at second step???

aiidalab_widgets_base/wizard.py
yakutovicha: > I thought about adding explicit validation at first, but we would essentially be duplicating the validation that Accordion already does.

It would give a little bit clearer error message that belongs to the wizard, not the accordion underneath. But either way is fine.

yakutovicha:
> [#792 (comment)](https://github.com/aiidalab/aiidalab-widgets-base/pull/792#issuecomment-5289938045)
> 
> > We can call our selected_index whatever, say, default_step. My point is to keep it general, in case a developer wishes to open their wizard in step 2 by default (for some reason). With open_first_step, we hardcode "first" 🙂
> 
> I am still not sure about this. "Keeping it general" feels like it results in a somewhat confusing API, since I can't imagine how the fully generality is useful at all (i.e. why would you _ever_ want your app to be opened at second step???

I also cannot imagine the use case for keeping it general, but that doesn't mean there isn't. Ultimately, I don't see severe consequences of any choice, tbh. So my opinion is as much of a value as tossing a coin. Claude is also unsure :).

yakutovicha (approved):

edan-bainglass:
@danielhollas I got a ping for review, but this is closed?

danielhollas:
Stupid github closed the PR after I merged #796. I will need to reopen, but the review request was mostly for the discussion regarding the discussion about `open_first_step` versus `selected_index`. Please see the last two comments and feel free to reply here.

danielhollas:
Or maybe I can reopen?

EDIT: Apparently not

danielhollas:
Let's try re-opening again

edan-bainglass:
> Stupid github closed the PR after I merged #796. I will need to reopen, but the review request was mostly for the discussion regarding the discussion about `open_first_step` versus `selected_index`. Please see the last two comments and feel free to reply here.

I'm still in favor of not hard coding `first_step` but also against the vague `selected_index`. I had suggested `default_step` as an alternative.

danielhollas:
Haha, looks like we're not converging. :sweat_smile: 
(sorry for the bikeshedding here, but since this is a public API it's worth to discuss a bit).

> We can call our selected_index whatever, say, default_step. My point is to keep it general, in case a developer wishes to open their wizard in step 2 by default (for some reason). With open_first_step, we hardcode "first" 🙂

> I'm still in favor of not hard coding first_step but also against the vague selected_index. I had suggested default_step as an alternative.

@edan-bainglass you seem to have the firmest opinion here, but could you expand your reasoning a little bit?

I understand the appeal of generality, but in this case this feels like a case of [premature abstraction](https://arendjr.nl/blog/2024/07/post-architecture-premature-abstraction-is-the-root-of-all-evil/#:~:text=I%20would%20adapt%20one,help%20maintainability.). 

I was originally like Sasha roughly fifty-fifty between the options, but now I am leaning more to my original boolean solution, because

1. It feels more obvious and easier to explain
2. You don't have to worry about validation. 
3. `default_step` feels confusing, and it's not even clear what type it should be. 
4. `default_step=None` means all steps are closed at the beginning, but that is not obvious. 
5. Seriously, if you think your `default_step` should be anything other than the first step, that feels like a design issue with the application, and goes against the concept of WizardStep widget imo.


I'll let this sit over the weekend, and unless convinced or otherwise stopped I am going to make an executive decision on Tuesday, as the author of the PR. :-) 

EDIT: Marking as draft to make it clear we have a decision to make

edan-bainglass:
The first step of my hypothetical wizard is optional (carries some defaults that are applicable most of the time). It still represents the first step (or rather, a pre-step), as it affects the next. But I prefer to start the app in the second step, with a friendly message that the user may visit the first step to change defaults if they wish. But by default, I'd want the `default_step` to be 2 (equal to `selected_index = 1`. If `default_step` is then a step index >= 1, with 0 or None being all are closed (that last part is not ideal). We could also just use `selected_index` and match the underlying accordion API. My point is that if we don't generalize, then tomorrow someone comes with that application, as it tends to happen.

Anyhow, let's proceed democratically. If @danielhollas and @yakutovicha believe that this is not necessary, carry on. If and when this use case comes around, we can create a wrapper `Wizard` to address it, and we'll have a concrete answer for the next major version 🙂 I'm fine with it.

danielhollas:
@yakutovicha given the discussion above, are you still fifty-fifty on the two proposed solutions?

yakutovicha:
> @yakutovicha given the discussion above, are you still fifty-fifty on the two proposed solutions?

Yes, mostly because I don't see a severe negative consequence of any decision. I spent some time trying to find a definitive factor that would make one decision definitely preferable - I didn't find any.

What I do have an opinion about is "a bad decision is better than no decision", simply because a bad decision can be fixed. 

@danielhollas, let's be pragmatic here; we've already spent enough time. Since this particular change is not a high-stakes call, please go ahead with your `open_first_step: bool` (or any other name you want), and if a new use case comes out, we can generalize that in the future.

danielhollas:
Alright, thank you both for the discussion.

> @danielhollas, let's be pragmatic here; we've already spent enough time. Since this particular change is not a high-stakes call, please go ahead with your open_first_step: bool (or any other name you want), and if a new use case comes out, we can generalize that in the future.

Yep, in the end I went with `open_first_step` as a simpler solution. If the need arises in the future, I will first eat my hat, and then implement `selected_index`. :-)

PR #797, opened 2026-08-20
Replace `ipw.Output` with `ipw.VBox` across the codebase

### LINKED ISSUE #717: Avoid using the `ipywidgets.Output` widget
The widget behaves quite unpredictably. For versions of ipywidgets, it fails to show
the content and prevents other widgets from being displayed.

yakutovicha:
`ipw.Output` has caused some display bugs in AiiDAlab before
(aiidalab/aiidalab-home#185), where content written into an `Output`
failed to render. In this PR, we replace `Output` with `VBox` everywhere.
`VBox` keeps children as an ordinary trait in the codebase.

fixes #717

### DIFF
--- a/aiidalab_widgets_base/nodes.py
+++ b/aiidalab_widgets_base/nodes.py
@@ -7,7 +7,6 @@ import ipywidgets as ipw
 import traitlets as tl
 from aiida import common, engine, orm
 from aiida.cmdline.utils.ascii_vis import calc_info
-from IPython.display import clear_output, display
 
 
 class AiidaNodeTreeNode(ipytree.Node):
@@ -54,7 +53,7 @@ class UnknownTypeTreeNode(AiidaNodeTreeNode):
     icon = tl.Unicode("file").tag(sync=True)
 
 
-class NodesTreeWidget(ipw.Output):
+class NodesTreeWidget(ipw.VBox):
     """A tree widget for the structured representation of a nodes graph."""
 
     nodes = tl.Tuple().tag(trait=tl.Instance(orm.Node))
@@ -80,15 +79,7 @@ class NodesTreeWidget(ipw.Output):
         self._tree = ipytree.Tree()
         self._tree.observe(self._observe_tree_selected_nodes, ["selected_nodes"])
 
-        super().__init__(**kwargs)
-
-    def _refresh_output(self):
-        # There appears to be a bug in the ipytree implementation that sometimes
-        # causes the output to not be properly cleared. We therefore refresh the
-        # displayed tree upon change of the process trait.
-        with self:
-            clear_output()
-            display(self._tree)
+        super().__init__(children=[self._tree], **kwargs)
 
     def _observe_tree_selected_nodes(self, change):
         for node in change["new"]:
@@ -131,7 +122,6 @@ class NodesTreeWidget(ipw.Output):
             )
         )
         self.update()
-        self._refresh_output()
 
     @classmethod
     def _to_tree_node(cls, node, name=None, **kwargs):
--- a/aiidalab_widgets_base/process.py
+++ b/aiidalab_widgets_base/process.py
@@ -5,10 +5,10 @@
 from __future__ import annotations
 
 import inspect
-import sys
 import threading
 import traceback
 import warnings
+from html import escape
 
 import ipywidgets as ipw
 import traitlets as tl
@@ -200,7 +200,7 @@ class ProcessMonitor(tl.HasTraits):
         self._monitor_thread_stop = threading.Event()
         self._monitor_thread_lock = threading.Lock()
 
-        self.log_widget: ipw.Output | None = kwargs.pop("log_widget", None)
+        self.log_widget: ipw.VBox | None = kwargs.pop("log_widget", None)
 
         super().__init__(**kwargs)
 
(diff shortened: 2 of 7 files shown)


danielhollas:
@yakutovicha thanks! 

According to Codecov, 32 lines are missing test coverage. Could you please add unit tests for those, if possible?

edan-bainglass (approved):
Thanks @yakutovicha. Fine in principle. One note regarding CSS - there was one case in the QE app where I was styling with the assumption that the widget is an `Output` (bad idea 😅). This has since been addressed. Just noting here, in case similar styling exists, though it looks like you're already addressing some styles (e.g., border). Good to go, once tests are added and passing👍

yakutovicha:
> @yakutovicha thanks!
> 
> According to Codecov, 32 lines are missing test coverage. Could you please add unit tests for those, if possible?

done ✅

aiidalab_widgets_base/bug_report.py
```diff
@@ -152,14 +152,13 @@ def install_create_github_issue_exception_handler(output, url, labels=None):
     .. highlight:: python
     .. code-block:: python
 
-        output = ipw.Output()
+        output = ipw.VBox()
         install_create_github_issue_exception_handler(
             output,
             url='https://github.com/aiidalab/aiidalab-qe/issues/new',
             labels=('bug', 'automated-report'))
```
danielhollas: If we're breaking the API, I'd rather have this function to return the output object, because then we have control over it and don't have to rely on the user passing in a correct widget.

```suggestion
        output = install_create_github_issue_exception_handler(
            url='https://github.com/aiidalab/aiidalab-qe/issues/new',
            labels=('bug', 'automated-report'))
```

aiidalab_widgets_base/nodes.py
```diff
@@ -80,15 +79,7 @@ def __init__(self, **kwargs):
         self._tree = ipytree.Tree()
         self._tree.observe(self._observe_tree_selected_nodes, ["selected_nodes"])
 
-        super().__init__(**kwargs)
-
-    def _refresh_output(self):
-        # There appears to be a bug in the ipytree implementation that sometimes
```
danielhollas, line 86: @yakutovicha just double checking, have you verified that this problem is gone by switching to VBox?

tests/test_bug_report.py
```diff
@@ -19,3 +21,42 @@ def test_fingerprint_parser():
...
+
+    assert len(output.children) == 1
+    msg = output.children[0]
+    assert isinstance(msg, ipw.HTML)
+    assert "aiidalab-qe/issues/new" in msg.value
+    assert "boom" in msg.value
+
+    # Restoring puts the original handler back.
+    restore()
+    assert fake_ipython._showtraceback == "original-handler"
```
danielhollas: Could we test that after restoring, raised Exceptions are actually raised and not caught, e.g. with `pytest.raises` or something.

tests/test_bug_report.py
```diff
@@ -19,3 +21,42 @@ def test_fingerprint_parser():
...
+    # Installing the handler replaces IPython's traceback display with a
+    # callable closure, not just some other placeholder value.
+    assert callable(fake_ipython._showtraceback)
+    assert fake_ipython._showtraceback != "original-handler"
+
+    traceback_lines = [
+        "Traceback (most recent call last):\n",
+        "ValueError: boom\n",
+    ]
+    fake_ipython._showtraceback(ValueError, ValueError("boom"), traceback_lines)
```
danielhollas: I don't understand how this works. Can you add a comment?

danielhollas (changes requested):
Thanks for adding tests!

Quick review for now, I haven't checked the new tests for ComputationalResource widget yet.

My main ask: Can we split the `bug_report` change into a separate PR please. It is a breaking change and deserves a separate discussion.

aiidalab_widgets_base/bug_report.py
danielhollas: Hmm, I am not sure if my suggestion makes sense, need to think about it more.

aiidalab_widgets_base/bug_report.py
```diff
@@ -174,7 +173,7 @@ def install_create_github_issue_exception_handler(output, url, labels=None):
 
     def create_github_issue_exception_handler(exception_type, exception, traceback):
         try:
-            output.clear_output()
```
danielhollas, line 177: Hmm, if we wanted to preserve backwards compatibility, we could maybe introspect the object and behave accordingly. Not sure if it is worth it.

aiidalab_widgets_base/bug_report.py
```diff
@@ -174,7 +173,7 @@ def install_create_github_issue_exception_handler(output, url, labels=None):
...
+            output.children = ()
```
danielhollas: Wait, doesn't this erase the whole application?

yakutovicha:
> My main ask: Can we split the `bug_report` change into a separate PR please. It is a breaking change and deserves a separate discussion.

done in #798

aiidalab_widgets_base/nodes.py
yakutovicha: Yes, the displayed tree is now properly reset after updating the node UUID.

yakutovicha:
@danielhollas, as you've requested, the bug report has been moved to a different PR. Also, I manually tested that `NodesTreeWidget` works as expected.

aiidalab_widgets_base/process.py
```diff
@@ -200,7 +200,7 @@ def __init__(self, callbacks=None, on_sealed=None, timeout=None, **kwargs):
         self._monitor_thread_stop = threading.Event()
         self._monitor_thread_lock = threading.Lock()
 
-        self.log_widget: ipw.Output | None = kwargs.pop("log_widget", None)
```
danielhollas, line 203: @yakutovicha this is again a breaking change, right? User must pass a VBox or something like that instead of ipw.Output?

danielhollas (approved):
@yakutovicha thanks. One question, looks like there is one breaking change in this PR as well? Otherwise looks good.

aiidalab_widgets_base/process.py
edan-bainglass: Indeed a breaking change, confirmed but this breaking the QE app (the OG use case of the log widget), where I had to update to using a `VBox`.

danielhollas: Cool, as long as it works that's great. We'll note it in the changelog. 

@yakutovicha Should be good to merge.

PR #798, opened 2026-08-24
Migrate bug report module: `Output` to `VBox`

yakutovicha:


tests/test_bug_report.py
```diff
@@ -19,3 +21,42 @@ def test_fingerprint_parser():
...
+    restore = bug_report.install_create_github_issue_exception_handler(
+        output,
+        url="https://github.com/aiidalab/aiidalab-qe/issues/new",
+        labels=("bug", "automated-report"),
+    )
+
+    # Installing the handler replaces IPython's traceback display with a
+    # callable closure, not just some other placeholder value.
+    assert callable(fake_ipython._showtraceback)
+    assert fake_ipython._showtraceback != "original-handler"
```
danielhollas:
```traceback
```

This assert can never be false given the assert just above it, let's remove it.

aiidalab_widgets_base/bug_report.py
```diff
@@ -152,14 +152,13 @@ def install_create_github_issue_exception_handler(output, url, labels=None):
     .. highlight:: python
     .. code-block:: python
 
-        output = ipw.Output()
+        output = ipw.VBox()
```
danielhollas, line 155: As noted in the original PR, can we change the API so that the output widget created inside the function and returned, rather than being passed in as an argument?

danielhollas (commented):
Thanks for splitting this out @yakutovicha 

Have you had a chance to test this yet?

tests/test_bug_report.py
yakutovicha: done in [0bc1d33](https://github.com/aiidalab/aiidalab-widgets-base/pull/798/commits/0bc1d33daf2bd5e6669a5ebcd8424cc3ac5982c4)

aiidalab_widgets_base/bug_report.py
yakutovicha: I wouldn't change it. This `output` is handled by the external app, while the function's job is to update it in place, not to decide where it lives. In `aiidalab-qe`, the caller even switches between two different existing widgets depending on config (`self.log_widget` or `self.view.output`), see [here](https://github.com/aiidalab/aiidalab-qe/blob/5db447b675461b21fdf5df98883bba3ff6ecdfb5/src/aiidalab_qe/app/main.py#L80).

Also, it wouldn't make the implementation simpler. The caller would still have to grab whatever's returned and manually place it. Moreover, the function already returns restore, so you'd end up needing a tuple return instead of a clean single value.

yakutovicha:
Tested together with the QE app; requires the following changes: https://github.com/aiidalab/aiidalab-qe/pull/1577.

danielhollas (approved):
Thanks @yakutovicha

PR #800, opened 2026-08-30
Remove pylint pragmas

danielhollas:


danielhollas:
This is trivial cleanup so merging without review

PR #804, opened 2026-08-30
Add pre-commit job to run ty

danielhollas:
In #781 I've added ty as a pre-commit hook, but I forgot to add a manual pre-commit job to CI, because ty cannot run on `pre-commit.ci` service.

This is the same setup we have in aiidalab-home (and aiidalab with mypy)


yakutovicha (approved):
👍

PR #805, opened 2026-09-01
Add typing to bug_report.py

danielhollas:
Added typing to `bug_report.py`, motivated by #798, this will help Apps catch the breaking change, since now they have to pass `ipw.Box` type instead of `ipw.Output`


yakutovicha (approved):
👍

PR #807, opened 2026-09-01
Remove custom ListOrTuppleError

danielhollas:
This custom exception class is only used in one place, and is kind of a weird exception to have anyways so let's remove it for the v3 release. 

In general, I don't think runtime-checking of types like this goes against python duck typing philosophy and errors like this should rather be called by a static type checker, so I'd be fine to also just remove this code altogether. Also notice that we only do this check for `importers`, and not e.g. for `editors`.

(This is I think a last cleanup I wanted to do before v3)

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -5,6 +5,7 @@ import functools
 import io
 import pathlib
 import tempfile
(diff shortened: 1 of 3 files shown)


yakutovicha (approved):
👍

PR #809, opened 2026-09-02
Drop python 3.9 support

danielhollas:
I think with the v3 release now ready we can drop the python 3.9 support.

If we wanted to keep possible backports easier, we could also instruct ruff to still have minimum python 3.9 support.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
     "PyCifRW~=4.4",
     "aiida-core>=2.8,<3",
     "ansi2html~=1.7",
-    "ase~=3.18",
+    "ase~=3.23",
     "humanfriendly~=10.0",
     "ipytree~=0.2",
     "traitlets~=5.4",
@@ -46,7 +46,7 @@ dev = [
     "bumpver>=2023.1129",
     "pre-commit>=4.0",
     "ty==0.0.77",
-    "pytest~=8.3.0",
+    "pytest~=8.4.0",
     "pytest-cov~=5.0",
     "pytest-docker~=3.1.0",
     "pytest-selenium~=4.1.0",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -21,7 +21,7 @@ classifiers = [
 
 dependencies = [
     "PyCifRW~=4.4",
-    "aiida-core>=2.7,<3",
+    "aiida-core>=2.8,<3",
```
danielhollas, line 24: We could also bump to 2.9 since that's gonna be the version in the official image.

edan-bainglass: After all, the AiiDA team decided enough non-breaking changes piled up to justify a 2.10 release (upcoming). I'll test it when I'm back.

edan-bainglass (approved):
LGTM! Okay to wait for aiida-core 2.10?

pyproject.toml
```diff
@@ -39,22 +39,22 @@ dependencies = [
...
+    "pytest~=8.4.0",
     "pytest-cov~=5.0",
     "pytest-docker~=3.1.0",
     "pytest-selenium~=4.1.0",
     "pytest-timeout~=2.3.0",
     "selenium~=4.23.0",
 ]
 optimade = ["ipyoptimade>=1.2.1,<2"]
-smiles = ["rdkit>=2021.09.2"]
+smiles = ["rdkit>=2024.9.6"]
```
danielhollas, line 57: One of the test was segfaulting on earlier rdkit versions.

pyproject.toml
```diff
@@ -21,9 +21,9 @@ classifiers = [
 
 dependencies = [
     "PyCifRW~=4.4",
-    "aiida-core>=2.7,<3",
+    "aiida-core>=2.8,<3",
     "ansi2html~=1.7",
-    "ase~=3.18",
+    "ase~=3.23",
```
danielhollas, line 26: Needed a bump to ensure numpy compatibility.

danielhollas (commented):
Why wait for 2.10? After all we're only setting minimum aiida-core version here. 

Once new aiida-core version comes we will test it here, but I don't think it needs to block v3 release.

yakutovicha (approved):

edan-bainglass:
> Why wait for 2.10? After all we're only setting minimum aiida-core version here.
> 
> Once new aiida-core version comes we will test it here, but I don't think it needs to block v3 release.

Ah, sorry. Missed that's its a minimum pin 👍

PR #810, opened 2026-09-02
More type-checking for structures.py and viewers.py

danielhollas:
Still a lot of type errors in structure.py and viewers.py, this was just some low-hanging fruit.


yakutovicha (approved):
👍

PR #814, opened 2026-09-07
Fix `StructureManagerWidget` defaulting to `CifData`

### LINKED ISSUE #813: `StructureManagerWidget` defaults to `CifData`, not `StructureData`
`StructureManagerWidget` declares `StructureData` as its default node class, but gets `CifData`:

```python
widget = awb.StructureManagerWidget(importers=[])
assert widget.node_class == "StructureData"  # fails: it is "CifData"
```

`SUPPORTED_DATA_FORMATS` is CifData-first, the data-format `RadioButtons` is built without an explicit `value` so it picks its first option, and `tl.link` immediately pushes that into `node_class`, overwriting the `@tl.default`.

yakutovicha:
The `RadioButtons` format selector had no explicit initial value, so it picked its first option, and the `tl.link` to `node_class` immediately overwrote the declared `StructureData` default with `CifData`.

Set the selector's initial value from `node_class` (defaulting to `StructureData`) so it agrees with the trait from construction, and validate `node_class` upfront instead of after building the widget.

fixes #813

### DIFF
--- a/aiidalab_widgets_base/structures.py
+++ b/aiidalab_widgets_base/structures.py
@@ -102,11 +102,19 @@ class StructureManagerWidget(ipw.VBox):
         self.structure_label = ipw.Text(description="Label")
         self.structure_description = ipw.Text(description="Description")
 
+        # Validate node_class before it's used to pick the format selector's default.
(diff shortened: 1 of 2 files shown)


cpignedoli (approved):
I guess this was moved from #769 so OK from my side

PR #815, opened 2026-09-09
Drop the special-cased default representation

yakutovicha:
Its fixed array name `_aiidalab_viewer_representation_default` doesn't encode the style from #768, so its type, size and color reset on reload. The name earned nothing else: "always present, cannot be deleted" belongs to the list, not to one member.

- Every representation gets an encoded style id. Removes `DEFAULT_REPRESENTATION` and `_sync_style_id_with_settings`; an unstyled name is renamed on the next apply.
- The delete button is hidden only on the last representation, so `deletable` is gone.
- Replaces `atom_show_threshold` with an `_AtomMarker` enum: `IN` (1), `OUT` (-1), `UNASSIGNED` (0). Membership is uniformly `== IN`.
- Unassigned always go to the first representation, so a new atom is displayed exactly once.
- An array is seeded only when the structure carries none.

### DIFF
--- a/aiidalab_widgets_base/viewers.py
+++ b/aiidalab_widgets_base/viewers.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import base64
 import copy
 import csv
+import enum
 import io
 import os
 import pathlib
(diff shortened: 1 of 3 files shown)


cpignedoli (approved):
tested on local instance tehre is a bug where slection must be refereshed otehrwise we cannot load a new structure but to be fixed in a new PR

